PR #18, opened 2026-04-28
fix(beam): preserve memory id across legacy and working memory

Jammu:
## Summary

This fixes legacy/BEAM dual-write ID drift by letting `Mnemosyne.remember()` pass the generated legacy memory ID into `BeamMemory.remember()`.

Without this, the legacy `memories` row and BEAM `working_memory` row can receive different IDs for the same memory. That makes ID-based operations such as `forget(memory_id)` only partially effective: the legacy row can be removed while the BEAM working-memory row remains as a ghost memory.

## Changes

- Add optional `memory_id` parameter to `BeamMemory.remember()`.
- Keep existing direct BEAM behavior unchanged by falling back to `_generate_id(content)` when `memory_id` is not supplied.
- Pass the generated legacy ID from `Mnemosyne.remember()` into the BEAM write.
- Strengthen tests to verify:
  - legacy and BEAM rows use the same ID
  - `forget()` removes both layers

## Validation

Ran against current upstream plus this patch:
text python -m pytest tests/test_beam.py -q 26 passed

Also validated through the Hermes Mnemosyne provider integration against this branch:
text MNEMOSYNE_REPO_PATH=/home/kyklis/workspace/Preprod/axdsan-mnemosyne-upstream \ python -m pytest tests/plugins/memory/test_mnemosyne_provider.py -q -o 'addopts=' 6 passed

Live Hermes provider write/recall/delete checks also confirmed that after delete both `memories` and `working_memory` have zero remaining rows for the stored ID.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -672,9 +672,15 @@ class BeamMemory:
     def remember(self, content: str, source: str = "conversation",
                  importance: float = 0.5, metadata: Dict = None,
                  valid_until: str = None, scope: str = "session",
+                 memory_id: str = None,
                  extract_entities: bool = False,
                  extract: bool = False) -> str:
         """Store into working_memory. Deduplicates exact content matches.
+
+        When called from the legacy-compatible Mnemosyne.remember() path,
+        memory_id is passed through so the legacy memories row and BEAM
+        working_memory row stay addressable by the same ID. Direct BEAM calls
+        still generate their own deterministic ID.
 
         Args:
             content: The text to remember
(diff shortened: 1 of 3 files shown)


AxDSan:
Rebased onto main and resolved conflicts in `beam.py` and `memory.py`.

The `remember()` method in both files now carries both your `memory_id` parameter AND the new `extract_entities` / `extract` params from main. No logic changed ,  just merged the signatures and docstrings.

293 tests pass. Authorship preserved. Thanks @Jammu for this fix.

PR #20, opened 2026-04-29
fix: consolidate sleep across inactive sessions

### LINKED ISSUE #19: Automatic sleep consolidation misses old working memories from inactive sessions
## Summary

Mnemosyne's documentation describes "automatic consolidation" / "sleep cycles", but in practice old `working_memory` rows can remain indefinitely if they belong to sessions that are no longer active. The current auto-sleep behavior appears to be opportunistic and session-scoped, not background/idle-based or cross-session.

## Environment

- `mnemosyne-memory`: 1.13.0
- Hermes memory provider path: `~/.hermes/plugins/mnemosyne/__init__.py`
- Active provider: `mnemosyne`
- Storage: local SQLite BEAM DB

## Observed behavior

On my setup:

- `working_memory`: 708 rows
- `episodic_memory`: 3596 rows
- `consolidation_log`: 0 rows
- Old eligible working rows: 184 rows older than the 12h sleep cutoff
- Hindsight shadow sync that reads `episodic_memory` sees no new rows because new memories are still in `working_memory`

I verified that no scheduled job is calling `mnemosyne_sleep`, and no consolidation has been recorded.

## Expected behavior

Given the docs mention automatic consolidation / sleep cycles, I expected old working memories to eventually be summarized and moved to episodic memory without requiring manual intervention for each historical session.

In particular, I expected one of these to happen:

1. background/idle consolidation runs periodically; or
2. the provider's auto-sleep consolidates all eligible old sessions; or
3. there is an explicit built-in `sleep_all_sessions` / maintenance command for scheduled use.

## What the code appears to do today

In the Hermes provider, auto-sleep is enabled and thresholded:

```python
self._auto_sleep_threshold = 50
self._auto_sleep_enabled = True
```

After each synced turn:

```python
self._turn_count += 1
if self._auto_sleep_enabled and self._turn_count % 10 == 0:
    self._maybe_auto_sleep()
```

And `_maybe_auto_sleep()` does:

```python
stats = self._beam.get_working_stats()
working = stats.get("total", 0)
if working > self._auto_sleep_threshold:
    self._beam.sleep()
```

Separately, `on_session_end()` also calls:

```python
self._beam.sleep()
```

However, `BeamMemory.sleep()` is session-scoped:

```sql
WHERE session_id = ? AND timestamp < ?
```

This creates a gap:

1. During an active session, auto-sleep may run every 10 turns, but most rows are still younger than the 12h cutoff, so `sleep()` no-ops.
2. Later, after those rows become older than 12h, that original session may no longer receive turns or session-end hooks.
3. Future sessions calling `sleep()` only consolidate their own `session_id`, not old abandoned sessions.
4. Therefore old `working_memory` rows can remain stuck indefinitely.

There also seems to be a mismatch: `get_working_stats()` counts total working memory across all sessions, but `_beam.sleep()` only consolidates the current session.

## Suggested fix

Add an explicit all-session maintenance primitive in core, for example:

```python
def sleep_all_sessions(self, dry_run: bool = False) -> dict:
    cutoff = datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS // 2)
    session_ids = SELECT DISTINCT session_id
                  FROM working_memory
                  WHERE timestamp < cutoff

    for session_id in session_ids:
        beam = BeamMemory(session_id=session_id, db_path=self.db_path)
        result = beam.sleep(dry_run=dry_run)
        accumulate(result)

    return {
        "sessions_scanned": ...,
        "sessions_consolidated": ...,
        "items_consolidated": ...,
        "summaries_created": ...,
        "errors": ...,
    }
```

Then expose it as either:

- `mnemosyne_sleep_all_sessions`, or
- `mnemosyne_sleep(all_sessions=True)`, or
- `hermes mnemosyne sleep --all-sessions`

The provider's `_maybe_auto_sleep()` could also call this all-session maintenance method when there are eligible old rows, rather than calling session-scoped `sleep()`.

## Why this seems safer than changing `sleep()` directly

I would not suggest changing current `BeamMemory.sleep()` to suddenly cross sessions, because callers may reasonably expect `BeamMemory(session_id=X).sleep()` to only affect session X.

A separate `sleep_all_sessions()` keeps the existing API semantics intact while providing the missing maintenance operation.

## Possible tests

- Create old `working_memory` rows in two different sessions.
- Call current `sleep()` for one session and verify only that session is consolidated.
- Call new `sleep_all_sessions()` and verify both sessions' eligible rows are consolidated.
- Verify `consolidation_log` has entries.
- Verify episodic summaries are vectorized when embeddings/sqlite-vec are available.

## Docs clarification

If the intended behavior is only manual/opportunistic sleep, the docs should probably clarify that "automatic consolidation" does not mean a background idle scheduler, and that users should schedule a sleep maintenance job if they want deterministic long-term consolidation.

wysie:
## Summary

Fixes #19.

This adds an explicit all-session sleep maintenance path so old working memories from inactive sessions do not remain stranded indefinitely.

Changes:
- Keep `BeamMemory.sleep()` session-scoped for backwards-compatible semantics.
- Add `BeamMemory.sleep_all_sessions(dry_run=False)` to consolidate eligible old `working_memory` rows across all sessions.
- Expose the new behavior through:
  - `Mnemosyne.sleep_all_sessions()`
  - global `sleep_all_sessions()` helper
  - Hermes MemoryProvider `mnemosyne_sleep(all_sessions=true)`
  - Hermes plugin/MCP sleep handlers via `all_sessions`
  - CLI: `hermes mnemosyne sleep --all-sessions`
- Update provider auto-sleep to use all-session maintenance when available.
- Add tests covering:
  - `sleep()` remains session-scoped
  - `sleep_all_sessions()` consolidates inactive sessions
  - dry-run preserves working memory
- Clarify README/provider docs.

## Why

The current auto-sleep hook can run only while a session is active, while `BeamMemory.sleep()` only consolidates the current `session_id`. If rows are still younger than the sleep cutoff during active use, they can become eligible only after the session is inactive, then never be touched again.

`sleep_all_sessions()` provides a deterministic maintenance primitive without changing the existing `sleep()` API behavior.

## Test plan

- `python -m pytest tests/test_beam.py::TestSleepCycle -q`
- `python -m compileall hermes_memory_provider hermes_plugin mnemosyne`

Note: I also ran `python -m pytest tests/test_beam.py -q` locally. One pre-existing environment-sensitive test failed on my machine: `TestCrossSessionRecall::test_tools_session_singleton_updates` with a duplicate legacy memory id from the local default DB path. The new `TestSleepCycle` tests pass cleanly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ hermes mnemosyne stats     # Shows working + episodic memory counts
 
 - **BEAM architecture** — Three tiers: hot working memory, long-term episodic memory, temporary scratchpad
 - **Hybrid search** — 50% vector similarity + 30% FTS5 rank + 20% importance, all inside SQLite
-- **Automatic consolidation** — Old working memories are summarized and moved to episodic memory via `mnemosyne_sleep()`
+- **Automatic consolidation** — Old working memories are summarized and moved to episodic memory via `mnemosyne_sleep()`; use `hermes mnemosyne sleep --all-sessions` for scheduled maintenance across inactive sessions
 - **Temporal triples** — Time-aware knowledge graph with automatic invalidation
 - **Export / import** — Move your entire memory database to a new machine with one JSON file
 - **Cross-session scope** — `remember(..., scope="global")` makes facts visible everywhere
@@ -276,8 +276,11 @@ hermes mnemosyne stats --global
 # Search memories
 hermes mnemosyne inspect "dark mode preferences"
 
-# Run consolidation (compress old working memory into episodic summaries)
+# Run consolidation for the current session (compress old working memory into episodic summaries)
 hermes mnemosyne sleep
+
+# Run maintenance consolidation across all sessions with eligible old working memory
+hermes mnemosyne sleep --all-sessions
 
 # Export all memories to a JSON file
 hermes mnemosyne export --output mnemosyne_backup.json
--- a/hermes_memory_provider/README.md
+++ b/hermes_memory_provider/README.md
@@ -65,7 +65,7 @@ User plugins take precedence over bundled plugins on name collision.
 |------|---------|
 | `mnemosyne_remember` | Store durable memory with importance, scope, expiry |
 | `mnemosyne_recall` | Hybrid search (50% vector + 30% FTS + 20% importance) |
-| `mnemosyne_sleep` | Consolidate working → episodic memory |
+| `mnemosyne_sleep` | Consolidate working → episodic memory; pass `all_sessions=true` for inactive-session maintenance |
 | `mnemosyne_stats` | Show working + episodic counts |
 | `mnemosyne_invalidate` | Mark memory as expired/superseded |
 | `mnemosyne_triple_add` | Add temporal fact to knowledge graph |
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -106,9 +106,24 @@ SLEEP_SCHEMA = {
     "name": "mnemosyne_sleep",
     "description": (
         "Run the Mnemosyne consolidation cycle. Compresses old working memories "
-        "into episodic summaries. Call after long sessions or when memory feels stale."
+        "into episodic summaries. Call after long sessions or when memory feels stale. "
+        "Set all_sessions=true to consolidate eligible old working memories across inactive sessions."
     ),
-    "parameters": {"type": "object", "properties": {}},
+    "parameters": {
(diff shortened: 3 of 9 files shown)


wysie:
Closing in favor of #21, which is rebased on current main after v2.0.0 and keeps the diff focused on the inactive-session sleep maintenance fix.

wysie:
Superseded by #21, a smaller rebased PR against current main.

PR #28, opened 2026-05-07
feat: add Hindsight memory importer

WXBR:
## Summary
- Add a Hindsight importer for JSON exports and the Hindsight HTTP memory-list API
- Import Hindsight records into `episodic_memory` directly so historical timestamps, source/fact type, session grouping, metadata, scope, and veracity are preserved
- Register the importer in the provider registry and add a simple `mnemosyne import-hindsight <file|url> [bank]` CLI entry point
- Add regression tests for timestamp preservation, episodic-only import, stable duplicate skipping, FTS indexing, and provider-registry usage

## Why
The generic importer path goes through `remember()`, which is correct for new memories but not for migrations: it assigns the current timestamp and writes to working memory. Hindsight migrations need long-term episodic records with their original timestamps and metadata intact.

## Test plan
- `python -m py_compile mnemosyne/core/importers/hindsight.py mnemosyne/core/importers/__init__.py mnemosyne/cli.py`
- `python -m pytest tests/test_importers/test_hindsight.py -q -o 'addopts='`
- `python -m pytest tests/test_importers -q -o 'addopts='`
- `ulimit -n 8192 2>/dev/null || true; python -m pytest tests -q -o 'addopts='`

Full suite result: `400 passed in 24.30s`.

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -158,6 +158,22 @@ def cmd_import(args):
     print(f"Imported {result.get('count', 0)} memories from {args[0]}")
 
 
+def cmd_import_hindsight(args):
+    """Import memories from a Hindsight JSON export or API."""
+    if not args:
+        print("Usage: mnemosyne import-hindsight <file.json|base_url> [bank]")
+        return
+    target = args[0]
+    bank = args[1] if len(args) > 1 else "hermes"
+    mem = _get_memory()
+    from mnemosyne.core.importers.hindsight import import_from_hindsight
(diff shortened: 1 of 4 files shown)


AxDSan:
@WXBR Thank you for this excellent contribution! The Hindsight importer is exactly what we needed for migration workflows — clean architecture, proper timestamp preservation, and thorough test coverage. Really appreciate you taking the time to wire everything up properly (registry, CLI, tests, the works). Merged! 🙏

PR #29, opened 2026-05-07
Add host LLM adapter so Mnemosyne can route through Hermes' authenticated provider

kohai-ut:
## Summary

Adds an opt-in host LLM adapter that lets Mnemosyne route both **consolidation** (`summarize_memories`) and **structured fact extraction** (`extract_facts`) through Hermes' authenticated `agent.auxiliary_client.call_llm(task="compression", ...)` when running inside Hermes — without dragging Hermes' auth into Mnemosyne core.

This unblocks Hermes users on OAuth-backed providers like `openai-codex` (ChatGPT/GPT-5 subscriptions) who today can't use `MNEMOSYNE_LLM_BASE_URL`/`MNEMOSYNE_LLM_API_KEY` because that path expects a plain OpenAI-compatible endpoint and API key, not OAuth.

## Why

`mnemosyne/core/local_llm.py` currently supports a remote OpenAI-compatible URL plus a local GGUF fallback. Hermes' `openai-codex` provider is OAuth/session-backed and uses provider-specific routing, headers, OAuth refresh, and Codex Responses API translation. Reproducing all of that inside Mnemosyne would be a heavy lift; reusing what Hermes already authenticates is dramatically simpler.

## Design

- New `mnemosyne/core/llm_backends.py` exposes a tiny `LLMBackend` Protocol (one method, `complete()`), a `CallableLLMBackend` dataclass for tests, and a process-global setter/getter. **Standalone Mnemosyne is unaffected** — the registry is empty by default.
- New `hermes_memory_provider/hermes_llm_adapter.py` implements `HermesAuxLLMBackend.complete()` that calls `agent.auxiliary_client.call_llm(task="compression", ...)`. `agent.*` is imported only inside the call path, never at module import time.
- `MnemosyneMemoryProvider.initialize()` registers the backend; `shutdown()` unregisters it (after briefly draining any in-flight `on_session_end` daemon thread, so post-timeout host calls can't race the unregister and silently fall through to a stale `MNEMOSYNE_LLM_BASE_URL`).
- `summarize_memories()` and `extract_facts()` consult the host first (when `MNEMOSYNE_HOST_LLM_ENABLED=true`). On host attempt with no usable text, the existing remote URL is **skipped** — falls straight to local GGUF, then `None`/`[]`. This prevents accidentally routing memory content to a stale URL the user forgot to clear.
- `llm_available()` now returns `True` when a host backend is registered, so Hermes-only users (no remote URL, no GGUF) don't get short-circuited by `beam.sleep()`'s availability gate.
- `on_session_end()` now runs sleep in a daemon thread with a 15s join (configurable via `SESSION_END_SLEEP_TIMEOUT_SECONDS`), mirroring the existing `_maybe_auto_sleep()` pattern in the same file. Prevents Hermes shutdown from blocking on a slow LLM call.
- Fact extraction uses `temperature=0.0` for determinism (avoids near-duplicate writes to the facts table on re-ingest); consolidation stays at 0.3.

The interface is general — any host (not just Hermes) can register a backend by implementing `complete()`. Documented for future agents in `docs/hermes-llm-integration.md`.

## Configuration

```bash
# Required: opt in to the host backend.
MNEMOSYNE_HOST_LLM_ENABLED=true

# Optional overrides (Hermes auto-resolution otherwise).
MNEMOSYNE_HOST_LLM_PROVIDER=openai-codex
MNEMOSYNE_HOST_LLM_MODEL=gpt-5.1-mini
MNEMOSYNE_HOST_LLM_N_CTX=32000  # bigger budget for Codex/GPT-class models
```

When `MNEMOSYNE_HOST_LLM_ENABLED` is unset/false, behavior is identical to before this PR.

## Verification

- **442 → 449 tests pass.** 7 new tests covering registry round-trip, host-route precedence, the host-skips-remote rule, gate semantics, shutdown drain race, daemon-thread exception logging, bullet-list output preservation, and remote-path determinism contract.
- Pre-existing standalone tests in `test_local_llm.py` and `test_extraction.py` still pass — proves backwards compatibility.
- **Live end-to-end verified**: a real `call_llm(task='compression')` round-trip through an `openai-codex` OAuth subscription completed successfully:
  - `agent.auxiliary_client` resolved provider → `openai-codex (gpt-5.5)`
  - `POST https://chatgpt.com/backend-api/codex/responses 200 OK`
  - Hermes' `extract_content_or_reasoning()` correctly extracted text from the Codex Responses API shape
  - Both plain-prose and bulleted-list responses parsed correctly post-fix
- Adversarial review pass via independent AI surfaced 7 findings; all addressed in commit `6ade86f` with regression tests.

## Test plan

- [x] Standalone Mnemosyne (no Hermes): `MNEMOSYNE_HOST_LLM_ENABLED` unset → existing remote/local/AAAK chain runs unchanged
- [x] Hermes user with `auxiliary.compression.provider: auto` and main provider `openai-codex` → consolidation + extraction both route through OAuth
- [x] Hermes user with explicit `MNEMOSYNE_HOST_LLM_PROVIDER`/`_MODEL` overrides → those propagate to `call_llm` correctly
- [x] Host attempt fails (import error, exception, empty response) → falls to local GGUF, **never** to `MNEMOSYNE_LLM_BASE_URL`
- [x] `on_session_end()` with a slow LLM → returns within join timeout, daemon continues in background, Hermes shutdown not blocked
- [x] `shutdown()` while `on_session_end` thread still running → drains briefly so the late host call doesn't race with backend unregister
- [x] `MNEMOSYNE_LLM_ENABLED=false` → disables every LLM-backed memory operation including host
- [x] Bulleted host response (Codex's natural output for fact lists) → parses correctly into individual facts

## Files

| Area | Files |
|---|---|
| New core module | `mnemosyne/core/llm_backends.py` |
| Modified core | `mnemosyne/core/local_llm.py`, `mnemosyne/core/extraction.py` |
| New adapter | `hermes_memory_provider/hermes_llm_adapter.py` |
| Modified provider | `hermes_memory_provider/__init__.py` |
| New tests | `tests/test_llm_backends.py`, `tests/test_hermes_llm_adapter.py`, `tests/test_hermes_memory_provider.py` |
| Modified tests | `tests/conftest.py`, `tests/test_local_llm.py`, `tests/test_extraction.py`, `tests/test_extraction_integration.py` |
| Docs | `docs/hermes-llm-integration.md` (new), `docs/hermes-integration.md` (link added) |
| Design doc | `docs/plans/2026-05-07-hermes-aux-llm-adapter.md` (new — capture decision log; happy to remove if you'd rather not have it in-tree) |

## Backwards compatibility

100%. The host adapter is opt-in via env var and the registry is empty by default. Existing standalone tests pass unmodified.


AxDSan:
This is a really clean piece of work. The protocol-based registry keeps Mnemosyne core Hermes-free, the A3 fall-through rule protects users from stale URL leaks, and the bounded shutdown/thread drain shows you thought through the race conditions.

The live verification against openai-codex OAuth gives me confidence this works in the wild, not just in tests. 448 tests passing with zero regressions is exactly the bar we want.

Thank you for the contribution -- and for the thorough PR description that made review straightforward.

Merged.

AxDSan (commented):
Review done, PR merged

PR #31, opened 2026-05-08
fix: support base install without numpy

kohai-ut:
## Summary

Fixes base installs of `mnemosyne-memory` when optional dense-retrieval dependencies are not installed.

A clean base install currently succeeds, and `mnemosyne --help` works, but normal memory commands such as `mnemosyne stats` fail immediately with:

```text
ModuleNotFoundError: No module named 'numpy'
```

The failure happens before Mnemosyne can fall back to keyword/FTS behavior because `numpy` is imported at module import time.

## Why this is a packaging contract issue

The README documents base install as:

```bash
pip install mnemosyne-memory
```

and documents `[all]` separately for dense retrieval / local LLM features:

```bash
pip install mnemosyne-memory[all]
```

The package description also emphasizes a zero/minimal-dependency base path. Given that contract, missing `numpy` / `fastembed` should disable dense/vector functionality rather than making core CLI commands fail.

## Changes

- Treat `numpy` as optional in `mnemosyne.core.embeddings`.
- Treat `fastembed` separately as optional, so dense embeddings are available only when both `numpy` and `fastembed` import successfully.
- Treat `numpy` as optional in `mnemosyne.core.beam`.
- Return no in-memory vector fallback results when `numpy` is unavailable.
- Add regression tests that simulate missing `numpy` and `fastembed` via subprocess import blocking.

## Test Plan

Observed the regression first: the new tests failed with `ModuleNotFoundError: No module named 'numpy'` from `embeddings.py`, then from `beam.py`.

After the fix:

```text
.venv/bin/python -m pytest tests/test_optional_embeddings.py -q
# 2 passed
```

```text
.venv/bin/python -m pytest tests/test_optional_embeddings.py tests/test_mnemosyne_stats.py tests/test_beam.py -q
# 88 passed
```

```text
.venv/bin/python -m pytest -q
# 454 passed
```

Also verified the actual packaging path in a fresh venv:

```bash
python3.11 -m venv /tmp/mnemosyne-base-install-fixed
/tmp/mnemosyne-base-install-fixed/bin/python -m pip install .
MNEMOSYNE_DATA_DIR=/tmp/mnemosyne-base-probe-fixed \
  /tmp/mnemosyne-base-install-fixed/bin/mnemosyne stats
```

Result:

```text
Mnemosyne Stats

  Total memories: 0
  Working memory: 0
  Episodic memory: 0
  DB path: N/A
```

Note: the full suite exits 0, but this environment emits an unrelated ignored `llama_cpp.Llama.__del__` cleanup traceback after the pytest summary.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -13,15 +13,21 @@ FTS5 for full-text retrieval.
 Hybrid ranking: 50% vector + 30% FTS rank + 20% importance.
 """
 
+from __future__ import annotations
+
 import sqlite3
 import json
 import hashlib
 import threading
 import math
-import numpy as np
 from datetime import datetime, timedelta
 from typing import List, Dict, Optional, Any, Set, Union
 from pathlib import Path
+
+try:
+    import numpy as np
+except ImportError:
+    np = None
 
 from mnemosyne.core import embeddings as _embeddings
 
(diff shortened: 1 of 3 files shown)


AxDSan:
Approved for merge. Clean fix, no conflicts, touches optional dependency paths only.

PR #34, opened 2026-05-08
chore: add mcp optional dependency metadata

kohai-ut:
## Summary

This is a small packaging metadata parity fix for the optional MCP integration.

The code and `setup.py` already describe an install path for MCP support:

```bash
pip install mnemosyne-memory[mcp]
```

but `pyproject.toml` did not currently define a matching `mcp` optional dependency group. This PR adds that missing extra to `pyproject.toml` and includes the MCP dependencies in the existing `all` extra.

## What changed

In `pyproject.toml`:

```toml
mcp = ["mcp>=1.0.0", "anyio>=4.0"]
```

and `all` now also includes:

```toml
"mcp>=1.0.0", "anyio>=4.0"
```

## Why

`mnemosyne/mcp_server.py` raises this install guidance when MCP is unavailable:

```python
RuntimeError("MCP not installed. Run: pip install mnemosyne-memory[mcp]")
```

`setup.py` already has the same optional extra:

```python
"mcp": ["mcp>=1.0.0", "anyio>=4.0"],
"all": [..., "mcp>=1.0.0", "anyio>=4.0"],
```

So this PR makes the modern package metadata in `pyproject.toml` agree with the existing code and `setup.py` contract.

## Scope

This PR intentionally does not change runtime behavior.

No code paths changed. No CLI behavior changed. No docs changed. This only updates packaging metadata so the documented/code-referenced extra exists in `pyproject.toml` too.

## Verification performed

I verified the TOML parses and that:

- `[project.optional-dependencies].mcp` exists
- it equals `['mcp>=1.0.0', 'anyio>=4.0']`
- both MCP dependencies are included in the `all` extra
- `setup.py` already contains the matching `mcp` extra

Commands/checks run locally:

```bash
python3 - <<'PY'
import tomllib
from pathlib import Path
pyproject = tomllib.loads(Path('pyproject.toml').read_text())
optional = pyproject['project']['optional-dependencies']
assert optional['mcp'] == ['mcp>=1.0.0', 'anyio>=4.0']
for dep in optional['mcp']:
    assert dep in optional['all']
print('pyproject optional dependencies OK')
PY

git diff --check
```

`git diff --check` reported no whitespace issues.

## Relationship to other PRs

This is intentionally separate from the docs-only PR that updates the README/API docs to refer to MCP instead of a stale REST server. This PR is the packaging follow-up: if users see `mnemosyne-memory[mcp]`, the extra should exist in `pyproject.toml` as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,8 +39,8 @@ classifiers = [
 [project.optional-dependencies]
 llm = ["ctransformers>=0.2.27", "llama-cpp-python>=0.2.0", "huggingface-hub>=0.20"]
 embeddings = ["fastembed>=0.3.0"]
-mcp = ["mcp>=1.0.0", "anyio>=4.0"]
-all = ["ctransformers>=0.2.27", "llama-cpp-python>=0.2.0", "huggingface-hub>=0.20", "fastembed>=0.3.0", "mcp>=1.0.0", "anyio>=4.0"]
+mcp = ["mcp>=1.0.0; python_version >= '3.10'", "anyio>=4.0; python_version >= '3.10'"]
+all = ["ctransformers>=0.2.27", "llama-cpp-python>=0.2.0", "huggingface-hub>=0.20", "fastembed>=0.3.0", "mcp>=1.0.0; python_version >= '3.10'", "anyio>=4.0; python_version >= '3.10'"]
 dev = ["pytest>=7.0", "build", "twine"]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,8 @@ classifiers = [
 [project.optional-dependencies]
 llm = ["ctransformers>=0.2.27", "llama-cpp-python>=0.2.0", "huggingface-hub>=0.20"]
 embeddings = ["fastembed>=0.3.0"]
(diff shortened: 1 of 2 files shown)


kohai-ut:
CI caught an important compatibility issue here.

The first version of this PR added `mcp>=1.0.0` directly to the new `mcp` extra and to `all`. That caused the Python 3.9 job to fail during dependency installation because current `mcp` releases require Python >=3.10:

```text
ERROR: Could not find a version that satisfies the requirement mcp>=1.0.0; extra == "all"
```

I pushed a follow-up commit that keeps the extra available but gates the MCP dependencies with Python-version markers:

```toml
mcp = [
  "mcp>=1.0.0; python_version >= '3.10'",
  "anyio>=4.0; python_version >= '3.10'",
]
```

and applies the same markers inside `all`.

I also updated `setup.py` with the same markers so the legacy and modern packaging metadata stay aligned.

Local verification before pushing the fix:

- parsed `pyproject.toml`
- checked the markers evaluate false on Python 3.9 and true on Python 3.10
- ran `pip install -e '.[all,dev]' --dry-run` inside a fresh local venv
- ran `git diff --check`

This should preserve Python 3.9 base/package CI while allowing MCP installs on Python versions that MCP supports.

AxDSan:
Approved for merge. Packaging metadata fix, no runtime impact.

PR #35, opened 2026-05-08
fix: llm_used detection always False + remote fallback blocks local LLM + Llama token prompt

ether-btc:
### 1. `llm_used` metadata always `False` in sleep consolidation

**File:** `mnemosyne/core/beam.py` (sleep method)

The `llm_used` metadata field was computed as:
```python
"llm_used": summary != f"[{source}] {aaak_encode(' | '.join(lines))}"
```

This comparison always evaluates `True` when LLM succeeds (LLM output never matches AAAK compressed format), making `llm_used` persistently `False` even when LLM summarization fires.

**Fix:** Use an explicit `llm_succeeded` boolean flag set `True` when LLM produces a summary.

### 2. Remote API failure blocks local LLM fallback

**File:** `mnemosyne/core/local_llm.py` (`summarize_memories()`)

When `MNEMOSYNE_LLM_BASE_URL` was set but the remote call failed (timeout, 401, etc.), `summarize_memories()` returned `None` immediately — the local llama-cpp-python backend was never reached.

**Fix:** Remote failure now falls through to local LLM. Added `MNEMOSYNE_FORCE_LOCAL=1` env var to skip remote API entirely (useful when OpenRouter quota is exhausted).

### 3. Llama chat-template tokens in prompt

**File:** `mnemosyne/core/local_llm.py` (`_build_prompt()`)

The prompt used Llama-specific tokens (`tiert`, `</s>`, ` ins`) which:
- Get echoed verbatim by some GGUF models
- Break with non-Llama providers

**Fix:** Replaced with plain-text instruction format that works with any model.

## Testing

- Verified `llm_available()` returns `True` with TinyLlama 1.1B Q4_K_M on ARM64
- Tested `summarize_memories()` end-to-end: produces coherent 1-2 sentence summaries
- All 113 previously AAAK-encoded episodic memories re-consolidated successfully with local LLM

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -2131,6 +2131,7 @@ class BeamMemory:
 
             # --- Try LLM summarization (chunked to fit context) ---
             summary = None
+            llm_succeeded = False
             if local_llm.llm_available():
                 chunks = local_llm.chunk_memories_by_budget(lines, source=source)
                 if chunks:
@@ -2158,6 +2159,7 @@ class BeamMemory:
                                     summary = " | ".join(chunk_summaries)
                     if summary:
                         llm_used_count += 1
+                        llm_succeeded = True
 
             # --- Fallback to aaak encoding ---
             if summary is None:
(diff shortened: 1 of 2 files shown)


AxDSan:
Approved for merge. Core fix but no conflicts with main. Touches BEAM and local_llm only.

PR #37, opened 2026-05-08
fix: support Hindsight imports from Hermes CLI

kohai-ut:
# fix: support Hindsight imports from Hermes CLI

## Summary

This PR expands the Hindsight migration work from PR #28 by exposing the same timestamp-preserving Hindsight importer through the Hermes CLI import command.

PR #28 added the correct migration path for Hindsight data: `HindsightImporter` reads Hindsight JSON/API records and writes them directly to Mnemosyne `episodic_memory`, preserving historical fields such as `mentioned_at`, `fact_type`, session grouping, metadata, source, scope, and veracity. That behavior is intentionally different from the generic importer base path, which routes records through `remember()` and would assign current timestamps / working-memory semantics.

Before this PR, the standalone Mnemosyne CLI could use that logic via:

```bash
mnemosyne import-hindsight <file|url> [bank]
```

but the Hermes CLI provider import path could not. `hermes mnemosyne import --from <provider>` treated every provider as an API-key provider, so `--from hindsight` was blocked by the generic `--api-key required` check and did not expose the Hindsight-specific `file_path` / `bank` inputs.

This PR fixes that gap by making Hermes CLI route `--from hindsight` into the existing PR #28 importer instead of duplicating or replacing its timestamp logic.

### Hermes CLI

Adds Hindsight-specific provider inputs to `hermes mnemosyne import`:

```bash
hermes mnemosyne import --from hindsight --file hindsight-export.json --bank hermes
hermes mnemosyne import --from hindsight --input hindsight-export.json --bank hermes
hermes mnemosyne import --from hindsight --base-url http://localhost:8888 --bank hermes
```

Implementation detail:

```python
result = import_from_provider(
    "hindsight",
    mem,
    file_path=file_path,
    base_url=base_url,
    bank=bank,
    dry_run=dry_run,
    session_id=session_id,
    channel_id=channel_id,
)
```

That means Hermes CLI now relies on the existing `HindsightImporter` registered by PR #28. It does not copy the importer, reimplement timestamp selection, or write directly to SQLite itself.

### Behavior

- `--from hindsight` no longer requires `--api-key`.
- Hindsight JSON exports can be passed via either `--file` or existing `--input`.
- Live Hindsight API imports can be passed via `--base-url`.
- `--bank` is forwarded to the Hindsight importer and defaults to `hermes`, matching PR #28.
- Hindsight imports continue to preserve historical timestamps because the real import still happens in `HindsightImporter`.
- Other providers still use the existing API-key flow unchanged.

### Docs

This PR also updates the Hermes-facing docs that were missing the new Hindsight CLI path:

- `docs/hermes-integration.md`
- `docs/api-reference.md`

The README already documented the `--from hindsight` Hermes examples; this PR brings the rest of the docs into alignment.

## Why this approach

The important design choice is not to generalize timestamp preservation across all importers.

Most providers still fit the generic importer model: transform provider records and store them through `remember()`. Hindsight is different because it is a historical migration of already-consolidated memory. Routing Hindsight through `remember()` would lose the original memory timeline and contaminate working memory with imported history.

PR #28 already encoded that distinction correctly. This PR keeps that architecture intact and simply exposes it through Hermes CLI.

Layering after this PR:

```text
Hermes CLI
  └── import_from_provider("hindsight", ...)
        └── HindsightImporter from PR #28
              └── direct episodic_memory insert preserving timestamps
```

## Tests

Added Hermes CLI regression tests covering:

- Hindsight file import via `--file` preserves `mentioned_at` in `episodic_memory`.
- Hindsight file import via `--input` also works.
- Hindsight `--base-url` path does not require an API key and forwards `base_url` / `bank`.
- Non-default `--bank` is forwarded into imported metadata.
- `--dry-run` returns success but writes no working or episodic rows.
- Missing `--file` / `--input` / `--base-url` gives a clear error.
- Non-Hindsight providers still require `--api-key`.

Verification commands run:

```bash
uv run --frozen --with pytest --with numpy python -m pytest tests/test_importers/test_hindsight.py tests/test_hermes_hindsight_cli.py -q
# 10 passed

uv run --frozen --with pytest --with numpy python -m pytest tests/test_importers tests/test_hermes_hindsight_cli.py -q
# 35 passed
```

## Review notes

Claude Code reviewed the implementation plan and final diff read-only. `gstack /review` was requested, but `gstack` is not installed in this environment, so Claude Code read-only diff review was used instead.

Claude's plan review suggested adding coverage for `--base-url`, non-default bank forwarding, and dry-run behavior; those tests were added before final verification.

## Non-goals

- Does not replace `mnemosyne import-hindsight`.
- Does not duplicate `HindsightImporter` logic in Hermes CLI.
- Does not change `BaseImporter`, `Mnemosyne.remember()`, or generic provider import semantics.
- Does not change API-key requirements for Mem0/Zep/Honcho/etc.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -654,6 +654,11 @@ mnemosyne export backup.json
 mnemosyne import backup.json
 mnemosyne import-hindsight export.json [bank]      # Import Hindsight JSON
 mnemosyne import-hindsight http://localhost:8888    # Import from live API
+
+# Hermes CLI routes through the same HindsightImporter, preserving timestamps
+hermes mnemosyne import --from hindsight --file export.json --bank hermes
+hermes mnemosyne import --from hindsight --input export.json --bank hermes
+hermes mnemosyne import --from hindsight --base-url http://localhost:8888 --bank hermes
 mnemosyne bank list
 mnemosyne bank create work
 mnemosyne bank delete work
--- a/docs/hermes-integration.md
+++ b/docs/hermes-integration.md
@@ -80,6 +80,12 @@ hermes mnemosyne inspect "query"    # Search memories
 hermes mnemosyne sleep              # Run consolidation
 hermes mnemosyne export --output backup.json
 hermes mnemosyne import --input backup.json
+
+# Import historical Hindsight memories via PR #28's timestamp-preserving importer
+hermes mnemosyne import --from hindsight --file hindsight-export.json --bank hermes
+hermes mnemosyne import --from hindsight --input hindsight-export.json --bank hermes
+hermes mnemosyne import --from hindsight --base-url http://localhost:8888 --bank hermes
+
 hermes mnemosyne clear              # Clear scratchpad
 hermes mnemosyne version            # Show version
 ```
--- a/hermes_memory_provider/cli.py
+++ b/hermes_memory_provider/cli.py
@@ -37,12 +37,14 @@ def register_cli(subparser):
 
     import_cmd = mn_cmds.add_parser("import", help="Import memories from a JSON file or another provider")
     import_cmd.add_argument("--input", "-i", type=str, help="Input JSON file path (for file imports)")
+    import_cmd.add_argument("--file", type=str, help="Provider file input, e.g. Hindsight JSON export")
     import_cmd.add_argument("--force", action="store_true", help="Overwrite existing records (file import)")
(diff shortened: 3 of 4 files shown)


AxDSan:
Approved for merge. Builds on merged PR #28. No conflicts. CLI extension only.

PR #47, opened 2026-05-09
fix: report CLI import and export counts

kohai-ut:
## Summary

Fixes a misleading success message from the public `mnemosyne export` CLI command.

Before this change, `mnemosyne export <path>` could successfully write an export file containing working memories, legacy memories, and triples, but still print:

```text
Exported 0 memories to <path>
```

That output was wrong. The export implementation already returns detailed count metadata, but the CLI was reading a nonexistent generic `count` key and defaulting to `0`.

This PR updates the CLI to report the actual exported record counts returned by `Mnemosyne.export_to_file()`.

## User-visible bug

A normal CLI path produced contradictory results:

```bash
tmp=$(mktemp -d)
MNEMOSYNE_DATA_DIR="$tmp" python -m mnemosyne.cli store 'export probe memory' probe 0.7
MNEMOSYNE_DATA_DIR="$tmp" python -m mnemosyne.cli export "$tmp/export.json"
python - <<PY
import json, pathlib
p = pathlib.Path('$tmp/export.json')
data = json.loads(p.read_text())
print({k: len(data.get(k, [])) for k in ['working_memory', 'episodic_memory', 'legacy_memories', 'triples', 'scratchpad']})
PY
```

Observed before the fix:

```text
Stored: 49d6f699c37f1de8
Exported 0 memories to /tmp/.../export.json
{'working_memory': 1, 'episodic_memory': 0, 'legacy_memories': 1, 'triples': 2, 'scratchpad': 0}
```

The export file was correct. The CLI message was not.

## Root cause

`Mnemosyne.export_to_file()` returns structured metadata:

- `working_memory_count`
- `episodic_memory_count`
- `scratchpad_count`
- `legacy_memories_count`
- `triples_count`

The CLI used:

```python
result.get("count", 0)
```

That key is not returned by `export_to_file()`, so the CLI always fell back to `0` even when records were exported.

## Fix

The export command now formats the detailed counts already returned by the export layer:

```text
Exported 1 working, 0 episodic, 1 legacy, 2 triples to <path>
```

This keeps the fix at the CLI boundary and does not change export file structure or storage behavior.

## Regression test

Added `test_export_reports_actual_exported_memory_counts`, which:

1. runs the public CLI through a subprocess;
2. stores one memory in an isolated `MNEMOSYNE_DATA_DIR`;
3. exports to a temp JSON file;
4. asserts the CLI reports the real counts instead of `0`; and
5. verifies the exported JSON contains the expected working, legacy, and triple records.

The test failed before the production change with:

```text
AssertionError: assert 'Exported 1 working, 0 episodic, 1 legacy, 2 triples' in 'Exported 0 memories to .../export.json\n'
```

## Scope / non-goals

- No export schema changes.
- No import behavior changes.
- No storage or BEAM logic changes.
- No attempt to redefine whether triples should be included in a total memory count. The CLI now reports the separate categories the export API already exposes, avoiding ambiguity.

## Verification

Targeted regression test before fix:

```text
FAILED tests/test_cli_errors.py::test_export_reports_actual_exported_memory_counts
```

Targeted tests after fix:

```text
3 passed in 0.56s
```

Full suite:

```text
464 passed, 1 warning in 6.77s
```

Manual smoke after fix:

```text
Stored: 1c0969bd6546a2be
Exported 1 working, 0 episodic, 1 legacy, 2 triples to /tmp/.../export.json
{'working_memory': 1, 'episodic_memory': 0, 'legacy_memories': 1, 'triples': 2, 'scratchpad': 0}
```

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -192,7 +192,15 @@ def cmd_import(args):
         _fail(f"Invalid JSON in import file {args[0]}: {e}")
     except ValueError as e:
         _fail(str(e))
-    print(f"Imported {result.get('count', 0)} memories from {args[0]}")
+    beam_stats = result.get("beam", {})
+    print(
+        "Imported "
+        f"{beam_stats.get('working_memory', {}).get('inserted', 0)} working, "
+        f"{beam_stats.get('episodic_memory', {}).get('inserted', 0)} episodic, "
+        f"{result.get('legacy', {}).get('inserted', 0)} legacy, "
+        f"{result.get('triples', {}).get('inserted', 0)} triples "
+        f"from {args[0]}"
+    )
 
 
 def cmd_import_hindsight(args):
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -168,7 +168,14 @@ def cmd_export(args):
     output_path = args[0] if args else os.path.join(DATA_DIR, "mnemosyne_export.json")
     mem = _get_memory()
     result = mem.export_to_file(output_path)
-    print(f"Exported {result.get('count', 0)} memories to {output_path}")
+    print(
+        "Exported "
+        f"{result.get('working_memory_count', 0)} working, "
+        f"{result.get('episodic_memory_count', 0)} episodic, "
+        f"{result.get('legacy_memories_count', 0)} legacy, "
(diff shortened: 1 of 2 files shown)


AxDSan:
Merged manually onto main. Conflict resolved.

PR #52, opened 2026-05-09
fix: honor data dir changes at runtime

kohai-ut:
## Summary

Fixes a runtime data-directory isolation bug where `MNEMOSYNE_DATA_DIR` changes made after module import were ignored by the default memory and bank paths.

This came up while running the full suite locally: `TestMnemosyneBankIsolation.test_data_isolation_between_banks` set `MNEMOSYNE_DATA_DIR` to a temporary directory, but `Mnemosyne(bank="default")` still used the module-level `DEFAULT_DB_PATH` computed at import time. On a machine with real Mnemosyne data under the default home directory, recall could return unrelated global memories ahead of the test memory, causing the isolation assertion to fail.

## Root cause

Several modules computed default paths only once at import time:

- `mnemosyne.core.memory.DEFAULT_DB_PATH`
- `mnemosyne.core.beam.DEFAULT_DB_PATH`
- `mnemosyne.core.banks.DEFAULT_DATA_DIR`

That works only when `MNEMOSYNE_DATA_DIR` is present before import. It breaks callers/tests that set or change the environment later in the same Python process.

The mismatch was especially visible for banks:

- `BankManager(data_dir)` could create a bank under a temporary data directory.
- `Mnemosyne(bank="default")` still used the import-time default home DB.
- `Mnemosyne(bank="personal")` used `BankManager()` without an explicit `data_dir`, which also used its import-time default.

So the test thought it had created two isolated temp-bank instances, but the actual default/non-default bank instances could point at the user's real data directory instead.

## Fix

Adds runtime path helpers and uses them when no explicit path is supplied:

- `mnemosyne.core.memory._default_data_dir()` / `_default_db_path()`
- `mnemosyne.core.beam._default_data_dir()` / `_default_db_path()`
- `mnemosyne.core.banks._default_data_dir()`

The public constants remain for compatibility, but constructors and default connection helpers now re-read `MNEMOSYNE_DATA_DIR` at the point they resolve an implicit path.

Explicit paths still win:

- `Mnemosyne(db_path=...)` remains unchanged.
- `BankManager(data_dir=...)` remains unchanged.
- `BeamMemory(db_path=...)` remains unchanged.

## Test cleanup adjustment

After the production fix, entity integration tests that set `MNEMOSYNE_DATA_DIR` correctly write module-level `remember()` data into the temp directory. Their teardown previously removed only `test_*.db*`, which left the real default `mnemosyne.db*` files behind inside the temp directory. The teardown now removes the whole temporary directory.

## TDD / verification

RED:

```text
uv run pytest tests/test_memory_banks.py::TestMnemosyneBankIsolation::test_data_isolation_between_banks -q
FAILED
```

The failure reproduced because `Mnemosyne(bank="default")` was using the import-time default DB rather than the temp directory set by the test.

GREEN:

```text
uv run pytest tests/test_memory_banks.py::TestMnemosyneBankIsolation::test_data_isolation_between_banks -q
1 passed

uv run pytest tests/test_entity_integration.py -q
8 passed

uv run pytest tests/test_memory_banks.py -q
26 passed

uv run pytest -q
463 passed, 1 warning
```

The remaining warning is the existing MCP `asyncio.get_event_loop()` deprecation warning.

## Non-goals

- Does not change the default data location when `MNEMOSYNE_DATA_DIR` is unset.
- Does not migrate existing DB files.
- Does not change explicit `db_path` / `data_dir` behavior.
- Does not remove the module-level constants, to avoid breaking imports that may reference them.

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -13,9 +13,8 @@ from pathlib import Path
 from typing import NoReturn
 
 # Data directory — respects MNEMOSYNE_DATA_DIR env var
-DATA_DIR = os.environ.get(
-    "MNEMOSYNE_DATA_DIR",
-    str(Path.home() / ".hermes" / "mnemosyne" / "data"),
+DATA_DIR = os.environ.get("MNEMOSYNE_DATA_DIR") or str(
+    Path.home() / ".hermes" / "mnemosyne" / "data"
 )
 os.makedirs(DATA_DIR, exist_ok=True)
 
--- a/mnemosyne/core/banks.py
+++ b/mnemosyne/core/banks.py
@@ -36,6 +36,13 @@ if os.environ.get("MNEMOSYNE_DATA_DIR"):
     BANKS_DIR = DEFAULT_DATA_DIR / "banks"
 
 
+def _default_data_dir() -> Path:
+    """Return the current default data directory, honoring runtime env changes."""
+    if os.environ.get("MNEMOSYNE_DATA_DIR"):
+        return Path(os.environ["MNEMOSYNE_DATA_DIR"])
+    return DEFAULT_DATA_DIR
+
+
 class BankManager:
     """
     Manage named memory banks.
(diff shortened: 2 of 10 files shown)


kohai-ut:
This data path clarification started as a single fix and then mushroomed the same change in multiple files.  I've tried to keep them here in PR 51 and 52.  They need a close look to see if this is what you want as a maintainer in the big picture.

AxDSan:
> This data path clarification started as a single fix and then mushroomed the same change in multiple files. I've tried to keep them here in PR 51 and 52. They need a close look to see if this is what you want as a maintainer in the big picture.

ill check em out ty!

PR #68, opened 2026-05-10
fix(binary_vectors): add module-level aliases + FTS5 phrase-to-OR fallback

steezkelly:
## Summary

Two fixes for Mnemosyne v2.5:

### Fix 1: binary_vectors.py — module-level aliases
 and  were class static methods only.  imports them as module-level functions, silently getting . This broke NAI-4 binary vector compression and NAI-5 graph bonus scoring on every load.

### Fix 2: beam.py — FTS5 phrase-to-OR fallback
In BEAM mode, when a multi-word FTS5 phrase query returns 0 results, fall back to an OR query of individual content words. Handles queries where the exact phrase isn't in any single memory entry but constituent words exist.

## Impact
- Fixes binary vector storage at consolidation (NAI-4)
- Fixes NAI-5 graph edge + fact bonus scoring
- Improves recall for phrase queries where no entry contains the full phrase

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -893,6 +893,17 @@ def _fts_search_working(conn: sqlite3.Connection, query: str, k: int = 20) -> Li
         "SELECT id, rank FROM fts_working WHERE fts_working MATCH ? ORDER BY rank LIMIT ?",
         (fts_query, k)
     ).fetchall()
+
+    # BEAM mode: if phrase query returns 0, fall back to individual word OR search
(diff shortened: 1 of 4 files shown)


steezkelly:
Update: Added another fix — FTS5 phrase-to-OR fallback in _fts_search_working. When a multi-word phrase query returns 0 results in BEAM mode, it now falls back to an OR query of individual content words. This handles queries where the exact phrase isn't in any single entry but constituent words exist in different memories.

PR #70, opened 2026-05-11
Fix: TripleStore silently invalidates multi-valued annotations (E6)

kohai-ut:
## Summary

`TripleStore.add()` auto-invalidates rows with matching `(subject, predicate)` regardless of `object`. Every production write uses annotation semantics (`(memory_id, "mentions", entity)`, `(memory_id, "fact", text)`, `(memory_id, "occurred_on", date)`, `(memory_id, "has_source", source)`), so each new annotation for a memory silently sets `valid_until` on prior annotation rows sharing the key. Every Mnemosyne database in the field has accumulated `valid_until` on entity / fact annotations any time a memory had more than one extracted.

Production reads happen to mask the data loss (`query_by_predicate` doesn't filter by `valid_until`), but the semantics are wrong and any consumer filtering by validity loses data. This PR fixes the bug by splitting storage into two purpose-specific tables.

## What changes

**`mnemosyne/core/triples.py`** retains current-truth semantics with auto-invalidation, suitable for facts like `(user, prefers, X)` later superseded by `(user, prefers, Y)`. `TripleStore.add_facts()` is deprecated — emits `DeprecationWarning` and routes writes to the new store.

**`mnemosyne/core/annotations.py`** (new) — `AnnotationStore` is append-only, with a `UNIQUE(memory_id, kind, value)` index + `INSERT OR IGNORE` so concurrent ingest and migration are idempotent. Hosts `mentions`, `fact`, `occurred_on`, `has_source` — all multi-valued by design.

**Production call sites migrated** (`beam.py` x5, `memory.py` x2) to write and read through `AnnotationStore`. `BeamMemory` caches a shared `AnnotationStore` instance (`self.annotations`) that reuses the thread-local connection — no per-call file-descriptor cost.

**`mnemosyne/migrations/e6_triplestore_split.py`** (new package directory) — idempotent, transactional migration with file-level backup. Auto-runs on first `BeamMemory.__init__` against a pre-E6 database. `MNEMOSYNE_AUTO_MIGRATE=0` env var opts out (then the manual `scripts/migrate_triplestore_split.py` is available). Fast-path `has_pending_migration()` short-circuits the auto-migrate hook on already-migrated databases via an index-driven `EXISTS` check — microseconds instead of full-table scans every init.

**`Mnemosyne.export_to_file` / `import_from_file`** bumped to schema 1.1 and now include the `annotations` section. 1.0 backups still import cleanly; their annotation-flavored triples rows get relocated by the auto-migrate hook.

## Migration story

- New installs: schema is created in the post-E6 shape directly. No migration.
- Existing PyPI / `git pull` users: auto-migrate on first BeamMemory init writes a file-level backup to `{db}.pre_e6_backup` and moves annotation-flavored rows. `MNEMOSYNE_AUTO_MIGRATE=0` requires running `python scripts/migrate_triplestore_split.py` manually.
- Rollback: `cp {db}.pre_e6_backup {db}` and pin to a pre-E6 release.
- See `UPDATING.md` for the full operator guide.

## Test plan

- [x] Regression test: multiple mentions / facts per memory survive in `AnnotationStore` (the bug fix verified end-to-end through `BeamMemory.remember`)
- [x] Auto-migrate test: pre-E6 DB → first BeamMemory init migrates rows + writes backup + idempotent re-init
- [x] Opt-out test: `MNEMOSYNE_AUTO_MIGRATE=0` skips migration + logs warning with row count
- [x] Migration script: classify by predicate, transactional, idempotent, dry-run, CLI argparse layer
- [x] Export/import: schema 1.1 round-trip + 1.0 backward compat
- [x] TripleStore characterization: documents the pre-E6 silent-invalidation behavior so future readers understand the split
- [x] Full suite: 513 passed (was 499 on main; +14 net new regression tests)

## Review trail

This branch went through `/review` with all 6 specialist subagents (testing, maintainability, security, performance, data-migration, api-contract) plus Claude adversarial + Codex adversarial in the always-on pass. Findings addressed in the commits tagged `fix(e6): /review findings` and `fix(e6): adversarial-pass fixes`.

Notable items addressed mid-review:
- Migration logic moved into the package (`mnemosyne/migrations/`) so pip-installed deployments actually get auto-migration (`pyproject.toml` excludes `scripts/`)
- Export/import gained annotations support (pre-fix: backup/restore silently dropped them)
- `UNIQUE(memory_id, kind, value)` + `INSERT OR IGNORE` makes concurrent ingest + migration safe
- `AnnotationStore` accepts a shared connection from BeamMemory (eliminates per-call fd leak)
- CLI migration script default DB path was `triples.db`, fixed to `mnemosyne.db` (production)
- `TripleStore.add_facts` deprecated shim now routes to AnnotationStore (pre-fix it wrote to triples but post-E6 recall reads annotations — facts went silently invisible)

Deferred to follow-up (documented in CHANGELOG):
- `mnemosyne_triple_add` MCP tool still allows annotation-flavored predicates into the legacy triples table
- `Mnemosyne.forget()` doesn't cascade-delete annotations
- Migration discards `valid_until` on legacy invalidated rows (resurrects "lost" data — arguably desired)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,30 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Simple Versioning](https://github.com/AxDSan/mnemosyne) (MAJOR.MINOR).
 
+## [Unreleased]
+
+### Fixed
+
+**E6 — TripleStore silent-destruction bug**
+- `TripleStore.add()` auto-invalidates rows with matching `(subject, predicate)` regardless of `object`. Every production write used annotation semantics (`(memory_id, "mentions", entity)`, `(memory_id, "fact", text)`, etc.), so each new annotation for a memory silently set `valid_until` on prior annotation rows with the same key. Effect: entity / fact graphs on each Mnemosyne database have lost data any time a memory had more than one entity or fact extracted.
+- Fix splits storage into two purpose-specific tables:
+  - `triples` table retains current-truth temporal semantics with auto-invalidation, suitable for facts like `(user, prefers, X)` later superseded by `(user, prefers, Y)`. No production caller writes here today; the table is preserved for future use.
+  - New `annotations` table (`mnemosyne/core/annotations.py`, `AnnotationStore`) is append-only and now hosts `mentions`, `fact`, `occurred_on`, `has_source` — all multi-valued by design.
+- Production call sites migrated to `AnnotationStore`:
+  - `BeamMemory._extract_and_store_entities`, `_extract_and_store_facts`, `_add_temporal_triple`
+  - `BeamMemory._find_memories_by_entity`, `_find_memories_by_fact`
+  - `Mnemosyne.remember(extract_entities=True)` and `Mnemosyne.remember(extract=True)`
+- **Auto-migration on first BeamMemory init.** Existing databases auto-migrate annotation-flavored rows from `triples` to `annotations` with a backup written to `{db}.pre_e6_backup`. Set `MNEMOSYNE_AUTO_MIGRATE=0` to disable auto-migration and run `python scripts/migrate_triplestore_split.py` manually instead.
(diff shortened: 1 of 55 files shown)


kohai-ut:
This PR has a data migration/update step for the TripleStore db to split it - this runs if an old format is detected.  (See UPDATING.md for the full operator guide.) The performance testing we discussed needs this PR implemented first and then follow on PR to build on it.

PR #84, opened 2026-05-12
fix: consolidate_fact concurrent same-SPO race (silent data loss under contention)

kohai-ut:
## TL;DR

`VeracityConsolidator.consolidate_fact` had a SELECT-then-INSERT/UPDATE race that could lose observations under concurrent same-SPO calls. Two threads both passing the no-match SELECT and both attempting INSERT would race on the PRIMARY KEY — one INSERT succeeded, the other raised `IntegrityError` which propagated up to (or was swallowed by) the caller's broad `except: pass`. Result: silent data loss.

This PR wraps the SELECT-then-write in `BEGIN IMMEDIATE` so concurrent calls serialize at SQLite's writer-lock level. Plus four critical-but-subtle fixes the /review army surfaced: `_record_conflict` premature commit, silent fallthrough on lock failure, missing WAL/busy_timeout configuration on the consolidator's own connection, and a deterministic test that actually proves the regression is fixed.

**12 regression tests, all passing.** Race was flagged by maintainability reviewer on PR #83 as pre-existing but exacerbated by the deterministic hash IDs in that PR.

## Why this matters

`consolidate_fact` is called once per extracted fact from `_ingest_graph_and_veracity` — the always-on enrichment path that fires from both `BeamMemory.remember` (single-row) and (post-E2) `BeamMemory.remember_batch`. Pre-fix, any concurrent caller hitting the same SPO could lose its observation.

For the BEAM-recovery experiment specifically: under E2's `remember_batch` enrichment, the batch runs in one thread so internal calls don't race against each other. The risk shape is concurrent batches OR concurrent direct callers — less likely for the experiment, more likely for production deployments with multi-process Hermes plugins.

Pre-fix the truncation in `consolidated_facts.id` (now fixed by PR #83's hash-based ID) sometimes "worked" by producing different IDs from semantically-same SPOs depending on whitespace. With PR #83's deterministic hashes, the race reliably surfaces. Either way, the underlying SELECT-then-write race is real; this PR closes it.

## What this PR does

**2 commits.** Implementation in commit 1; comprehensive /review hardening in commit 2.

### Core fix

`consolidate_fact` now wraps its SELECT-then-INSERT/UPDATE in `BEGIN IMMEDIATE` when not already in a transaction. The RESERVED lock acquired at BEGIN serializes concurrent same-SPO writers at the SQLite engine level. Nested-tx detection via `conn.in_transaction` skips the BEGIN when the caller already owns one.

`VeracityConsolidator.__init__` now applies `PRAGMA journal_mode=WAL` + `PRAGMA busy_timeout=5000` to connections it constructs (matches what `beam._get_connection` does for connections it owns). Without these, the BEGIN IMMEDIATE pattern's contention semantics are wrong.

### Commit 2 /review hardening

4-source convergence on critical issues in commit 1:

| Finding | Sources | Fix |
|---|---|---|
| `_record_conflict` unconditional `commit()` ended our BEGIN IMMEDIATE scope mid-method; fact INSERT durable before all conflicts recorded; rollback couldn't undo | Codex struct + Codex adv + Maintainability + Claude (4-source HIGH) | Added `commit: bool = True` kwarg; `consolidate_fact` passes `commit=False`. Backward compat preserved for non-nested callers. |
| Silent fallthrough on `BEGIN IMMEDIATE` failure → ran unprotected SELECT-then-write, reintroducing the race | Codex struct + Codex adv + Maintainability + Claude (4-source HIGH) | Let `OperationalError` propagate. Callers see a clear failure (lock held past busy_timeout) instead of silent data loss. |
| `VeracityConsolidator.__init__` didn't set WAL or busy_timeout on its own connection — under `journal_mode=DELETE` BEGIN IMMEDIATE blocks readers AND raises instantly without busy_timeout | Claude C3 | Apply PRAGMAs in `__init__` for connections we own. |
| DEFERRED outer transactions don't actually serialize (no writer lock until first write); two threads in DEFERRED outer txs can both pass our SELECT before either writes | Codex adv H3 + Claude H3 | Cannot fix from inside `consolidate_fact`; documented as a caller-contract requirement: race safety inside an outer tx requires either (a) outer tx is `BEGIN IMMEDIATE`/`BEGIN EXCLUSIVE`, OR (b) caller is the only writer. |
| Barrier-only concurrency tests unreliable (GIL + WAL writer mutex serialize statements faster than race window opens) | Claude H1 + Codex adv MED | Added `test_race_window_widening_demonstrates_serialization` — monkey-patches `bayesian_update` to sleep 50ms inside the critical section, deterministically widening the window. |
| Rollback failure silently swallowed | Maintainability MED | Now logs at ERROR with the rollback exception; original exception still propagates. |

## What is NOT in this PR (intentional, documented)

- **`resolve_conflict`, `resolve_conflict_by_facts`, `run_consolidation_pass`** have similar SELECT-then-write race patterns. Concurrent same-conflict resolution can still be last-writer-wins. Out of scope; documented in `consolidate_fact`'s docstring as a known limitation. Fix shape is the same `BEGIN IMMEDIATE` wrap — easy to apply in a follow-up.
- **DEFERRED outer-tx caller-side fix**: needs the caller (e.g., E2's `_deferred_commits` from PR #82) to use `BEGIN IMMEDIATE` instead of DEFERRED for full race safety across concurrent batches. Documented in this PR's docstring; E2 is single-threaded per batch so the issue doesn't bite under the BEAM-experiment workload.
- **Python sqlite3 3.12+ isolation_level reworks**: documented as a future caveat; current behavior tested on 3.11.

## Performance characterization

Per-call overhead of BEGIN IMMEDIATE under no contention: ~2-5 microseconds (lock-state change, no fsync). For 250K calls in a BEAM-scale batch ingest: ~1-2 seconds total — negligible. Under contention, busy_timeout=5000ms gives concurrent writers a 5-second window to acquire the lock; severe contention surfaces as an `OperationalError` to the caller (correct behavior).

When called from within a deferred-commits scope (E2 PR #82's `_BeamConnection` pattern), `conn.in_transaction` is True and we skip the BEGIN — zero overhead, the outer scope's commit-deferral plus single-threaded batch execution provides the serialization we need.

## Test plan

- [x] All 12 tests in `tests/test_consolidate_fact_concurrency.py` pass (6 baseline + 6 hardening)
- [x] All 24 existing `test_beam_e4_remember_batch_veracity.py` tests pass
- [x] All 20 existing `test_integration.py` tests pass

### DIFF
--- a/mnemosyne/core/veracity_consolidation.py
+++ b/mnemosyne/core/veracity_consolidation.py
@@ -128,6 +128,22 @@ class VeracityConsolidator:
         else:
             self.db_path = db_path or Path.home() / ".hermes" / "mnemosyne" / "data" / "mnemosyne.db"
             self.conn = sqlite3.connect(str(self.db_path), check_same_thread=False)
+            # Apply the same PRAGMA settings BeamMemory's _get_connection
+            # uses (journal_mode=WAL, busy_timeout=5000ms). Without these,
+            # `BEGIN IMMEDIATE` in `consolidate_fact` runs under default
+            # `journal_mode=DELETE` which blocks readers too AND raises
+            # `database is locked` immediately under contention instead
+            # of waiting through the busy_timeout. /review (Claude
+            # adversarial C3) caught the connection-setup gap.
+            try:
+                self.conn.execute("PRAGMA journal_mode=WAL")
+                self.conn.execute("PRAGMA busy_timeout=5000")
+            except sqlite3.Error:
+                # Best-effort: some constrained environments (e.g.,
+                # in-memory DBs) don't support WAL. Continue without —
+                # the BEGIN IMMEDIATE path still works, just with
+                # different contention semantics.
+                pass
         self.conn.row_factory = sqlite3.Row
         self._owns_connection = conn is None
         self._init_tables()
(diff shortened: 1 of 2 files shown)


kohai-ut:
Closing in favor of the helper-based path:

- **Production fix:** moved to #86, which routes `consolidate_fact` through the shared `_serialized_write` context manager introduced in #85. The helper provides the same `BEGIN IMMEDIATE` + `_record_conflict(commit=False)` semantics as this PR's inline patch, plus same-connection `threading.RLock` protection (which this PR didn't have).
- **Regression coverage:** the 12 tests in this PR's `tests/test_consolidate_fact_concurrency.py` were ported verbatim onto #86 in commit [c89d708](https://github.com/kohai-ut/mnemosyne/commit/c89d708) — all 12 pass against the helper-based implementation without modification, proving the protections are equivalent. No coverage loss from closing this PR.
- **Why not just merge all three (#84 → #85 → #86)?** #86 immediately refactors away #84's inline pattern, so merging this would have introduced a "added inline" → "moved to helper" churn within two commits. The cherry-pick keeps history cleaner while preserving every regression test.

Suggested merge order: #85 first (introduces the `_serialized_write` helper + RLock + sibling-method fixes), then #86 (migrates `consolidate_fact` onto the helper + carries the ported regression tests).

PR #86, opened 2026-05-12
refactor: DRY consolidate_fact onto _serialized_write helper

kohai-ut:
## TL;DR

DRY refactor of `VeracityConsolidator.consolidate_fact` to use the `_serialized_write` helper introduced in PR #85. Closes the same-connection race for consolidate_fact (which was left open by PR #84's inline-only pattern), unifies all 4 write methods on one canonical concurrency pattern, and removes ~85 lines of duplicated transaction-management code.

**14 regression tests, all passing.** Codex structured /review: **GATE PASS**, no P0/P1/P2 issues.

## What this PR closes

After PR #85 introduced `threading.RLock` on the instance + applied `_serialized_write` to 3 sibling methods, `consolidate_fact` was the ONLY method on `VeracityConsolidator` still using PR #84's inline pattern. Two consequences this PR closes:

1. **Same-connection race left open for consolidate_fact.** PR #85's RLock protects 3 of 4 methods. `consolidate_fact` didn't acquire it because it didn't go through the helper. Two threads sharing one `VeracityConsolidator` instance (e.g., via `polyphonic_recall.py:88`'s `conn=conn` passthrough) could still race in consolidate_fact.

2. **Code duplication.** Two patterns for the same concurrency concern made it easy for future modifications to drift between methods.

## What this PR does

Single commit:

- `consolidate_fact` body wrapped in `with self._serialized_write():` — no inline `BEGIN IMMEDIATE`, no try/except for tx management, no `self.conn.commit()` calls (the helper handles all of that)
- `_record_conflict` gains `commit: bool = True` kwarg (mirroring PR #84's pattern); consolidate_fact passes `commit=False` from inside the helper scope so the fact INSERT + conflict rows commit atomically
- New test `test_consolidate_fact_same_connection_serializes_via_rlock` (mirrors PR #85's `test_same_connection_writers_serialize_via_rlock` but exercises consolidate_fact)

Net effect: **all 4 write methods now share one canonical pattern + all 4 are protected by the RLock**.

## PR stacking + maintainer notes

This PR depends on PR #85's `_serialized_write` helper and `_write_lock` RLock.

**Merge order options:**

| Order | Result |
|---|---|
| #85 first, then this | Clean. consolidate_fact starts un-fixed (pre-#84 state on this branch's base), this PR fixes it via the helper. |
| #84, #85, this | Clean. #84's inline pattern lands, #85 adds helper for siblings, this PR migrates consolidate_fact onto helper. |
| #85, this, #84 | Awkward. #84 would re-introduce the inline pattern that this PR removed. Maintainer should probably close #84 in this case. |

**Cleanest story:** #85 + this PR together supersede #84. If the maintainer prefers to land the helper-based approach end-to-end, they can close #84 in favor of #85 + this. The end state is identical for consolidate_fact's race-fix behavior; this PR just uses the canonical helper instead of inline.

## Test plan

- [x] All 14 tests in `tests/test_consolidate_fact_sibling_races.py` pass (13 from #85 + 1 new)
- [x] All 20 tests in `tests/test_integration.py` pass
- [x] All 24 tests in `tests/test_beam_e4_remember_batch_veracity.py` pass
- [x] Codex structured /review: GATE PASS
- [ ] Claude adversarial review (in progress)

### DIFF
--- a/mnemosyne/core/veracity_consolidation.py
+++ b/mnemosyne/core/veracity_consolidation.py
@@ -292,108 +292,143 @@ class VeracityConsolidator:
                         veracity: str = "unknown", source: str = None) -> ConsolidatedFact:
         """
         Add or update a fact in consolidation.
-        
+
         Args:
             subject: Fact subject
             predicate: Fact predicate
             object: Fact object
             veracity: Veracity tier
             source: Source memory ID
-            
+
         Returns:
             ConsolidatedFact: The consolidated result
+
+        Concurrency: SELECT-then-INSERT/UPDATE wrapped in
+        ``_serialized_write``. Pre-fix two threads with same SPO
+        could both pass the no-match SELECT and both attempt
+        INSERT, raising IntegrityError. Bayesian confidence math
+        is path-dependent so concurrent UPDATEs also raced.
+
+        Post-fix: the helper's RLock serializes same-instance
+        callers + `BEGIN IMMEDIATE` serializes cross-connection
+        writers via the SQLite writer lock. Caller-contract caveat
+        for DEFERRED outer txs: see `_serialized_write` docstring.
+
+        Pre-DRY (PR #84) this method had its own inline
+        BEGIN IMMEDIATE pattern. The helper-based version here
+        unifies it with the other three write methods so all four
+        share one canonical pattern AND the instance RLock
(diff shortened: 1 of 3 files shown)


kohai-ut:
85 must land before 86.

PR #93, opened 2026-05-12
fix: restore PR #82/#83/#91 work stripped by the May 12 merge resolution

kohai-ut:
## TL;DR

Merge commit `016d395` ("Merge remote-tracking branch 'origin/main' into fix/pre-experiment-fidelity") resolved several conflicts by keeping the helper functions but reverting the call sites or dropping module-level helpers entirely. Three previously-merged PRs landed only partially. **25 tests fail on current `main`** as a direct result.

This PR restores the dropped wiring. No new logic; the helpers were already present.

**Post-fix: 0 failed, 979 passed, 10 skipped.**

### PR #82 — `remember_batch` enrichment parity

`_add_temporal_triple` and `_ingest_graph_and_veracity` still exist as methods, but `remember_batch` no longer calls them per row. Net effect: the polyphonic engine's graph + fact voices have no data to fuse — E5's 4-voice RRF silently collapses to 2 voices for any batch-ingested corpus.

**Fix:** per-row enrichment loop after the embedding block. Defensive `try/except` so a single bad row doesn't kill the batch. Also restored the `MEMORY_ADDED` event emission for streaming observability parity with `remember()`.

**12 tests in `test_e2_remember_batch_enrichment.py` were failing; now pass.**

### PR #83 — `consolidated_facts.id` collision fix

`compute_fact_id` exists at `veracity_consolidation.py:38` (SHA-256 of NFC-normalized SPO with length-prefix framing) but `consolidate_fact` at line 529 still used the old `f"cf_{subject}_{predicate}_{object}".replace(" ", "_")[:100]` truncating pattern — the exact bug PR #83 was filed to fix.

**Fix:** the call site now uses `compute_fact_id(subject, predicate, object)`.

**2 tests in `test_consolidate_fact_id_collision.py` + 1 integration test were failing; now pass.**

### PR #91 — env-parser + telemetry helpers (entirely missing)

Three module-level helpers entirely missing from main:

- **`_env_float(name, default)`** — defensive float-from-env parser that falls back to `default` on empty / whitespace / invalid values, emits WARN on unparseable values. Without it, `MNEMOSYNE_STATED_WEIGHT=` (empty) crashes module load via `float("")` ValueError.
- **`_detect_veracity_weight_overrides()`** — returns the list of set `MNEMOSYNE_*_WEIGHT` env vars (filtering empties).
- **`_warn_about_veracity_weight_overrides(force=False)`** — single-emit-per-process WARN when veracity weight env overrides drift the recall multiplier vs the consolidator's hardcoded `VERACITY_WEIGHTS`.

**Fix:** all three helpers restored. Weight constants route through `_env_float`. Module-load calls `_warn_about_veracity_weight_overrides()` once.

**9 tests in `test_telemetry_and_env_followups.py::TestC32...` were failing; now pass.**

## Verification

| | Pre-fix | Post-fix |
|---|---|---|
| `test_e2_remember_batch_enrichment.py` | 5/17 pass | 17/17 pass |
| `test_consolidate_fact_id_collision.py` | 19/21 pass | 21/21 pass |
| `test_telemetry_and_env_followups.py` | 31/40 pass | 40/40 pass |
| `test_integration.py::TestVeracityConsolidation` | 1 failing | All pass |
| **Full suite** | **25 failed, 961 passed** | **0 failed, 979 passed, 10 skipped** |

## How this happened — process lesson worth recording

The merge of `origin/main` into the PR #91 branch (`fix/pre-experiment-fidelity`) brought in PRs #82, #83, #85, #86, #87, #80, #89 — many of which touched the same hunks PR #91 had modified. The conflict resolution chose `theirs` (main) for the call-site hunks but `ours` (PR #91) for the helper-definition hunks, leaving helpers present but unused. This is the classic "merge resolution kept the wrong side per hunk" failure mode.

It's a failure mode no /review pass or static analysis can catch — the helpers compile, type-check, and import cleanly. Only the tests that exercise the call sites end-to-end reveal the missing wiring.

**Recommendation for future multi-PR-overlap merges:** when resolving conflicts between branches that both modified the same hunks, run the existing test suite on the merge commit BEFORE pushing. With Mnemosyne's ~990-test suite running in ~17 seconds, this is a cheap safety net that would have flagged the 25 regressions immediately. Something like:

```bash
git merge --no-commit <branch>
pytest --tb=no -q
# if anything regressed vs the pre-merge baseline, the resolution chose wrong per-hunk
```

The current CI on `main` is showing the same 25 failures right now, so this is also blocking unrelated PRs that sit on top of main (notably #92, which is failing CI on these exact tests). Merging this PR first unblocks the others without rebasing them.

## How to detect this kind of regression cheaply in the future

Any test that references a helper function and passes on the PR branch but fails post-merge indicates a stripped call site. The pattern this PR found:

```bash
git diff <pr-merge-base> HEAD -- mnemosyne/core/beam.py | grep '^-def '
# any -def helper present on PR branch but missing on main, where tests
# still reference the helper, signals exactly this regression
```

## Test plan

- [x] All 25 previously-failing tests now pass
- [x] No new regressions: full suite 979 passed, 0 failed
- [x] Recovery is mechanical (no new logic, no API changes)

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -189,11 +189,88 @@ TIER3_MAX_CHARS = int(os.environ.get("MNEMOSYNE_TIER3_MAX_CHARS", "300"))
 # Env-var overrides remain so operators can tune ranking; documented
 # drift risk: if `MNEMOSYNE_*_WEIGHT` is set, recall scoring diverges
 # from consolidation confidence math (consolidator doesn't honor env).
-STATED_WEIGHT = float(os.environ.get("MNEMOSYNE_STATED_WEIGHT", str(_VW_DEFAULTS["stated"])))
-INFERRED_WEIGHT = float(os.environ.get("MNEMOSYNE_INFERRED_WEIGHT", str(_VW_DEFAULTS["inferred"])))
-TOOL_WEIGHT = float(os.environ.get("MNEMOSYNE_TOOL_WEIGHT", str(_VW_DEFAULTS["tool"])))
-IMPORTED_WEIGHT = float(os.environ.get("MNEMOSYNE_IMPORTED_WEIGHT", str(_VW_DEFAULTS["imported"])))
-UNKNOWN_WEIGHT = float(os.environ.get("MNEMOSYNE_UNKNOWN_WEIGHT", str(_VW_DEFAULTS["unknown"])))
+def _env_float(name: str, default: float) -> float:
+    """Parse an env var as float; fall back to `default` on empty or
+    invalid values rather than crashing at module load.
+
+    Pre-fix `float(os.environ.get("MNEMOSYNE_STATED_WEIGHT", "1.0"))`
+    raised ValueError when the env var was set to empty (`export
+    MNEMOSYNE_STATED_WEIGHT=`) because `os.environ.get` returns `""`
+    (the value), not the default — `float("")` then crashed import
+    BEFORE the C32 override-WARN could fire. Restored from PR #91
+    after the merge stripped it.
+    """
+    raw = os.environ.get(name, "")
+    raw = raw.strip()
+    if not raw:
+        return default
+    try:
+        return float(raw)
+    except ValueError:
+        logger.warning(
+            "%s=%r is not a valid float; falling back to default %s",
+            name, raw[:80], default,
+        )
+        return default
+
+
+# Veracity weighting (memory confidence)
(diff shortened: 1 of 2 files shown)


kohai-ut:
Combined into #92 so the bundled CI run passes (the docs+harness changes in #92 were failing CI on the exact 25 main-regression tests this PR was fixing). Recovery commit is now cherry-picked onto #92's branch as the second commit. Both sets of changes ship together when #92 merges.

PR #108, opened 2026-05-13
fix: two bugs in PR #106's identity-memory test fixture blocking CI

kohai-ut:
## Summary

PR #106 added ``tests/test_identity_memory.py`` whose fixture errors on **every PR's CI matrix** with two separate setup failures, currently blocking all five of my open PRs (#98, #101, #102, #105, #107). Same fixture errors would block every future PR until fixed.

The fixture is correct in intent — the test BODIES exercise real behavior. Only the setup has two bugs that combined to make the suite show `5 errors` and mark the CI run failed.

### Bug 1: `BeamMemory(db_path=<str>)` crashes at first connection

```python
beam = BeamMemory(db_path=db_path)  # db_path is a tempfile string
```

`BeamMemory.__init__` stored the string as-is, then `_get_connection` at `beam.py:326` did `path.parent.mkdir(parents=True, exist_ok=True)` → ``AttributeError: 'str' object has no attribute 'parent'``.

**Fix**: `BeamMemory.__init__` now coerces any non-Path `db_path` via `Path(db_path)` before storage. Backward-compatible — real `Path` inputs stay `Path`, strings get coerced, `None` still falls back to `_default_db_path()`. The implicit Path-only contract is now explicit and tolerant.

### Bug 2: `beam.initialize()` is not a thing

```python
beam = BeamMemory(db_path=db_path)
beam.initialize()  # <-- AttributeError
```

`BeamMemory` has no `initialize()` method. `MnemosyneMemoryProvider` does, and the test author may have left this over from a refactor. `BeamMemory.__init__` already opens the connection and runs `init_beam()` for schema setup.

**Fix**: drop the spurious `beam.initialize()` call from the fixture.

## Test plan

- [x] New `tests/test_beam_db_path_string_coercion.py`: 4 regression tests pinning the db_path acceptance contract across `str` / `Path` / `None` inputs, plus an end-to-end smoke that exercises the `.parent.mkdir(parents=True)` path that the original AttributeError hit.
- [x] PR #106's 5 identity tests now pass (verified locally with the fix).
- [x] 143/143 in the broader slice across hermes_memory_provider, hermes_plugin_session, plugins, integration, data_dir_scripts.
- [ ] CI matrix will confirm.

## Why this PR exists

This fix could/should land on main directly without going through a PR, but since the CI bug is blocking five separate open PRs that I've been shipping, this is the fastest path to unblock all of them at once. Maintainer can land this first, then rebase the other five PRs onto the cleaned main.

If you'd prefer to fold these two changes into a different PR, happy to close this one — just say the word.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1340,6 +1340,17 @@ class BeamMemory:
         self.author_id = author_id
         self.author_type = author_type
         self.channel_id = channel_id or session_id  # default channel = session
+        # Coerce path-like inputs (e.g. tempfile-produced strings) to a
+        # Path so downstream consumers like _get_connection that do
+        # `path.parent.mkdir(...)` don't blow up with
+        # ``AttributeError: 'str' object has no attribute 'parent'``.
+        # The previous contract was implicit (Path-only); making it
+        # explicit here is backward-compatible -- a real Path stays a
+        # Path -- and unbreaks every caller that passes a string,
+        # including the test_identity_memory.py fixtures added by
+        # PR #106 which now run on every PR's CI matrix.
+        if db_path is not None and not isinstance(db_path, Path):
+            db_path = Path(db_path)
         self.db_path = db_path or _default_db_path()
         self.use_cloud = use_cloud  # Enable LLM fact extraction during remember()
         self._extraction_client = None  # Lazy-loaded ExtractionClient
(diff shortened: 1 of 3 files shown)


kohai-ut:
FYI, this PR will fix the failing CI on the other waiting PRs I did.

PR #120, opened 2026-05-13
fix(entities): block meta-word extraction noise in mention annotations

ether-btc:
## Summary

**Problem**: 179 episodic memories produced 7,591 'mentions' annotations (42x amplification). Top extracted "entities" were noise: ASSISTANT(394), USER(375), SKILL(104), Review(104), TARGET(93), CLASS(91), LEVEL(86), SIGNALS(86).

**Root causes**:
1. Stopword list had only standard English stopwords, no meta/system words from LLM-generated summaries
2. Stopword filter was case-sensitive (caught 'user' but not 'USER')
3. Multi-word entities like 'The USER', 'Be ACTIVE' passed because only single-word stopwords were checked

**Fix** ():
1. Expanded  with 25+ meta/system words: ASSISTANT, USER, SKILL, CLASS, LEVEL, SIGNALS, PHASE, API, PI, SUMMARY, ACTIVE, BE, NOT, WHETHER, AI, MEMORY, MNEMOSYNE, CONVERSATION, FACT, SIGNAL, HERMES, AGENT, MODEL, SYSTEM, NOTE, TASK, PROJECT, RESULT, OUTPUT, INPUT, DATA, STEP, PROCESS, POINT, WAY, THING, TIME, WORK
2. Made stopword filter case-insensitive ( on entity)
3. Added any-word-stopword filter: if ANY word in a multi-word entity is a stopword, drop the entity

**Verification**:
- Real entities still extract: Alice, Bob, New York, @abdias, #Python, Hercules
- Noise eliminated: The USER, Be ACTIVE, ASSISTANT, signals, LEVEL, target

## Files changed

- : expanded stopwords, case-insensitive filter, any-word filter

## Testing

Run: `python3 -c "from mnemosyne.core.entities import extract_entities_regex; ..."`


AxDSan (commented):
Good fix overall, the core change is solid. A few things to flag before merging:

**Core logic (entities.py) looks good** -- case-insensitive filtering and any-word-stopword dropping solves the right problems. 49 tests pass.

**Concerns:**

1. **CONTINUE_HERE.md shouldn't be in the repo.** This is a 51-line session continuation doc. Commit messages and PR description are the right place for this -- it adds noise to the project root.

2. **Duplicate stopwords in entities.py.** "assistant", "memory", "be", "no", and "all" appear in both the standard set and the meta set. Set dedup hides this but it suggests the two lists were merged without cleanup.

3. **scripts/cleanup_noisy_mentions.py has a different hardcoded stopword list.** Smaller list, uppercase values. Maintenance hazard -- anyone adding stopwords to entities.py wont know to update the script. Consider importing from entities.py.

4. **Same noise filter logic duplicated in two places** (entities.py write-time + annotations.py retrieval-time) with different stopword sets. A shared constant would prevent drift.

5. **Scope budget:** The actual fix is ~30 lines. The PR adds 235 including a cleanup script (115 lines) and continuation doc (51 lines). Cleanup script is useful but could be a follow-up.

**Nit:** Test changes change existing semantic expectations significantly ("The Quick Brown Fox" used to be kept, now dropped). Worth noting for consumers.

ether-btc:
Thanks for the thorough review, AxDSan. All 6 concerns have been addressed in commit 8370526, pushed to the branch.

**1. CONTINUE_HERE.md removed** ✅
Deleted from repo. Session continuation docs belong in the commit trail and PR description, not the project root.

**2. Stopwords deduplicated** ✅
Split the single flat set into `_STANDARD_STOP_WORDS` and `_META_STOP_WORDS`, then unioned into `ENTITY_EXTRACTION_STOP_WORDS`. Removed the cross-set duplicates (`assistant`, `memory`, `be` were in both) and the within-set duplicate (`assistant` was listed twice). Each named group is independently auditable now with no duplicates.

**3. Cleanup script imports from entities.py** ✅
`scripts/cleanup_noisy_mentions.py` now does:
```python
sys.path.insert(0, ...)
from mnemosyne.core.entities import ENTITY_EXTRACTION_STOP_WORDS
```
No more hardcoded parallel list — the old `NOISE_WORDS` set had a smaller list with mixed casing, a maintenance hazard that could drift. The script now uses the same `_is_noisy()` function logic (case-insensitive, any-word match) as the write-time and retrieval-time filters. Anyone adding a stopword to `entities.py` doesn't need to touch the script.

**4. Annotations filter imports from entities.py** ✅
`_ENTITY_STOP_WORDS` in `annotations.py` is now `frozenset(ENTITY_EXTRACTION_STOP_WORDS_SOURCE)` — a derived reference from the canonical set in `entities.py`, not a separate declaration. Verified: both hold exactly 117 stopwords and are identical. No more filter logic duplication between write-time and retrieval-time paths — both reference the same constant, so drift is impossible.

**5. Semantic change documented in tests** ✅
The `test_stop_words_filtered` test clearly asserts that "The Quick Brown Fox" produces `["Quick", "Brown", "Fox"]` — the multi-word phrase is dropped because "The" contaminates it, but the three single-word capitalized entities survive. The test comments explain why.

**6. Cleanup script kept** ✅
Agreed it's useful for production DB recovery. Now that it shares the canonical import, it's no longer a maintenance burden — just a consumer of the single source of truth.

AxDSan (commented):
CI is failing on Python 3.9 with 2 test failures, both pre-existing tests the PR didn't touch.

**Failure 1:** `test_extract_and_store_entities`
`'Mnemosyne' not found in ['Abdias', 'New York']`

"Mnemosyne" is in the stopword list. The any-word-stopword filter drops it. The project name is not noise.

**Failure 2:** `test_find_memories_by_entity`
`0 != 2` -- expected 2, got 0

Cascading from failure 1. Entities that should be extracted are filtered out, entity recall returns nothing.

**The issue:** The stopword list is too aggressive. It mixes "words that appear frequently in LLM output" with "words that are never valid entities." "mnemosyne", "memory", "agent", "model", "system", "note", "task", "project", "result", "output", "input", "data", "step", "process", "point", "way", "thing", "time", "work" -- these are not noise in context. They're project names and technical terms that come up in real conversations about AI.

You already have the retrieval-time filter in annotations.py as defense-in-depth. The write-time stopword list can be less aggressive.

Trim the stopword list to things that are genuinely NEVER entities (assistant, user, skill, review, target, class, level, signals, phase, pi, summary, added, replying, false, true, none, null, signal). Drop the project keywords and generic words. The case-insensitive fix and the any-word-stopword logic are both worth keeping, they just need a smaller list.

AxDSan (changes requested):
## Code Review: PR #120 — Entity Extraction Noise Fix

@ether-btc good approach — the any-word-stopword filter and retrieval-time defense-in-depth are both solid ideas. But the stopword list is too aggressive and **CI is failing**.

### What is working

- Case-insensitive stopword matching (catches `USER` now, not just `user`) ✅
- Any-word-stopword filter correctly blocks `"The USER"`, `"Active Signal"` ✅
- Retrieval-time filter in `annotations.py` (non-destructive, no DB deletes) ✅
- Cleanup script with `--dry-run` ✅
- Comment #1 (AxDSan, 6 concerns) fully addressed in commit `8370526` ✅

### What is broken

**CI fails** — two integration tests in `tests/test_entity_integration.py`:

```
test_extract_and_store_entities: 'Mnemosyne' not found in ['Abdias', 'New York']
test_find_memories_by_entity: 0 != 2
```

**Root cause:** The `_META_STOP_WORDS` set includes words that are valid entities in AI/technical conversations. Specifically, these should be **removed** from the stopword list:

> `mnemosyne`, `memory`, `agent`, `model`, `system`, `note`, `task`, `project`, `result`, `output`, `input`, `data`, `step`, `process`, `point`, `way`, `thing`, `time`, `work`

These are project keywords and generic terms that can legitimately appear as entities in context. The genuinely never-valid noise words to keep are about 19: `assistant`, `user`, `skill`, `review`, `target`, `class`, `level`, `signals`, `phase`, `api`, `pi`, `summary`, `added`, `replying`, `false`, `true`, `none`, `null`, `signal`.

### What needs to happen

1. **Trim `_META_STOP_WORDS`** — keep only the 19 genuinely never-valid noise words. Remove the 18 project-keyword/generic entries.
2. **Update `tests/test_entity_integration.py`** — the two failing tests need assertions aligned with the new behavior (or should pass unchanged if the stopword list is properly trimmed).

### Recommendation

**❌ Request changes.** The architecture is good, the implementation is clean, but the stopword list needs to be less aggressive. Once the list is trimmed and CI passes, this is ready to merge.

ether-btc:
Trimmed META_STOP_WORDS from 46 -> 23 words per your recommendation. Removed: ai, memory, mnemosyne, conversation, fact, hermes, agent, model, system, note, task, project, result, output, input, data, step, process, point, way, thing, time, work. All 32 entity tests + all 8 integration tests pass (40/40). Sets remain synchronized between entities.py and annotations.py (94 stopwords, identical).

ether-btc:
@AxDSan — all three rounds of review feedback addressed:

- **Review 1 (6 concerns):** Fixed in `8370526` — removed CONTINUE_HERE.md, split stopwords into `_STANDARD_STOP_WORDS` + `_META_STOP_WORDS`, cleanup script now imports from `entities.py`, both files reference shared constant.
- **Reviews 2 & 3 (CI failures):** Fixed in `e9c1c9b` — trimmed `_META_STOP_WORDS` from 46→23, removed all project-keyword/generic terms incorrectly flagged as noise (`mnemosyne`, `memory`, `agent`, `model`, `system`, `data`, `input`, `output`, `result`, etc.). Kept only genuinely never-valid noise.

**Final state:**
- `_META_STOP_WORDS`: 46 → 23 words
- `ENTITY_EXTRACTION_STOP_WORDS`: 94 total (71 standard + 23 meta)
- CI: ✅ 5/5 checks passing (Python 3.9, 3.10, 3.11, 3.12 + build)
- Tests: ✅ 40/40 passing

Branch is ready for re-review. Let me know if anything else needs adjustment.

steezkelly:
I re-checked the current head because the review state still shows `CHANGES_REQUESTED` even though the latest CI is green.

Local focused verification on the current PR branch:

```bash
/home/steve/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_entities.py tests/test_entity_integration.py -q
# 40 passed
```

The specific earlier regression called out in review appears fixed: `Mnemosyne` is no longer in the meta-stopword set, and the integration tests that expect it as an entity now pass. `CONTINUE_HERE.md` is also gone from the final diff. Final changed files are scoped to:

- `mnemosyne/core/entities.py`
- `mnemosyne/core/annotations.py`
- `scripts/cleanup_noisy_mentions.py`
- `tests/test_entities.py`

One small follow-up thought, non-blocking for this PR: the current any-word stopword behavior intentionally splits/drops quoted entities containing standard stop words (e.g. `"Bank of America"` becomes `Bank` / `America` rather than the full phrase). That looks pre-existing/adjacent to the extraction heuristic rather than part of the review blocker, but it may be worth a future entity-extraction refinement if phrase fidelity matters.

AxDSan (approved):
Alright, the stopword list is trimmed down to the genuinely never-valid noise words. Mnemosyne, memory, agent, model — all the project terms that were getting blocked — are back in play. CI is green across the board.

The any-word-stopword filter and retrieval-time defense-in-depth in annotations.py are both solid. Cleanup script properly imports from the canonical source now.

Integration tests pass without modification because the stopword list no longer eats project keywords. That's the right outcome.

Previous concerns all addressed. APPROVE. Thanks for sticking with the back-and-forth on this ether-btc 🤝

PR #128, opened 2026-05-13
feat: link memory — graph traversal, recall integration, and provider tools

coe0718:
Implements #125 — weighted graph traversal on top of the existing episodic_graph engine.

### 1. Enhanced graph traversal (episodic_graph.py)
- `find_related_memories` now accepts `edge_type` and `min_weight` filter params
- Returns rich dicts (`memory_id`, `edge_type`, `weight`, `depth`) instead of bare ID lists
- Proper deduplication across BFS depth levels

### 2. Recall integration (polyphonic_recall.py)
- `_graph_voice` now runs graph traversal from seed memory IDs discovered via entity extraction
- Discovers indirectly related memories through multi-hop ctx edges
- Score decays with hop distance (0.4 / depth)

### 3. Tool surface
- **mnemosyne_graph_query** — agent-driven multi-hop graph traversal with edge_type/min_weight filtering
- **mnemosyne_graph_link** — lets agents declare semantic edges between memories
- Both wired into provider dispatch (17 tools total) and plugin.yaml

## Testing
16 new tests covering: basic BFS traversal, edge_type filtering, weight threshold, depth control, result format, agent-declared edge types, auto-populated edges from beam.remember(), polyphonic voice integration, tool dispatch, and error paths.

None of these changes are on by default — polyphonic recall is gated behind `MNEMOSYNE_POLYPHONIC_RECALL=1` and the graph tools are only used when explicitly called. Zero behavior change for existing deployments.

PR #130, opened 2026-05-14
refactor: migrate test_extraction callers from deprecated TripleStore.add_facts to AnnotationStore.add_many

coe0718:
## Summary

Migrates two test functions in `test_extraction.py` from the deprecated `TripleStore.add_facts()` shim to `AnnotationStore.add_many()`:

- `test_triplestore_add_facts` — now calls `AnnotationStore.add_many` directly instead of routing through the deprecation shim with `warnings.catch_warnings`. Removed the `"x"` test case since `add_many` doesn't filter by string length.
- `test_triplestore_add_facts_empty` — same migration for the empty-list case.

Extracted from review feedback on PR #128 (graph traversal) where AxDSan noted these changes were orthogonal to the PR scope and belonged in their own PR.

## Testing

19/19 tests passing, zero warnings.

### DIFF
--- a/tests/test_extraction.py
+++ b/tests/test_extraction.py
@@ -162,35 +162,29 @@ def test_extract_facts_safe_exception_handling():
 
 
 def test_triplestore_add_facts():
-    """Test TripleStore.add_facts() batch storage.
+    """Test TripleStore.add_facts() batch storage — migrated to AnnotationStore.
 
-    Post-E6: add_facts is a deprecation shim that routes writes to the
-    AnnotationStore (not the triples table). This was changed during the
-    /review adversarial pass — pre-redirect, deprecated callers' facts
-    went into the triples table but the new recall path read from
-    annotations, making the facts silently invisible.
+    Post-E6: add_facts is a deprecation shim. Test calls AnnotationStore.add_many
+    directly instead.
     """
-    import warnings
     from mnemosyne.core.annotations import AnnotationStore
 
     with tempfile.TemporaryDirectory() as tmpdir:
         db_path = Path(tmpdir) / "test.db"
         init_triples(db_path)
 
-        triples = TripleStore(db_path=db_path)
-        with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


steezkelly:
I reviewed this locally against the current PR head and current `origin/main`.

Commands/results:

```bash
/home/steve/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_extraction.py -q
# 19 passed in 0.62s

/home/steve/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_extraction.py tests/test_e6a_followup_gaps.py -q
# 33 passed in 3.41s

# trial merge current origin/main into the PR branch
# result: clean
```

Conclusion: the test migration itself behaves correctly and I did not see a merge conflict.

One non-blocking cleanup/coverage note before merge: the updated tests now call `AnnotationStore.add_many()` directly, but still carry `test_triplestore_add_facts*` names/docstrings and the top-level `TripleStore` import is now unused. Also, this removes the direct test coverage that `TripleStore.add_facts()` remains a compatibility shim to `AnnotationStore` during the deprecation window. If that shim compatibility is still important for one release, I would either keep/add a tiny explicit deprecated-shim test elsewhere, or rename these tests fully to `AnnotationStore` so the coverage intent is clear.

coe0718:
Fixing now Thanks

coe0718:
Addressed — renamed both test functions to `test_annotation_store_add_many*`, removed the unused `TripleStore` import, and updated the test name list in the runner. One reference to "deprecated TripleStore.add_facts shim" remains in the docstring as migration context. All 19 tests still passing.

AxDSan (approved):
Clean migration, zero regressions, all tests pass. Deprecation warnings gone from the primary extraction test file.

Three unmigrated callers remain in test_extraction_integration.py — they still work through the shim but emit deprecation warnings. Not blocking, but worth cleaning up in a follow-up PR for consistency.

Thanks coe 🙌

PR #131, opened 2026-05-14
fix: read EMBEDDING_DIM from MNEMOSYNE_EMBEDDING_DIM env var

flooryyyy:
## Problem

`EMBEDDING_DIM` in `beam.py` is hardcoded to `384` (bge-small-en-v1.5), while every other config value in the same block properly reads from an `os.environ` variable with a fallback default. This means users who set `MNEMOSYNE_EMBEDDING_DIM=768` in their env file are silently ignored — the hardcoded 384 always wins.

## Impact

Silent dimension mismatch failure: when the user's embedding model returns vectors of a different dimensionality (e.g. 768-dim from jina-embeddings-v5-nano), the `vec_episodes` table was created with 384-dim slots because the env var was ignored at table creation time. This causes:

- `mnemosyne_sleep` consolidation fails on all sessions with `OperationalError: Dimension mismatch ... Expected 768 dimensions but received 384`
- `mnemosyne_recall` fails with similar dimension mismatch
- No error surfaced to the user — it's a silent failure until they debug the vector table

## Fix

One-line change: read `EMBEDDING_DIM` from `os.environ` with a default of `"384"` for backward compatibility, matching the exact pattern used by every other config value in the same block (lines 226-239).

## Verification

- Existing users with no `MNEMOSYNE_EMBEDDING_DIM` set: no change (defaults to 384 as before)
- Users with `MNEMOSYNE_EMBEDDING_DIM=768`: env var now respected, vectors match table dimensions

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -222,7 +222,7 @@ def _default_db_path() -> Path:
     return _default_data_dir() / "mnemosyne.db"
 
 # Config
-EMBEDDING_DIM = 384  # bge-small-en-v1.5
+EMBEDDING_DIM = int(os.environ.get("MNEMOSYNE_EMBEDDING_DIM", "384"))  # bge-small-en-v1.5
 WORKING_MEMORY_MAX_ITEMS = int(os.environ.get("MNEMOSYNE_WM_MAX_ITEMS", "10000"))
 WORKING_MEMORY_TTL_HOURS = int(os.environ.get("MNEMOSYNE_WM_TTL_HOURS", "24"))
 EPISODIC_RECALL_LIMIT = int(os.environ.get("MNEMOSYNE_EP_LIMIT", "50000"))


steezkelly:
I checked this locally because it has no GitHub check runs attached right now. The one-line fix does what it says:

```bash
MNEMOSYNE_EMBEDDING_DIM=768 /home/steve/.hermes/hermes-agent/venv/bin/python - <<'PY'
from mnemosyne.core import beam
print(beam.EMBEDDING_DIM)
PY
# 768
```

I also ran a nearby env-parser regression slice:

```bash
/home/steve/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_voice_attribution_and_beam_opt.py::TestBeamOptimizationsEnvParser -q
# 21 passed
```

The report/fix looks valid. The only thing I’d add before merge, if you want the repo to lock this behavior down, is a tiny regression test that imports/reloads `mnemosyne.core.beam` with `MNEMOSYNE_EMBEDDING_DIM=768` and asserts `beam.EMBEDDING_DIM == 768`, plus a README / codebase-surface entry so the env var is discoverable. Not blocking from my side; the bug is real and this patch fixes the immediate ignored-env-var problem.

WegoW:
I've stumbled upon a similar issue today, when i set MNEMOSYNE_EMBEDDING_MODEL=openai/text-embedding-3-small on Openrouter. Dimension mismatch 384 vs 1532.

AxDSan (changes requested):
One-line change, right idea, but incomplete.

## The problem

EMBEDDING_DIM is defined in FOUR separate places. This PR only touches one:

| File | Line | Status |
|---|---|---|
| beam.py | 225 | Updated by PR |
| binary_vectors.py | 30 | Still hardcoded 384 |
| shmr.py | 36 | Still hardcoded 384 |
| benchmark_beam_sota.py | 78 | Still hardcoded 384 |

If someone sets MNEMOSYNE_EMBEDDING_DIM=768 and switches models, binary_vectors and shmr still operate at 384d -- silent dimension mismatches, broken vector operations.

## The better approach

Move the env var check into embeddings._get_embedding_dim() so it acts as a single override point:
- env var set -> use env var
- env var not set -> auto-detect from model name -> fallback to 384

Then replace all hardcoded EMBEDDING_DIM = 384 with imports from embeddings.

Also needs: docs update in docs/configuration.md, and a test verifying the env var override works.

AxDSan (changes requested):
One-line change, right idea, but incomplete.

The problem: EMBEDDING_DIM is defined in FOUR separate places. This PR only touches one (beam.py). binary_vectors.py, shmr.py, and benchmark_beam_sota.py still hardcode 384. If someone sets MNEMOSYNE_EMBEDDING_DIM=768, vector ops silently break.

Better approach: move the env var check into embeddings._get_embedding_dim() as a single override point (env var -> auto-detect from model -> fallback 384). Then replace all hardcoded copies with imports from embeddings.

Also needs: docs update in docs/configuration.md, and a test.

AxDSan:
Here's the fix path to make this clean:

**In embeddings.py, update `_get_embedding_dim()` to be the single source of truth:**

```python
def _get_embedding_dim(model_name: str) -> int:
    # Allow env var override
    env_dim = os.environ.get("MNEMOSYNE_EMBEDDING_DIM")
    if env_dim is not None:
        return int(env_dim)
    
    # Auto-detect from model name
    dims = {
        "BAAI/bge-small-en-v1.5": 384,
        "BAAI/bge-base-en-v1.5": 768,
        "BAAI/bge-large-en-v1.5": 1024,
        "openai/text-embedding-3-small": 1536,
        "openai/text-embedding-3-large": 3072,
    }
    return dims.get(model_name, 384)
```

**Then replace the hardcoded copies:**

- `binary_vectors.py:30` → `from mnemosyne.core.embeddings import _get_embedding_dim; EMBEDDING_DIM = _get_embedding_dim(model_name)`
- `shmr.py:36` → same
- `benchmark_beam_sota.py:78` → same
- Keep your `beam.py:225` change but switch to `from mnemosyne.core.embeddings import _get_embedding_dim`

**Add docs**: Entry in `docs/configuration.md` for `MNEMOSYNE_EMBEDDING_DIM`

**Add test**: Verify `os.environ["MNEMOSYNE_EMBEDDING_DIM"] = "768"` overrides correctly.

One file as the single source of truth, env var as override, everyone imports from the same place.

AxDSan:
Superseded by #138 (same fix, updated for current main). Thanks for the contribution @flooryyyy — the approach was right and ether-btc carried it forward.

AxDSan:
My mistake closing this earlier without proper credit. Reopening and closing correctly:

@flooryyyy's original PR was the right approach and the basis for this fix. @ether-btc carried it forward in #138 (merged) with the same solution updated for current main.

Thank you for the contribution @flooryyyy — your work directly shipped in the merged commit. Apologies for the sloppy close earlier.

flooryyyy:
no worries. i actually noticed this quite a few days ago but i didn't bother reporting until i updated mnemosyne, changes got discarded, and it broke again. told my agent to open a PR sine it was a small change and i was too busy to look further, so all credits to @ether-btc :D

PR #133, opened 2026-05-14
fix(provider): stop truncating prefetched memories by default

### LINKED ISSUE #132: [FIX] fix inserted memory truncation
## Problem / Motivation

Memories injected into the `<memory-context>` prompt block are being truncated to 200 characters each. The prefixed timestamp and importance metadata also seem to count against this per-memory limit (but I have not verified). This causes loss of meaning and renders memories useless. 

## Proposed Solution

Quick fix: change `/hermes_memory_provider/__init__.py prefetch()` (lines 791-793) from: 

```
                content = r.get("content", "")[:200]
                if len(r.get("content", "")) > 200:
                    content += "..."
```

to 

`content = r.get("content", "")`

...to stop the truncation completely. 

## Alternatives Considered

It would be more more complex, but adding env vars to control this would add versatility. 

Something like MNEMOSYNE_PREFETCH_CONTENT_CHARS=800, with 0 meaning “no truncation”

Additionally if truncation is to continue happening, it should be on a word or sentence boundary, not at a hard character boundary. 

## Impact

Anyone who is automating mnemosyne's memory creation with an LLM or AAAK. So, effectively everyone. 
Only people who hand-craft their memories down under the limit would be safe from this.

## Willing to Contribute

- [ ] I am willing to submit a PR for this feature
- [ ] I need guidance on where to start
- [X] Just sharing the idea

steezkelly:
## Summary
- confirm and remove the hardcoded 200-character cap from Hermes provider `prefetch()` memory injection
- add `MNEMOSYNE_PREFETCH_CONTENT_CHARS` as an opt-in per-memory cap (`0` / unset = no truncation)
- when a cap is configured, truncate on a word boundary instead of splitting mid-token
- document the env var and add regression coverage for issue #132

## Verification
- `/home/steve/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_prefetch_content_truncation.py tests/test_c13_memory_context_single_injection.py tests/test_c27_provider_init_error_visible.py -q`
- `/home/steve/.hermes/hermes-agent/venv/bin/python -m json.tool mnemosyne_codebase_surface.json >/tmp/mnemosyne_surface_valid.json`

Fixes #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -629,6 +629,7 @@ Importers preserve source metadata where available. `HindsightImporter` uses a d
 | `MNEMOSYNE_RECENCY_HALFLIFE` | `168` | Recency decay halflife in hours (1 week) |
 | `MNEMOSYNE_EP_LIMIT` | `50000` | Episodic memory recall limit |
 | `MNEMOSYNE_SLEEP_BATCH` | `5000` | Max working memories to fetch for consolidation |
+| `MNEMOSYNE_PREFETCH_CONTENT_CHARS` | `0` | Per-memory character limit for Hermes provider prefetch injection; `0` disables truncation |
 
 ### Local LLM (ctransformers/GGUF)
 
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -77,6 +77,42 @@ def _get_triple_module():
     return add_triple, query_triples
 
 
+def _prefetch_content_char_limit() -> int:
+    """Return the per-memory prefetch content limit.
+
+    ``0`` means no truncation. This is the default because the old hardcoded
+    200-character cap often removed the actual fact from LLM-authored memories.
+    Operators that need tighter prompt budgets can set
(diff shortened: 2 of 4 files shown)


AxDSan (approved):
LGTM. This is exactly the right fix at the right layer.

Before this, memories were double-truncated: beam.recall() caps at 500 chars, then the provider choked that down to 200 with a blind [:200] slice. LLM-authored facts after position 200 got silently eviscerated. 

Now: default is no truncation (preserves everything beam returns), and when an operator does need a prompt budget cap, it truncates on word boundaries instead of mid-token. Clean.

Tests are focused and well-scoped. The word-boundary edge case (limit landing exactly on a space) is sub-character-level nitpicking — not a real issue.

One thing to keep an eye on: beam.py still has pervasive content[:500] recall truncation that's separate from this provider fix. Different layer, different battle — but worth a follow-up ticket.

Thanks for catching and fixing this steezkelly 🎉

PR #138, opened 2026-05-14
fix: read EMBEDDING_DIM from MNEMOSYNE_EMBEDDING_DIM env var

ether-btc:
## Summary

Makes  configurable via  env var instead of hardcoded at 384. Follows PR #131 (flooryyyy) approach.

## Changes

**mnemosyne/core/beam.py:**
```python
try:
    _emb_dim = int(os.environ.get("MNEMOSYNE_EMBEDDING_DIM", "384"))
    EMBEDDING_DIM = _emb_dim if _emb_dim > 0 else 384
except ValueError:
    EMBEDDING_DIM = 384  # bge-small-en-v1.5
```

- Reads from  env var (default 384)
- Guards against zero/negative values (falls back to 384)
- Guards against non-integer garbage (falls back to 384, no crash)

**tests/test_beam.py:**  — 3 tests:
-  — verifies default is 384
-  — verifies it's an assignable int
-  — verifies correct int parser

## Why

vec0 schema dimensions must match the embedding model's actual output. Hardcoding 384 causes sqlite errors at table-creation time for operators using different models (e.g., 768 for text-embedding-3-small).

## CI

81 tests pass (TestEmbeddingDimConfig 3 + TestBeamSchema 10 + TestSleepCycle 2 + test_plugins.py 70).

Closes #131


AxDSan:
Note: This fix was originally contributed by @flooryyyy in #131. Thanks to both contributors — @flooryyyy for the original approach and @ether-btc for carrying it forward to merge.

PR #143, opened 2026-05-15
fix: add strict fact annotation matching

wysie:
## Summary
- Add an opt-in strict fact annotation matcher behind `MNEMOSYNE_STRICT_FACT_MATCH=1`
- Tokenize fact/query text, ignore common stopwords, and require stronger matches before admitting fact candidates
- Preserve existing fact matching behavior when the flag is unset

## Why
The current fact matcher admits a memory if any query word appears as a substring in a fact annotation. Natural-language queries often include common words like "where", "the", "what", or "use", which can pull unrelated facts into the recall candidate set.

## Safety
- Default behavior is unchanged unless `MNEMOSYNE_STRICT_FACT_MATCH=1` is set
- Exact phrase matches still work
- Distinctive single tokens such as paths, domains, and long identifiers remain supported

## Test Plan
- `python -m pytest tests/test_beam.py::TestFactAnnotationMatching -q`
- `python -m py_compile mnemosyne/core/beam.py tests/test_beam.py`

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -182,6 +182,7 @@ DEFAULT_DATA_DIR = Path.home() / ".hermes" / "mnemosyne" / "data"
 DEFAULT_DB_PATH = DEFAULT_DATA_DIR / "mnemosyne.db"
 
 import os
+import re
 
 def _env_truthy(name: str) -> bool:
     """Parse an env var as truthy. Accepts `1`/`true`/`yes`/`on`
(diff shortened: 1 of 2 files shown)


ether-btc (approved):
## Contributor Review: Approved with one concern

Reviewed the full diff. Good direction — strict fact matching solves real recall noise.

**What works:**
- Stopword exclusion in token set — correct
- Exact phrase fallback — correct
- Multi-token overlap check — correct
- Default-off safety gate — correct

**One concern to address before merge:**

The single-token relaxation is too permissive. Currently accepts ANY token with length >= 8, even bare lowercase strings with no path/domain/version structure:

```
return len(token) >= 8 or "." in token or "/" in token ...
```

A token like "abcdefgh" (8 random lowercase letters) would pass on length alone, but has zero distinctive signal. Meanwhile pure-digit tokens like "127", "0", "0", "1" (from splitting "127.0.0.1") enter the overlap set and can skew results.

**Fix:** Require AND, not OR:
- `len(token) >= 8 AND any structural marker (.:/-_)`
- Add `and not token.isdigit()` to the token filter

I've submitted this as **PR #150** — a minimal 2-change patch that applies cleanly to your branch. The tests pass with the changes.

**Verdict: Approve subject to PR #150 landing first** (or fold its logic in). The direction is right; the single-token threshold needs tightening.

PR #144, opened 2026-05-15
fix: index Hindsight imports for semantic recall

WXBR:
## Summary

Fixes Hindsight imports so direct episodic imports are indexed for semantic recall instead of only preserving rows/FTS content.

Changes:
- generate sqlite-vec and binary vectors for newly inserted Hindsight episodic memories when embedding support is available
- preserve Hindsight provenance (`source = hindsight:<fact_type>`, `veracity = imported`, original metadata)
- add import quality flags and an import quality score to metadata
- add optional `skip_low_value` filtering for obvious memory-review/meta/junk rows
- keep embedding generation best-effort so imports still work without optional embedding/vector dependencies

## Why

The direct Hindsight importer preserved data, timestamps, and FTS, but newly imported episodic rows could miss the vector indexes used by semantic recall. That means the memories existed, but semantic search could not fully see them.

## Tests

```bash
python -m pytest tests/test_importers/test_hindsight.py -q -o 'addopts='
# 5 passed
```

### DIFF
--- a/mnemosyne/core/importers/hindsight.py
+++ b/mnemosyne/core/importers/hindsight.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import re
 import urllib.parse
 import urllib.request
 from datetime import datetime, timezone
@@ -25,6 +26,16 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 from mnemosyne.core.importers.base import BaseImporter, ImporterResult
+
+try:  # Optional semantic-search backfill for direct episodic imports.
+    from mnemosyne.core import embeddings as _embeddings
+    from mnemosyne.core.beam import _vec_available, _vec_insert
+    from mnemosyne.core.binary_vectors import maximally_informative_binarization as _mib
+except Exception:  # pragma: no cover - importer must still work without embeddings extras
+    _embeddings = None
+    _vec_available = None
+    _vec_insert = None
+    _mib = None
 
 
 class HindsightImporter(BaseImporter):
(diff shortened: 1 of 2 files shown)


ether-btc (approved):
## Contributor Review: Approved with suggestions

Reviewed the full diff. The implementation is solid:

**What works:**
- Direct FTS5 + vector backfill on import — closes the semantic recall gap correctly
-  filtering with metadata scoring — clean separation
- Best-effort fallback when embeddings unavailable — import never fails on optional deps

**Two improvement opportunities (non-blocking):**

1. **Silent failure in _backfill_import_embedding** — the function has 3 early-return paths with no logging:
   - embeddings unavailable → silent return
   - embed() returns None → silent return  
   - UPDATE binary_vector fails → propagates (correct)
   
   Add  calls to the first two return paths so operators can diagnose import quality issues.

2. **binary_vector UPDATE has no error handling** — if the column doesn't exist in an older schema, this will raise an exception. Wrap in  with a debug log, or check schema before updating.

I've submitted these as **PR #149** on top of your branch — a minimal diagnostic patch (logging + error handling) that applies cleanly to your diff. Feel free to land it separately or fold it in.

**Verdict: Approve**. Good to merge as-is; the two items above are nice-to-have improvements, not blockers.

ether-btc (approved):
## Contributor Review: Approved with suggestions

Reviewed the full diff. The implementation is solid.

**What works:**
- Direct FTS5 + vector backfill on import — closes the semantic recall gap correctly
- skip_low_value filtering with metadata scoring — clean separation
- Best-effort fallback when embeddings unavailable — import never fails on optional deps

**Two improvement opportunities (non-blocking):**

1. **Silent failure in _backfill_import_embedding** — the function has 3 early-return paths with no logging:
   - embeddings unavailable → silent return
   - embed() returns None → silent return
   - UPDATE binary_vector fails → propagates (correct)

   Add logger.debug calls to the first two return paths so operators can diagnose import quality issues.

2. **binary_vector UPDATE has no error handling** — if the column doesn't exist in an older schema, this will raise an exception. Wrap in try/except with a debug log.

I've submitted these as PR #149 on top of your branch — a minimal diagnostic patch (logging + error handling) that applies cleanly to your diff. Feel free to land it separately or fold it in.

**Verdict: Approve**. Good to merge as-is; the two items above are nice-to-have improvements, not blockers.

PR #146, opened 2026-05-16
feat: proactive memory linking (Phase 5)

coe0718:
## Overview

This PR implements **proactive memory linking** — the natural follow-up to PR #128. When a new memory is stored via `remember()`, the system automatically creates weighted graph edges to related existing memories.

Two zero-LLM strategies, non-blocking, gated behind `MNEMOSYNE_PROACTIVE_LINKING=1`.

### 1. Content Similarity (FTS5)
Builds a broad OR query from content keywords (stop words removed, FTS5-safe only), runs a MATCH against `fts_working`, and creates `related_to` edges with rank-decayed weights.

### 2. Entity Overlap (Annotations)
Queries the `annotations` table for `kind="mentions"` entries shared with other memories, and creates `references` edges at weight 0.8.

## Tests
11 new tests: content similarity, entity overlap, unrelated content, self-reference avoidance, env var gating, edge deduplication, non-blocking, edge types.

All 105+ existing tests pass. Backward compatible — no schema changes, no behavior changes when disabled.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -2057,6 +2057,122 @@ class BeamMemory:
             except Exception:
                 pass  # Veracity failures are non-blocking
 
+        # Phase 5: Proactive linking — auto-create graph edges to related memories
+        self._proactively_link(memory_id, content)
+
+    def _proactively_link(self, memory_id: str, content: str):
+        """Phase 5: Auto-create graph edges between new memory and related existing memories.
+
+        Two zero-LLM strategies:
+        1. Content similarity via recall() — top-K via FTS5 + vector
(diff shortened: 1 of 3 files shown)


coe0718:
This is the final addition to the memory system. Drafted whenever you’re ready.

AxDSan:
Merged into experiment/recursive-recovery-v2 with authorship preserved. Cherry-picked commit b9be637 + ether-btc's strict matching hardening applied as follow-up. Will land on main when experiment branch graduates.

Thanks @coe0718 for Phase 5 — the proactive linking is a solid architectural foundation.

coe0718:
Awesome thanks!

PR #150, opened 2026-05-17
fix(beam): tighten strict fact matching — drop pure-digit tokens and require structural markers

ether-btc:
## Summary

Two targeted fixes to the strict fact annotation matcher (behind MNEMOSYNE_STRICT_FACT_MATCH=1):

1. Exclude pure numeric tokens from the token set — IPs (127.0.0.1), port numbers (8765), and version strings get split into atoms like 127, 0, 0, 1 by the tokenizer. These pollute the overlap count and can cause false-positive matches.

2. Require BOTH length >= 8 AND at least one structural marker (.:/-_) for single-token matches. Previously a bare 8-char lowercase string like abcdefgh would pass on length alone. A genuine single-token match needs a path/domain/version structure.

## Changes

- In _fact_match_tokens: add `and not token.isdigit()` filter
- In _strict_fact_matches single-token branch: replace OR chain with `len(token) >= 8 and has_structure` (AND)

## Motivation

The single-token relaxation was too permissive — it let bare 8-char strings through without structural markers. Meanwhile pure-digit tokens from IPs/versions were entering the token set and skewing recall. Both bugs cause recall noise in opposite directions.

## Testing

python -m pytest tests/test_beam.py::TestFactAnnotationMatching -q — all 3 existing tests pass.

---

Contributor review of upstream PR #143 — AxDSan/mnemosyne

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -182,6 +182,7 @@ DEFAULT_DATA_DIR = Path.home() / ".hermes" / "mnemosyne" / "data"
 DEFAULT_DB_PATH = DEFAULT_DATA_DIR / "mnemosyne.db"
 
 import os
+import re
 
 def _env_truthy(name: str) -> bool:
     """Parse an env var as truthy. Accepts `1`/`true`/`yes`/`on`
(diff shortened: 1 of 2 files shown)


ether-btc (commented):
## Contributor Review

Tightens the strict fact matcher: excludes pure-digit tokens from overlap counting, and requires structural markers for single-token matches. All existing tests pass. Good to merge alongside the parent PR.

ether-btc:
Linked to upstream #143 — this patch tightens the strict fact matcher by excluding pure-digit tokens and requiring structural markers for single-token matches.

ether-btc:
Fixes #143 — tightens strict fact matcher (digit exclusion + structural AND logic)

AxDSan:
Closing in favor of #143 (original author: @wysie). The two targeted fixes in this PR (pure-digit token filter and AND condition for single-token structural markers) are solid catches. We'll apply them as a follow-up commit to the experiment branch with attribution. Thanks for the review @ether-btc!

PR #156, opened 2026-05-20
Fix temporal recall timezone normalization

zombopanda:
## Summary

- Normalize temporal query times and parsed memory timestamps to UTC-aware datetimes before temporal scoring.
- Treat naive `datetime` values as UTC for backward compatibility.
- Add regression coverage for offset-aware imported timestamps with the default `query_time=None` path.

## Bug

Temporal recall can crash when a memory timestamp is offset-aware but the query time is naive.

This shows up with Hindsight imports because Hindsight memory rows commonly preserve ISO timestamps like:

```text
2026-05-20T08:29:42.958511+00:00
```

Mnemosyne parses those as offset-aware datetimes. But when recall is called without an explicit `query_time`, `_parse_query_time(None)` returned `datetime.now()`, which is offset-naive. With temporal scoring enabled, `_temporal_boost()` then compares the two:

```python
if ts > query_time:
```

Python raises:

```text
TypeError: can't compare offset-naive and offset-aware datetimes
```

## Impact

The most visible failure is in Hermes' `MnemosyneMemoryProvider.prefetch()` path. It calls recall with a temporal boost, catches exceptions, and returns an empty context block. So the agent can appear to have working memory tools while pre-turn memory injection silently stops working.

The same failure is reproducible through `mnemosyne_recall` with `temporal_weight > 0` on imported offset-aware data. Recall without temporal scoring can still work, which makes the bug easy to miss in basic store/recall tests.

## Repro Shape

A minimal failing shape before this PR:

```python
from mnemosyne.core.beam import _parse_query_time, _temporal_boost

query_time = _parse_query_time(None)  # previously naive datetime.now()
_temporal_boost("2026-05-20T08:29:42.958511+00:00", query_time, 48)
```

In the Hermes provider path this appeared as:

```python
mnemosyne_recall({"query": "pandabot", "limit": 2, "temporal_weight": 0.2})
```

returning the TypeError, while `prefetch()` returned `""` because the provider swallowed the exception.

## Fix

- `_parse_query_time(None)` now returns an aware UTC `datetime`.
- Explicit naive query times are normalized to UTC-aware datetimes.
- Parsed memory timestamps are normalized before comparison.
- The temporal boost behavior remains compatible for naive historical values by treating them as UTC.

## Tests

Added/updated regression coverage in `tests/test_temporal_recall.py` for:

- default query time with offset-aware timestamps;
- explicit naive query time with offset-aware timestamps;
- temporal recall over imported-style offset-aware memories;
- future timestamp clamping after normalization.

Validation run:

- `uv run pytest tests/test_temporal_recall.py`
- `uv run pytest tests/test_configurable_scoring.py tests/test_prefetch_content_truncation.py`
- `PYTHONPATH=. uv run pytest tests/test_beam.py`

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -25,7 +25,7 @@ import threading
 import math
 
 logger = logging.getLogger(__name__)
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import List, Dict, Optional, Any, Set, Union
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


AxDSan:
Good stuff @zombopanda. The timezone thing was eating context silently. Nasty bug. Thanks for digging into it.

PR #157, opened 2026-05-20
fix: add missing 'instruction' keys and German false positives in MULTILINGUAL_PATTERNS

WegoW:
The  dict was missing the `'instruction'` key for both `'en'` and `'de'` language entries. This caused a `KeyError` when `extract_and_store_facts()` called `pat['instruction'].replace('IMPVERBS', ...)`, silently breaking instruction extraction for all languages.

Also:
- Populated DE `instruction_false_positives` with `['du solltest gehen']` (was empty `[]`)
- Expanded the bare-question filter to match both `should` and `sollte` with English/German pronouns

The 'instruction' patterns themselves were already implemented in the multilingual refactor (commit ad6fe05) — they just weren't registered in the pattern dict.

Ref: https://github.com/AxDSan/mnemosyne/commit/ad6fe05


AxDSan:
Clean fix @WegoW. The KeyError was hitting anyone using multilingual patterns without 'instruction' keys. The German false-positive filter was a nice extra. Your PR also exposed a test import that was already broken on main. Got that cleaned up too. Win/win.

PR #161, opened 2026-05-21
feat(embeddings): custom API endpoints, Jina model dims, SSL cert config

flooryyyy:
## Summary

Four small improvements to `embeddings.py` that make Mnemosyne work better with self-hosted/compatible embedding servers and NixOS environments.

### Changes

1. **Custom endpoint detection** (`_is_api_model()`): When `OPENROUTER_BASE_URL` is set to a non-OpenRouter URL, treat any model name as API mode. This lets users run local/self-hosted embedding servers (llama-server, vLLM, etc.) without having to rename their models to match OpenAI patterns like `text-embedding-*`.

2. **Jina model dimensions** (`_get_embedding_dim()`): Added `jina-embeddings-v5-omni-nano` (768d) and `jina-embeddings-v5-omni-small` (1024d) to the built-in dimension table. These are popular multimodal embedding models.

3. **SSL certificate configuration**: `_embed_api()` now checks `SSL_CERT_FILE` and `REQUESTS_CA_BUNDLE` env vars for custom CA bundles. NixOS and enterprise environments need this because Python's `urllib` can't find system certificates by default.

4. **Auth-optional for custom endpoints**: `_embed_api()` and `available()` no longer require an API key when the endpoint is a custom (non-OpenRouter) URL. Local servers typically don't need authentication.

### Test Plan
- [ ] `_is_api_model` still returns true for existing OpenAI model patterns
- [ ] Dimension lookup returns 768 for jina-nano, 1024 for jina-small
- [ ] `SSL_CERT_FILE=/path/to/ca-bundle.crt` → `load_verify_locations()` called
- [ ] Custom endpoint + no API key → `_embed_api()` proceeds (no auth header)
- [ ] Custom endpoint with `jina-embeddings-v5-omni-nano` model name → routes to API

### Migration
No breaking changes. Default behavior unchanged for existing OpenRouter users.

### DIFF
--- a/mnemosyne/core/embeddings.py
+++ b/mnemosyne/core/embeddings.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import json
 import os
+import ssl
 import urllib.request
 from typing import List, Optional
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


AxDSan:
Good stuff @flooryyyy. Four clean improvements: custom endpoint detection, Jina dims, SSL cert support, and auth-optional for self-hosted servers. Tests are thorough. Merged.

PR #167, opened 2026-05-24
fix(memory): load sqlite-vec extension in memory._get_connection

theheavenlyd3mon:
memory._get_connection() creates connections without loading the sqlite-vec extension, while beam._get_connection() does load it.

This means Mnemosyne.remember() writes embeddings to the legacy memory_embeddings table (JSON fallback) but cannot insert into the vec_episodes virtual table, breaking vector search for the high-level API.

The fix adds the same sqlite-vec loading that beam._get_connection uses:
  try:
      import sqlite_vec
      conn.enable_load_extension(True)
      sqlite_vec.load(conn)
  except Exception:
      pass

Closes the gap between the two connection factories so vector search works regardless of which code path creates the connection.


AxDSan (approved):
7 lines, fixes a real connection issue. Simple, clean, targeted. Thanks Noctis! 🙏

PR #169, opened 2026-05-25
fix: report active provider DB path in diagnose

achrllrogia45:
## Summary

When profile isolation is enabled, `mnemosyne_diagnose` only reports the legacy/default DB path from `run_diagnostics()`. The active provider uses a profile bank at a different path (e.g. `data/banks/sisyphus/mnemosyne.db`), which causes operator confusion.

This PR adds:
- `active_provider_db_path` — the actual DB the provider is reading/writing
- `profile_isolation_enabled` — whether profile isolation is active
- `active_provider_counts` — working_memory, episodic_memory, facts counts from the active DB
- `active_provider_counts_error` — graceful error if count query fails

Legacy diagnostic output is unchanged.

## Tests

New file: `tests/test_hermes_memory_provider_diagnose_active_db.py`

Covers:
- Active DB path and counts appear when beam is initialized
- No active beam keeps base diagnostics unchanged
- Count query failure reports error without crashing

```
40 passed in 6.34s
```

## Changes

- `hermes_memory_provider/__init__.py` — 32 insertions in `_handle_diagnose`
- `tests/test_hermes_memory_provider_diagnose_active_db.py` — 82 lines new

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1480,6 +1480,38 @@ class MnemosyneMemoryProvider(MemoryProvider):
     def _handle_diagnose(self, args: Dict[str, Any]) -> str:
         from mnemosyne.diagnose import run_diagnostics
         result = run_diagnostics()
+
+        # run_diagnostics() reports Mnemosyne's legacy/default DB path. When
+        # Hermes profile isolation is enabled, the active provider may use a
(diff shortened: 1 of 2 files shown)


AxDSan (approved):
Clean PR. Solves a real ops pain — diagnose now shows the actual active provider DB path instead of the default one. Tests included (40 passing). Nice work Hilman! 🙏

PR #172, opened 2026-05-25
fix(core): generate and save working memory vector on remember

Gautam-J:
## Summary
This PR resolves a silent degradation in working memory vector retrieval (hybrid search). 

Currently, while `remember_batch` generates and saves embeddings into the `memory_embeddings` table, the single-item `remember()` method only commits the text to `working_memory` and completely skips the embedding generation stage. 

Because `mnemosyne_recall` (and linear recall paths) rely on joining `working_memory` with `memory_embeddings` during vector similarity calculations (using `_wm_vec_search`), any memory stored via conversational single-item writes (the standard conversational workflow) ended up with **0** vector representation and was missed entirely by semantic searches.

## Changes
- Added embedding generation in `BeamMemory.remember()` right after writing to `working_memory`.
- Successfully writes the generated F32 dense vector to the `memory_embeddings` table, making all newly stored single-item memories immediately discoverable via dense vector similarity matching.

## Verification
- Verified locally that conversational writes (using `mnemosyne_remember`) now generate vectors in `memory_embeddings` automatically.
- Verified that subsequent `mnemosyne_recall` queries successfully match using semantic vector scores with a `dense_score` of `0.7271` instead of `0.0000`.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1915,6 +1915,24 @@ class BeamMemory:
         self.conn.commit()
         self._trim_working_memory()
 
+        # Generate vector embedding for working memory hybrid search
+        if _embeddings.available():
+            try:
+                vec = _embeddings.embed([content])
+                if vec is not None and len(vec) > 0:
+                    model = _embeddings._DEFAULT_MODEL
+                    emb_json = _embeddings.serialize(vec[0])
+                    cursor.execute(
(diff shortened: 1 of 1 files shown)


AxDSan (approved):
LGTM. This catches a real silent data loss bug — single-item remember wasn't generating embeddings, making those memories invisible to vector search. 18 lines, clean, focused. Verified locally with dense_score 0.7271 vs 0.0000 before. Thanks Gautam! 🙏

PR #181, opened 2026-05-26
feat: unified recall merges shared surface results

achrllrogia45:
Reopens #177 (closed by the v3.1.0 history rewrite) on a fresh base, with a benchmark added per CONTRIBUTING rule #2.

Adds `shared_surface_read` config flag. When enabled, `mnemosyne_recall` returns results from both the private bank and the shared surface in one call. Each result is tagged with `bank` ("private" or "surface") so callers can tell where it came from. Default stays off  existing setups keep current behavior.
example case:

Why this matters: with profile isolation on, every agent has its own private bank, but they share one surface DB for cross-agent facts. Today an agent has to choose between `mnemosyne_recall` and `mnemosyne_shared_recall` upfront. With this flag on, recall just searches both and merges by score  agents get the full picture without picking the right tool.

Behavior:
- `shared_surface_read=false` (default): recall stays private-only, results tagged `bank=private`
- `shared_surface_read=true`: also searches the surface beam, merges both lists by score, truncates to `top_k` overall, surface rows tagged `bank=surface` and `shared_surface=true`
- Surface init or recall failure logs a warning and falls back to private-only  never breaks recall

Resolution order is `kwargs > config.yaml > default false`, same pattern as the other shared-surface options. `mnemosyne_shared_recall` is unchanged.

## Benchmark

`tools/bench_unified_recall.py` measures `mnemosyne_recall` p50/p95 with the flag off vs on.

Local run on this branch (200 private rows, 50 shared rows, 100 queries):

```
baseline (private_only):  p50=3.94ms  p95=19.2ms
merged   (with surface):  p50=3.81ms  p95=19.0ms
delta:                    -0.13ms p50  (-3.25%)
```

The merged path adds one extra `surface_beam.recall()` call plus a Python sort over the combined list. Cost is within measurement noise here. Anyone reviewing can reproduce with `python tools/bench_unified_recall.py`.

## Tests

`tests/test_hermes_memory_provider_unified_recall.py` — 10 tests:

- default behavior keeps recall private-only
- private results get `bank=private` tag
- surface read on includes shared rows
- surface rows tagged `bank=surface` + `shared_surface=true`
- merged results contain both banks
- top_k truncation applies after merge
- surface init failure falls back to private-only
- config schema entry exists with default false
- config.yaml `shared_surface_read: true` is read at init
- kwarg overrides config.yaml

```
tests/test_hermes_memory_provider_unified_recall.py ..........  [10 passed]
+ broader regression suite (shared_crud, shared_multi_agent,
  temporal_recall, all_15_tools): 71 passed total, 0 regressions
```

## Changes

- `hermes_memory_provider/__init__.py` — 50 insertions, 1 deletion (flag + merge logic + schema entry)
- `tests/test_hermes_memory_provider_unified_recall.py` — 202 lines new
- `tools/bench_unified_recall.py` — 152 lines new

Total: 403 insertions across 3 files. No schema migration. Backward compatible. Single feature, single commit.

Refs the prior closed PR #177; rebased onto v3.1.0.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -568,6 +568,12 @@ class MnemosyneMemoryProvider(MemoryProvider):
         self._surface_beam: Optional[Any] = None
         self._shared_surface_bank = "surface"
         self._shared_surface_path: Optional[Path] = None
+        # When true, mnemosyne_recall merges shared-surface results into the
+        # private bank's recall response. Each result is tagged with `bank`
+        # ("private" or "surface") so callers can distinguish provenance.
+        # Default false preserves existing behavior for deployments that have
+        # not opted in.
+        self._shared_surface_read = False
         # C27: capture init exception so downstream methods can surface it
         # instead of silently no-op'ing. `_beam is None AND _init_error is None`
         # means a deliberate skip (subagent/cron/skill_loop context, or pre-init);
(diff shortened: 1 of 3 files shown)


AxDSan:
Clean work @achrllrogia45. Benchmark included, tests tight, opt-in flag so nobody gets surprised. The surface/private merge by score is exactly how it should work.

Merged with rebase. Thanks!

PR #179, opened 2026-05-26
fix: improve recall precision

WXBR:
## Summary
- Tightens recall lexical gates so broad/noisy queries abstain instead of returning weak one-token matches.
- Improves multi-aspect recall so distinct relevant facts can survive top-k truncation without flooding with duplicate/checklist-style rows.
- Backfills imported Hindsight embeddings into the canonical `memory_embeddings` table and supplements MEMORIA recall with source rows when relevant.
- Adds generic regression coverage using filler fixtures only.

## Validation
- `PYTHONPATH=. python3 -m pytest tests/test_recall_precision_regressions.py tests/test_c4_recall_diagnostics.py tests/test_beam_e3_additive_sleep.py tests/test_hermes_memory_provider.py tests/test_hermes_memory_provider_diagnose_active_db.py tests/test_hermes_memory_provider_thread_isolation.py tests/test_hermes_plugin_tools.py -q`
  - 114 passed
- `ulimit -n 8192 2>/dev/null || true; PYTHONPATH=. python3 -m pytest tests/ -q`
  - 1335 passed, 18 skipped

## Privacy
- Regression fixtures are generic filler data.
- Candidate diff was scanned for private markers, high-confidence secret assignments, and email addresses before commit.

PR #182, opened 2026-05-26
feat: add memory audit log for mutation traceability

achrllrogia45:
Reopens #176 (closed by the v3.1.0 history rewrite) on a fresh base, with a benchmark added per CONTRIBUTING rule #2.

Adds an audit log so every memory mutation is recorded for traceability. Useful for debugging cross-agent operations, reconstructing intent over time, and answering "who wrote this / when / why" against shared surface entries which is the natural follow-up to v3.1.0's shared CRUD now that multiple agents can write to the same surface.
>user: i prefer VPN x
>agent a: `write memory user prefer x`
>agent b: <thinking>this memory from agent a, user prefer VPN x</thinking> 

A new SQLite table `memory_events` is created alongside the existing schema; every mutation tool (`remember`, `forget`, `invalidate`, `shared_remember`, `shared_forget`, `sleep`, `import`) emits a row capturing `action`, `memory_id`, `bank` (private | surface), `source_tool`, and timestamp. No-op operations (forget on a missing ID, dry runs, failed writes) are not audited.

This PR adds:
- `hermes_memory_provider/audit.py` `AuditLog` class, `memory_events` table schema, append-only writes
- `_init_audit_log` and `_audit_event` hooks in the provider
- Hooks wired into all mutation tools listed above
- Audit writes are best-effort: a failure to record never blocks the underlying mutation
- Tests + benchmark for the new code path

Existing tools and behavior are unchanged.

## Benchmark

`tools/bench_audit_log.py` measures `remember()` latency with the audit log toggled off vs on, using one provider and interleaved chunks to cancel cold-start asymmetry.

Local run (200 ops per config, interleaved in chunks of 10):

```
audit_off:  p50=1.02ms  p95=1.34ms  mean=1.03ms
audit_on:   p50=1.20ms  p95=5.54ms  mean=2.49ms
delta:      +0.18ms p50  (+17.7%)   p95 grows ~4x
```

Honest disclosure: typical-case overhead is small (~0.2ms p50). Tail latency (p95) grows more the audit log opens its own SQLite connection and the per-mutation INSERT can occasionally block on the audit DB's WAL. Acceptable for traceability, and it never blocks the underlying mutation, but reviewers should know the tail is not flat. Reproducible with `python tools/bench_audit_log.py`.

If this overhead is unacceptable for hot-path use, a follow-up could batch audit inserts or write asynchronously. Out of scope for this PR.

## Tests

New file: `tests/test_hermes_memory_provider_audit.py` 12 tests covering:

- Audit table is created on init
- `remember` and `shared_remember` emit correct `action` + `bank` + `source_tool`
- `forget` and `shared_forget` audit only when an actual deletion happens
- `invalidate` records the lifecycle change
- `sleep` consolidation events recorded
- `import` events recorded with provenance
- Audit failure path: mutation still succeeds when audit insert fails
- Multi-agent scenario: separate banks audit independently, shared surface shared

```
12 audit tests + 37 broader tests = 49 passed
```

## Changes

- `hermes_memory_provider/__init__.py` — 57 insertions (audit init + hook calls)
- `hermes_memory_provider/audit.py` — 126 lines new
- `tests/test_hermes_memory_provider_audit.py` — 165 lines new
- `tools/bench_audit_log.py` added in this rebased version

Total feature: 348 insertions across 3 files. No schema migration to existing tables. Backward compatible.

Refs the prior closed PR #176; rebased onto v3.1.0.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -568,6 +568,7 @@ class MnemosyneMemoryProvider(MemoryProvider):
         self._surface_beam: Optional[Any] = None
         self._shared_surface_bank = "surface"
         self._shared_surface_path: Optional[Path] = None
+        self._audit: Optional[Any] = None
         # C27: capture init exception so downstream methods can surface it
         # instead of silently no-op'ing. `_beam is None AND _init_error is None`
         # means a deliberate skip (subagent/cron/skill_loop context, or pre-init);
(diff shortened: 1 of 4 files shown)


AxDSan:
Clean work @achrllrogia45. Own SQLite connection, best-effort writes, hooks on all 7 mutation paths. The p95 tail is duly noted but acceptable for the traceability this unlocks.

Merged (resolved minor init conflict with #179's field addition). Thanks!

Issue #193, opened 2026-05-28
feat(embeddings): add MNEMOSYNE_EMBEDDING_API_URL/KEY as preferred env var names


flooryyyy:
PR #161 added custom API endpoint support but used `OPENROUTER_BASE_URL` and `OPENROUTER_API_KEY` as the env var names. These names are misleading — they imply OpenRouter-specific usage, but the endpoint is just a generic OpenAI-compatible API. This also conflicts if someone uses OpenRouter for their LLM but a different endpoint (e.g., local llama-swap) for embeddings.

This PR adds `MNEMOSYNE_EMBEDDING_API_URL` and `MNEMOSYNE_EMBEDDING_API_KEY` as the preferred names, restoring backward compatibility with v2.8.x naming convention. The old `OPENROUTER_*` names continue to work as fallbacks.

No breaking changes — existing configs continue to work.

Changes:
- `mnemosyne/core/embeddings.py`: env var lookup chain, now checks MNEMOSYNE_EMBEDDING_API_URL/KEY first
- `mnemosyne/__init__.py`: bump 3.1.0 → 3.2.0
- `CHANGELOG.md`: added v3.2.0 entry
- `README.md`: added env vars to Environment Variables table
- `docs/configuration.md`: added Custom Embedding Endpoint section

AxDSan:
Fixed in #194. @flooryyyy, merged with the env var rename and version pinned to 3.1.1. Thanks for catching the misleading OPENROUTER naming.

PR #194, opened 2026-05-28
feat(embeddings): add MNEMOSYNE_EMBEDDING_API_URL/KEY as preferred env var names

flooryyyy:
## Description

PR #161 added custom API endpoint support but used `OPENROUTER_BASE_URL` and `OPENROUTER_API_KEY` as the env var names. These names are misleading — they imply OpenRouter-specific usage, but the endpoint is just a generic OpenAI-compatible API. This also conflicts if someone uses OpenRouter for their LLM but a different endpoint (e.g., local llama-swap) for embeddings.

This PR adds `MNEMOSYNE_EMBEDDING_API_URL` and `MNEMOSYNE_EMBEDDING_API_KEY` as the preferred names, restoring backward compatibility with v2.8.x naming convention. The old `OPENROUTER_*` names continue to work as fallbacks.

**No breaking changes** — existing configs continue to work. The new names are checked first, falling back to the old names if not set.

Also fixes the CHANGELOG v3.1.0 entry that referenced `MNEMOSYNE_EMBEDDING_BASE_URL` but never implemented it.

## Changes
| File | Change |
|------|--------|
| `mnemosyne/core/embeddings.py` | Added MNEMOSYNE_EMBEDDING_API_URL/KEY as preferred env var names with OPENROUTER_* fallback in all 4 read sites |
| `mnemosyne/__init__.py` | Bump 3.1.0 → 3.2.0 |
| `CHANGELOG.md` | Added v3.2.0 entry referencing #193 |
| `README.md` | Added embedding env vars to Environment Variables table |
| `docs/configuration.md` | Added Custom Embedding Endpoint section |

## Related
- Issue: #193
- PR #161 (original custom endpoint support)
- PR #187 (MNEMOSYNE_EMBEDDINGS_VIA_API opt-in)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Simple Versioning](https://github.com/AxDSan/mnemosyne) (MAJOR.MINOR).
 
+## [3.2.0] — 2026-05-28
+
+### Added
+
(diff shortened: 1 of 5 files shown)


flooryyyy:
hmm if everything is set in ~/.hermes/.env, having OPENROUTER_BASE_URL or OPENROUTER_BASE_URL could potentially interfere if backwards compatibility is implemented

this was an oversight on my part. what's the preferred way forward?

PR #201, opened 2026-05-29
Update installation script link

0rangeFox:
## Description

As the title says, it's using the outdated url for "One-Liner Deploy" and it's updated for new script url also with an suggestion on `Hermes Integration` section.

## Related Issue

Closes #(issue)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

## Checklist

- [ ] `CHANGELOG.md` updated with a brief entry
- [ ] Code follows the project's principles:

### DIFF
--- a/docs/integrations/hermes-mcp.md
+++ b/docs/integrations/hermes-mcp.md
@@ -21,7 +21,7 @@ Tools register as native Hermes commands.
 If you installed Mnemosyne via `deploy_hermes_provider.sh`, it's already active:
 
 ```bash
-curl -sSL https://raw.githubusercontent.com/AxDSan/mnemosyne/main/deploy_hermes_provider.sh | bash
+curl -sSL https://raw.githubusercontent.com/AxDSan/mnemosyne/main/scripts/install.sh | bash
 ```
 
 This symlinks the provider into `~/.hermes/plugins/mnemosyne`.


0rangeFox:
I think we could use this PR to update some instructions about installing because for me is a bit confuse, looks there's 3 ways to setup it for "Hermes Agent".

1. The first one, through the README, on section _[Hermes Plugin (23 tools)](https://github.com/AxDSan/mnemosyne/blob/main/README.md#hermes-plugin-23-tools)_, using the `pip install mnemosyne-hermes` for all-in-one, and below on that _description_ there's a link where forwards to 3rd point.
2. On _[Integration Documents](https://github.com/AxDSan/mnemosyne/blob/main/docs/integrations/README.md)_, also there's a option for "Hermes Agent" instructions which also redirect to another "all-in-one" tool by using the script provided on this repo.
3. The last one, _[Hermes Integration](https://github.com/AxDSan/mnemosyne/blob/main/docs/hermes-integration.md)_ inside of _[Documents](https://github.com/AxDSan/mnemosyne/blob/main/docs)_ folder, has totally another way to install the memory system and plugin.

Also, I would like to report too, I'm on MacOS, with [Hermes Agent v0.15.1 (v2026.5.29)](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.5.29) and installed via their bash script, and when I install this memory system through the [`install.sh` script](https://github.com/AxDSan/mnemosyne/blob/main/scripts/install.sh) gives an error:
```bash
═══════════════════════════════════════════════════════════
  Mnemosyne Installer for Hermes Agent
═══════════════════════════════════════════════════════════

📁 Hermes home: /Users/0rangefox/.hermes
📁 Hermes venv: /Users/0rangefox/.hermes/hermes-agent
📁 Mnemosyne source: /Users
📁 Plugin target: /Users/0rangefox/.hermes/hermes-agent/plugins/memory/mnemosyne

✓ Hermes found
🔗 Creating symlink...

📦 Installing dependencies into Hermes venv...
⚠️  Hermes venv not found, installing to system Python...
bash: line 101: pip: command not found
```

But I actually checked and there's venv inside of `/Users/0rangefox/.hermes/hermes-agent` folder, and to be able to install this I was needed to execute the following commands in this order:
```bash
cd ~/.hermes/hermes-agent/venv
uv pip install fastembed
uv pip install mnemosyne-memory
./bin/python -m mnemosyne.install
hermes memory setup
hermes tools disable memory
```

Or I could do this way too, which I think is better way to do:
```bash
cd ~/.hermes/hermes-agent/venv
uv pip install fastembed
uv pip install mnemosyne-hermes
./bin/python -m mnemosyne.install
hermes memory setup
hermes tools disable memory
```

Just a side note, for upgrade, it's necessary to do the following commands:
```bash
cd ~/.hermes/hermes-agent/venv
uv pip install --upgrade mnemosyne-hermes && hermes gateway restart
```

AxDSan:
@0rangeFox appreciate you catching the dead URL. The one-line fix is correct and I'm merging it.

But your comment is the real value here. You hit the pain points:

1. Three competing install methods. README says pip install. Integrations page says run a script. The hermes-integration.md doc says something else. A new user lands and has no idea which path to take. That's a conversion killer.

2. The install script breaks on MacOS. It fails to find the venv, falls back to system Python, then crashes because pip isn't on the PATH. Every MacOS + Hermes user hits this.

3. The upgrade path is undocumented. You had to reverse-engineer it yourself.

If you're up for it, I'd love your help on making this right. The goal is one install method, tested on MacOS + Linux, with the other docs deleted to prevent confusion. I just opened issue #204 tracking this.

Would you be interested in taking a pass at the install script fix? Happy to point you at the relevant parts.

Merging your PR now. Thanks for the contribution.

AxDSan:
Correction: the companion issue is #208 (not #204).

PR #206, opened 2026-05-30
feat(embeddings): Make MNEMOSYNE_EMBEDDING_API_URL independent of OPENROUTER_BASE_URL

mia-fourier:
feat(embeddings): Make MNEMOSYNE_EMBEDDING_API_URL independent of OPENROUTER_BASE_URL

Remove fallback to OPENROUTER_BASE_URL and OPENROUTER_API_KEY from embedding config.
Embedding models may use local llama.cpp, OpenAI, Anthropic, or any other provider —
not just wherever the general OpenRouter endpoint points.

Also fixes a bug in _embed_api() where base_url was read from env but then the stale
_OPENAI_BASE_URL was used for the actual URL construction.

### DIFF
--- a/mnemosyne/core/embeddings.py
+++ b/mnemosyne/core/embeddings.py
@@ -36,8 +36,10 @@ _FASTEMBED_AVAILABLE = np is not None and TextEmbedding is not None
 _FASTEMBED_CACHE_DIR = os.path.join(os.path.expanduser("~/.hermes"), "cache", "fastembed")
 
 # --- OpenAI-compatible API ---
-_OPENAI_API_KEY = os.environ.get("MNEMOSYNE_EMBEDDING_API_KEY", os.environ.get("OPENROUTER_API_KEY", os.environ.get("OPENAI_API_KEY", "")))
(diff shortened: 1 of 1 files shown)


AxDSan:
@mia-fourier good catch on the bug. The read-and-discard on MNEMOSYNE_EMBEDDING_API_URL is real. One thing before we merge: the PR removes the OPENROUTER_BASE_URL fallback. That's a breaking change for anyone who sets OPENROUTER_BASE_URL and expects embeddings to follow. Common pattern. The fix is simpler. On line 137, replace _OPENAI_BASE_URL with the local base_url variable you already read on line 132. That way your fix works AND the fallback chain is preserved. Two character change. Want to push that update? If not I can apply it and merge.

AxDSan:
Merged. @mia-fourier you already had the base_url fix in the latest commit, so my requested change was already done. Two birds with one commit.

Thanks for the clean fix. The _OPENAI_BASE_URL bug was real and the env independence makes embedding setup less fragile.

PR #216, opened 2026-06-01
fix: improve local multilingual recall

### LINKED ISSUE #215: [BUG] Local recall tokenization drops non-ASCII Latin characters, hurting multilingual retrieval
## Summary

Local recall currently tokenizes lexical recall gates with an ASCII-oriented pattern.  
As a result, words containing non-ASCII Latin characters such as German umlauts or `ß` can be split or partially dropped before recall scoring.

This hurts recall quality for multilingual or non-English notes, even when dense multilingual embeddings are used.

## Impact

Queries and stored facts containing words such as:

- `Stoßlüften`
- `Bürgeramt`
- `Primärquellen`
- `Mensa-Plan`

may not match reliably because the lexical recall layer does not preserve these words as complete tokens.

This is especially visible in local/offline setups using multilingual SentenceTransformers or FastEmbed models.

## Expected behavior

Recall tokenization should preserve Unicode Latin-script words with diacritics as complete tokens.

For example:
Stoßlüften im Bürgeramt: Primärquellen prüfen
should produce tokens including:

text
stoßlüften
bürgeramt
primärquellen
## Actual behavior

The current tokenization can split or degrade these terms into ASCII fragments, reducing lexical recall quality before dense embedding similarity can fully compensate.

## Reproduction scenario

1. Store facts containing non-English Latin text with diacritics.
2. Query using the same natural-language terms.
3. Compare recall results against equivalent ASCII-only terms.
4. Observe weaker or missing matches for words containing characters such as `ß`, `ü`, `ä`, `ö`, or accented Latin characters.

## Proposed fix

Use Unicode-aware tokenization for recall lexical gates instead of ASCII-only matching.

A suitable pattern should preserve Unicode alphanumeric word characters while avoiding punctuation and underscores.

This improves recall for German and other Latin-script languages without changing the embedding model itself.

## Additional improvement

It would also be useful to document known local multilingual embedding model dimensions, for example:

- `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
- `sentence-transformers/paraphrase-multilingual-mpnet-base-v2`

This makes local multilingual configuration easier and avoids manual dimension lookup.

## Notes

I have a local patch prepared that:

- updates recall tokenization to be Unicode-aware
- adds regression coverage for multilingual recall tokens
- documents local multilingual embedding options
- adds known dimensions for common multilingual SentenceTransformers models

I can open a PR if this direction looks acceptable.

## Environment

- Mnemosyne: current `main` branch
- Python: 3.13
- OS: Linux ARM64
- Setup: local SQLite memory store with local FastEmbed/SentenceTransformers embeddings
- Language tested: German text containing Unicode Latin characters such as `ä`, `ö`, `ü`, and `ß`

dplush:
## Summary

- Makes recall tokenization Unicode-aware for Latin-script languages
- Preserves words with diacritics such as `Stoßlüften`, `Bürgeramt`, and `Primärquellen`
- Adds regression coverage for multilingual local recall
- Documents local multilingual embedding model options

Fixes #215

## Test Plan

- `pytest tests/test_multilingual_local_recall.py`
- Full test suite checked locally; one unrelated pre-existing baseline failure was observed on `main`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [3.4.0] — 2026-06-01
+
+### Added
+
+- **Known dimensions for local SentenceTransformers multilingual models.**
+  `paraphrase-multilingual-MiniLM-L12-v2`, `all-MiniLM-L6-v2`, and
+  `paraphrase-multilingual-mpnet-base-v2` are now listed for low-resource
+  local multilingual embedding setups.
+
+### Fixed
+
+- **Unicode recall tokenization for Latin-script languages.** Recall lexical
+  gates now keep diacritics inside tokens, so words like `Stoßlüften`,
+  `Bürgeramt`, and `Primärquellen` are no longer split into ASCII fragments.
+
 ## [3.3.0] — 2026-06-01
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 *Zero-dependency AI memory that works everywhere. SQLite-backed. Sub-millisecond.*
 
 [![Python](https://img.shields.io/badge/Python-3.9+-blue.svg)](https://python.org)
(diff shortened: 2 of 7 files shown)


AxDSan:
Triggering CI run

PR #221, opened 2026-06-01
fix: strip <think> tags from LLM consolidation output + upgrade default model

carvalab:
Two changes in this PR.

Strip closed think tags from LLM output. Models like Qwen and MiniCPM emit thinking/reasoning blocks. Added stripping in _try_host_llm(), _clean_output(), and BeamMemory.consolidate_to_episodic(). Only closed tags are stripped. Unclosed tags are left alone because there is no way to tell where thinking ends and the response begins.

Switch the default local model from TinyLlama-1.1B to MiniCPM5-1B ([Artificial Analysis Intelligence Index](https://artificialanalysis.ai/articles/minicpm5-1b-the-leading-1b-open-weights-model), top score for sub-4B open models at 17.9). Same Q4_K_M quantization, similar size (~656MB vs ~600MB), Apache 2.0 license, 128K context window.

Updated docs (configuration.md, hermes-llm-integration.md, architecture.md) and code comments to reference MiniCPM5-1B instead of TinyLlama.

8 tests for think tag stripping: closed tags stripped, unclosed tags preserved, multiple blocks, passthrough, empty result, host LLM integration.

This fix the #222 issue.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -62,7 +62,7 @@ The `sleep()` function moves stale working memories into episodic memory:
 
 1. Fetches working memories past TTL or below importance threshold
 2. Groups them by source
-3. Attempts LLM summarization (local TinyLlama, remote OpenAI-compatible, or AAAK fallback)
+3. Attempts LLM summarization (local MiniCPM5-1B, remote OpenAI-compatible, or AAAK fallback)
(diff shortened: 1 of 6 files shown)


AxDSan:
Triggering CI run

PR #223, opened 2026-06-01
feat(core): add LLM model fallback chain for remote summarization

scubamount:
## Summary

`_call_remote_llm()` currently picks **one** model from `MNEMOSYNE_LLM_MODEL` against **one** URL from `MNEMOSYNE_LLM_BASE_URL`. If that combination 404s (e.g. omlx evicts the pinned model after its idle window), the call fails, the remote tier returns `None`, and the existing fallback chain drops straight to local GGUF / `None`. Memory consolidation silently degrades to AAAK encoding.

This adds a model-level fallback chain **within** the remote tier.

## New env vars (all optional, empty disables the chain)

| Var | Purpose | Default |
|---|---|---|
| `MNEMOSYNE_LLM_FALLBACK_MODELS` | Comma-separated model names tried in order | (disabled) |
| `MNEMOSYNE_LLM_FALLBACK_BASE_URL` | Endpoint for fallbacks | inherits `MNEMOSYNE_LLM_BASE_URL` |
| `MNEMOSYNE_LLM_FALLBACK_API_KEY` | Auth for fallbacks | inherits `MNEMOSYNE_LLM_API_KEY` |

## Retry policy

- **Retry**: 404, 400, 5xx, connection errors, timeouts. The other model might exist on the same host.
- **Don't retry**: 401, 403, 429. Bad credentials and rate limits won't be fixed by swapping model names — propagate so the user knows.
- **Don't retry** on a successful 200 with empty body — the endpoint is healthy, the prompt may just be wrong.

## Example

```bash
export MNEMOSYNE_LLM_BASE_URL=http://localhost:8080/v1
export MNEMOSYNE_LLM_MODEL=Qwen3.6-35B-A3B-PARO
export MNEMOSYNE_LLM_FALLBACK_MODELS=Qwen3.6-27B-MXFP4-MTP,Qwen2.5-32B-Instruct
```

## Implementation

- New private helper `_call_remote_llm_with_model(prompt, model, temperature, *, base_url, api_key, timeout)` returns a 3-tuple `(text, status, exc)` so callers can decide whether to retry.
- New `_is_retryable_status(status_code)` centralizes the status-code policy.
- Public `_call_remote_llm()` becomes a thin orchestrator: build candidate list (primary + deduped fallbacks), iterate, return first success.
- No public API breakage. The 5-tier fallback chain (host, remote, llama-cpp, ctransformers, None) is unchanged. The remote tier is now multi-model.

## Tests

12 new cases in `TestRemoteLLMFallback`:

- Status-code policy table-driven test (9 cases)
- Primary success skips fallback
- 404 / 5xx / connection error trigger fallback
- 401 / 429 do NOT trigger fallback
- Multiple fallbacks iterate in order
- Returns `None` when every model fails
- Empty fallback list → only primary tried
- Primary deduplicated from `LLM_FALLBACK_MODELS` (no self-retry)
- Fallback base URL/key override takes effect
- Fallback inherits primary base URL/key when not overridden

Test count: 30 / 30 pass in `tests/test_local_llm.py`. Full suite: 1390 pass, 44 pre-existing failures (all in `test_c4_recall_diagnostics`, `test_recall_precision_regressions`, `test_temporal_recall`, etc. — recall/beam scoring, none touching this change; verified by stashing the diff and re-running).

## Files changed

- `mnemosyne/core/local_llm.py` (+108 / -25)
- `tests/test_local_llm.py` (+186 / -0)

## Notes for review

- One pre-existing test (`TestRemoteLLM::test_call_remote_llm_with_mock_response`) had a brittle mock that didn't set `status_code` on the fake response. I added `status_code=200` to the mock — minimal, makes the test reflect real httpx behavior.
- Pre-existing LSP errors at lines 99/207/213/504 (`hf_hub_download` overload, `ctransformers` API drift, unbound `httpx`) are unrelated to this change and exist on `main`.

### DIFF
--- a/mnemosyne/core/local_llm.py
+++ b/mnemosyne/core/local_llm.py
@@ -37,6 +37,14 @@ LLM_BASE_URL = os.environ.get("MNEMOSYNE_LLM_BASE_URL", "").rstrip("/")
 LLM_API_KEY = os.environ.get("MNEMOSYNE_LLM_API_KEY", "")
 LLM_REMOTE_MODEL = os.environ.get("MNEMOSYNE_LLM_MODEL", "")
 
+# Retryable errors only: 404/400 (model-not-found), 5xx, connection. Not 401/403/429.
+LLM_FALLBACK_MODELS = [
+    m.strip() for m in os.environ.get("MNEMOSYNE_LLM_FALLBACK_MODELS", "").split(",")
+    if m.strip()
+]
+LLM_FALLBACK_BASE_URL = os.environ.get("MNEMOSYNE_LLM_FALLBACK_BASE_URL", "").rstrip("/") or LLM_BASE_URL
+LLM_FALLBACK_API_KEY = os.environ.get("MNEMOSYNE_LLM_FALLBACK_API_KEY", "") or LLM_API_KEY
+
 # Host LLM adapter (Hermes or another agent). Disabled by default to preserve
 # existing standalone behavior. When MNEMOSYNE_HOST_LLM_ENABLED=true and a
 # backend is registered via mnemosyne.core.llm_backends.set_host_llm_backend(),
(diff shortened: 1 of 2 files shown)


AxDSan:
Triggering CI run

AxDSan:
Merged to main via commit 6c6ed98. Thanks @scubamount!

AxDSan:
Merged in 6c6ed98

PR #228, opened 2026-06-02
fix: add missing capture groups to multilingual named_months regex patterns

Whishp:
## Problem

`mnemosyne_remember` raises `IndexError: no such group` when the input contains Russian, Italian, or Spanish date strings **or** Russian/Italian negation sentences. This is triggered inside `extract_and_store_facts()` where the code unconditionally calls `m.group(1)` on pattern matches that have zero capture groups.

### named_months patterns (original fix)

| Language | Old pattern | Groups | Problem |
|----------|-----------|--------|---------|
| Russian (ru) | `r'(?:...)...` | **0** | `m.group(1)` → IndexError |
| Italian (it) | `r'(?:...)...` | **0** | `m.group(1)` → IndexError |
| Spanish (es) | `r'(\\d{1,2})...de...(enero|...)...(?:de(\\d{4}))?'` | **3** | `m.group(1)` returns only day number instead of full date |

English and German named_months patterns were correct (1 capture group).

### negation patterns (added in v2)

| Language | Old pattern | Groups | Problem |
|----------|-----------|--------|---------|
| Russian (ru) | `r'(?:...)...` | **0** | `m.group(1)` → IndexError |
| Italian (it) | `r'(?:...)...` | **0** | `m.group(1)` → IndexError |

English, German, and Spanish negation patterns already had >=1 capture group.

### named_months

1. **Russian `named_months`**: wrapped `(?:...)` → `((?:...))` — now has 1 capture group
2. **Italian `named_months`**: same fix
3. **Spanish `named_months`**: consolidated all 3 groups into 1 outer group

### negation

4. **Russian `negation`**: wrapped `(?:...)` → `((?:...))` — now has 1 capture group
5. **Italian `negation`**: same fix

### Defense-in-depth

6. Wrapped all three `extract_and_store_facts()` calls (`remember()` dedup path, `remember()` new-row path, `remember_batch()`) in `try/except` so regex extraction failures never block memory storage

## Impact

Any Russian, Italian, or Spanish text containing date references or Russian/Italian negation phrases causes `remember()` to crash. The `try/except` guard prevents similar undiscovered pattern issues from affecting memory writes.

## Full verification

All 35 patterns (5 languages × 7 types: named_months, negation, decision, entity, sequence, instruction, preference) were programmatically verified:
- All have >=1 capture group
- entity patterns have exactly 2 (used with m.group(1) + m.group(2))
- Runtime tested with real text samples for all languages
- All compiled without regex syntax errors

## Statistics

- Files changed: 1 (`mnemosyne/core/beam.py`)
- Additions: +18, Deletions: -6

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -3380,7 +3380,7 @@ class BeamMemory:
             'named_months': r'((?:Januar|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember|Jan|Feb|Mär|Apr|Mai|Jun|Jul|Aug|Sep|Okt|Nov|Dez)\s+\d{1,2}(?:\.)?\s*(?:\d{4})?)',
         },
         'ru': {
-            'negation': r'(?:Я(?: |\')?(?:никогда|не)|(?:никогда|не)\s+будет)\s+[^.,;!?\n]{6,120}',
+            'negation': r'((?:Я(?: |\')?(?:никогда|не)|(?:никогда|не)\s+будет)\s+[^.,;!?\n]{6,120})',
             'decision': r'(?:решил|решила|решили|выбрал|выбрала|выбрали|перешёл|перешла|перешли|переключился|переключилась|переключились|переехал|переехала|переехали|поменял|поменяла|поменяли)\s+([^.,;!?\n]{2,120})',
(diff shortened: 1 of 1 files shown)


AxDSan:
Merged! Thanks @Whishp for catching the capture group issue.

Note: the try/except blocks currently swallow regex failures silently for those languages. I'll add INFO-level logging as a follow-up so extraction failures are observable. Keeping the merge unblocked — your fix is more important than the logging polish.

PR #230, opened 2026-06-03
fix(beam): pass connection by keyword to VeracityConsolidator in recall

### LINKED ISSUE #229: Consolidated-facts recall passes connection positionally → spawns repr-named SQLite files and silently returns nothing
## Summary

In the `consolidated_facts` recall path ("Source 2") in `mnemosyne/core/beam.py`, the BeamMemory thread-local connection is passed **positionally** to `VeracityConsolidator`:

https://github.com/AxDSan/mnemosyne/blob/main/mnemosyne/core/beam.py#L6104

```python
consolidator = VeracityConsolidator(self.conn)
```

But the constructor signature is:

```python
class VeracityConsolidator:
    def __init__(self, db_path: Path = None, conn=None):
```

So `self.conn` (a `_BeamConnection`) binds to **`db_path`**, and `conn` stays `None`. That takes the `conn is None` branch, which runs:

```python
self.conn = sqlite3.connect(str(self.db_path), check_same_thread=False)
```

`str(<_BeamConnection object>)` evaluates to `"<mnemosyne.core.beam._BeamConnection object at 0x...>"`, which becomes a literal filename.

## Impact

1. **File litter:** every consolidated-facts recall creates a fresh empty SQLite DB in the CWD named `<mnemosyne.core.beam._BeamConnection object at 0x...>`. In practice this accumulates several per day under normal use.
2. **Silent recall failure (the important one):** that throwaway DB is empty, so `get_consolidated_facts()` always queries a blank database. Source 2 returns nothing — the consolidated-facts recall path is effectively dead.

## Root cause

Keyword-vs-positional mismatch. The correct call site one screen up gets it right:

https://github.com/AxDSan/mnemosyne/blob/main/mnemosyne/core/beam.py#L1993

```python
self.veracity_consolidator = VeracityConsolidator(conn=self.conn, db_path=self.db_path)
```

## Suggested fix

```python
consolidator = VeracityConsolidator(conn=self.conn, db_path=self.db_path)
```

PR incoming.

Milgauss:
Fixes #229.

## Problem

In the `consolidated_facts` recall path ("Source 2"), the connection is passed positionally:

```python
consolidator = VeracityConsolidator(self.conn)   # beam.py:6104
```

`VeracityConsolidator.__init__(self, db_path=None, conn=None)` — so `self.conn` binds to `db_path`, `conn` stays `None`, and the constructor falls into the `conn is None` branch:

```python
self.conn = sqlite3.connect(str(self.db_path), ...)
```

`str(<_BeamConnection object>)` becomes a filename, so each recall:

1. creates an empty SQLite file named `<mnemosyne.core.beam._BeamConnection object at 0x...>` in the CWD, and
2. queries that empty throwaway DB instead of the real one, so `get_consolidated_facts()` always returns nothing — Source 2 is silently dead.

## Fix

One line, matching the already-correct call site at `beam.py:1993`:

```python
consolidator = VeracityConsolidator(conn=self.conn, db_path=self.db_path)
```

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -6101,7 +6101,7 @@ class BeamMemory:
         try:
             # Import lazily so the consolidator module is only loaded on-demand
             from mnemosyne.core.veracity_consolidation import VeracityConsolidator
-            consolidator = VeracityConsolidator(self.conn)
+            consolidator = VeracityConsolidator(conn=self.conn, db_path=self.db_path)
             seen_subjects: set = set()
             for word in query_lower.split()[:6]:
                 if len(word) < 3:


AxDSan:
Merged!  One-line fix for a nasty silent bug. Appreciate the clean PR.

If you've got bandwidth, a regression test for the VeracityConsolidator construction path would be a great follow-up. The correct pattern at beam.py:1993 + the fixed call at beam.py:6104 should both be covered.

Thanks again @Milgauss!

PR #239, opened 2026-06-05
docs: propose tags and scope unification RFC

achrllrogia45:
## Description

What does this change do and why? Keep it focused.

## Related Issue

https://github.com/AxDSan/mnemosyne/issues/205

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [x] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

## Checklist

- [ ] `CHANGELOG.md` updated with a brief entry

### DIFF
--- a/docs/rfc/0001-tags-and-scope-unification.md
+++ b/docs/rfc/0001-tags-and-scope-unification.md
@@ -1,0 +1,649 @@
+# RFC 0001: Tags and Scope Unification for Recall
+
+**Status:** Draft
+**Author:** achrllrogia45
+**Related issue:** https://github.com/AxDSan/mnemosyne/issues/205
+**Target version:** TBD
+
+---
+
+## Summary
+
+`scope` currently exists on `remember()` but not on `recall()`. This creates an API symmetry gap: users can write scoped memories, but cannot explicitly ask recall to include or exclude a scope.
+
+This RFC proposes a small unification:
+
+1. Keep `scope` for backward compatibility.
+2. Add first-class `tags` to memories.
+3. Treat `scope` as one reserved tag namespace: `scope:<value>`.
+4. Add tag filtering to `recall()`.
+5. Preserve current default recall behavior unless a caller asks for filters.
+
+In short:
+
+```python
+remember("Fix login bug", scope="session", tags=["project:app", "area:auth"])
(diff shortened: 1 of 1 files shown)


AxDSan:
@achrllrogia45 this is excellent. Read it cover to cover.

The tags-over-scope model is smarter than what I originally proposed. I was thinking "extend scope to do more things" and you correctly identified that tags are the right primitive with scope as one reserved namespace. This avoids the v1/v2 redesign trap.

A few thoughts:

**The core design I'm adopting.**
Tags as the general mechanism, scope as a reserved tag namespace, scope= as convenience syntax. This is the right call. The tag format rules (namespace:value, dedup, soft limits) are well specified and I can implement from the sketch as-is.

**Schema question.**
You have denormalized JSON in working_memory + episodic_memory columns as v1. I'd add a memory_tags join table from day one. JSON is easier to write but the recall hot path needs an index for tag filtering. A join table with (memory_id, tag, table_name) gets us proper indexed lookups without the migration headache later. The schema cost is one extra table. Worth it.

**Backfill approach.**
The deterministic scope backfill is smart. Idempotent, no data loss. The _parse_tags tolerant reader is exactly right. recall must never crash on bad data.

**Candidate pool widening.**
The max(top_k * 5, 100) overfetch when tags are present is the right call for v1. SQL pushdown later.

**One gap.**
The RFC doesn't address how session_id interacts with scope filtering. If someone does recall("bug", tags=["scope:global"]), does the current session_id filter still apply? The answer should be no. scope:global explicitly means "all sessions." But if someone does recall("bug", tags=["area:auth"]) without a scope tag, session_id should still scope it. Worth adding a sentence on this.

Overall this is ready to adopt as the design doc. Going to merge it as-is and start the implementation sketch. Happy for you to take the lead on the implementation PR if you want it.

AxDSan:
@achrllrogia45 this is merged to main (d14f3e2). Tags-over-scope model adopted as the design direction.

Thank you for the deep RFC work. This is now the reference doc for scope filtering. Like I said in the review, happy for you to take the lead on the implementation PR if you want to keep rolling.

achrllrogia45:
great, thanks and happy to help for the lead at the implementation PR for these.

PR #246, opened 2026-06-06
feat(triples): complete the temporal lifecycle — explicit valid_until, triple_end, and multi-value adds

Milgauss:
## Problem

`TripleStore`'s schema already has `valid_until`, but the public API and the Hermes tool surface can't actually use the temporal model:

1. `add()` always auto-closes the prior open `(subject, predicate)` and exposes no `valid_until`. So a predicate that is legitimately **multi-valued** silently clobbers itself — e.g. `('user','speaks','English')` then `('user','speaks','Spanish')` leaves only the second valid — and a fact with a **known end date** can't be recorded.
2. There's **no way to end a fact without replacing it.** When a relationship simply ends (e.g. `('alice','works_at','AcmeCorp')` once she leaves), you must invent a replacement triple just to close the old one.
3. **No historical queries.** `query()` accepts `as_of` internally but it isn't exposed; subject matching is also case-sensitive.

## What this PR adds

Fully back-compatible — every new param is optional and defaults to today's behavior; no schema migration (columns already exist):

- `TripleStore.add(..., valid_until=None, supersede=True)` (+ module `add_triple`):
  - `valid_until` — record an explicit expiry.
  - `supersede=False` — insert **without** closing prior same-`(subject,predicate)` triples, enabling multi-valued facts. `True` keeps current auto-close behavior.
  - also passes through `source` / `confidence` (already columns).
- `TripleStore.end(subject, predicate, object=None, valid_until=None) -> int` (+ module `end_triple`): close open triples for a `(subject, predicate)` — or one exact triple when `object` is given — by setting `valid_until` (default today). Returns rows closed.
- `query()` subject match is now case-insensitive (`COLLATE NOCASE`).
- Hermes provider: `mnemosyne_triple_add` gains optional `valid_until`/`source`/`confidence`/`supersede`; new `mnemosyne_triple_end` tool (mirrors `mnemosyne_invalidate`); `mnemosyne_triple_query` gains `as_of`.
- `tests/test_triples_lifecycle.py`.

## Examples

\`\`\`python
# multi-valued fact — both stay valid
add_triple("user", "speaks", "English", supersede=False)
add_triple("user", "speaks", "Spanish", supersede=False)

# end a fact with no replacement
end_triple("alice", "works_at", "AcmeCorp")          # valid_until = today

# record a known end date up front
add_triple("project_x", "status", "active", valid_until="2026-12-31")

# historical query
triple_query(subject="alice", as_of="2025-01-01")
\`\`\`

## Back-compat

`supersede` defaults to `True` (== current auto-close); all other new params default to today's values. Existing callers and stored data are unaffected. `tests/test_triples_lifecycle.py` passes.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -504,7 +504,10 @@ TRIPLE_ADD_SCHEMA = {
     "name": "mnemosyne_triple_add",
     "description": (
         "Add a temporal fact triple (subject, predicate, object) to the knowledge graph. "
-        "Example: ('user', 'prefers', 'neovim'). Use for structured relationships."
+        "Example: ('user', 'prefers', 'neovim'). Use for structured relationships. "
+        "By default a new triple supersedes any prior fact with the same subject+predicate; "
+        "set supersede=false for multi-valued facts that should coexist "
+        "(e.g. ('user','speaks','English') and ('user','speaks','Spanish'))."
     ),
     "parameters": {
         "type": "object",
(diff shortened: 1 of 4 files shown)


AxDSan (approved):
## Review: PR #246 — triples temporal lifecycle

Biggest PR in the batch (+196/-27). Adds the full temporal model to TripleStore that was already baked into the schema but never exposed through the API:

**What it adds (all back-compat, all optional):**
- `TripleStore.add(..., valid_until=None, supersede=True)` — `supersede=False` enables multi-valued facts (speaks English + Spanish)
- `TripleStore.end(subject, predicate, object=None, valid_until=None)` — close open triples without replacing them
- `query()` is now case-insensitive (`COLLATE NOCASE`)
- New Hermes tools: `mnemosyne_triple_end`, `mnemosyne_triple_query` gains `as_of`

**Code:** 3 files in core (`triples.py`), 1 in provider (`hermes_memory_provider/__init__.py`), new test file (66 lines). All mergeable, no reviews.

**Risk assessment:**
- Schema columns already exist — no migration risk
- All new params default to current behavior
- Test file covers the lifecycle: add-with-end, multi-valued, historical query, end-without-replacement
- The case-insensitive query is a behavioral change but a strict improvement (nobody wants case-sensitive subject matching)

**One concern:** `supersede=True` as default means this doesn't break anything, but I need to check if any internal code path calls `add()` and expects the auto-close behavior — they'd still get it. Confirmed: `supersede` defaults to `True` == current auto-close. Safe.

**Verdict:** LGTM. Well-designed API, solid tests, zero back-compat risk. Milgauss continues to deliver clean, well-specified PRs.

PR #247, opened 2026-06-07
perf(recall): bound entity-aware recall fuzzy-match cost

Milgauss:
## Summary

`_find_memories_by_entity` (in `BeamMemory.recall`) passes the **entire recall query** into `find_similar_entities`, which Levenshtein-compares it against *every* known `mentions` entity — **O(len(query) × N)** pure-Python. On a long query against a large mentions store this pins a CPU core, and because the loop holds the GIL it stalls the whole event loop.

Observed in a live asyncio deployment: a multi-KB query against a ~2.7k-entity mentions store hung the process for minutes; a single recall call measured **~38 s** (≈14 ms × 2,711).

Two complementary fixes:

- **`entities.py` `find_similar_entities`**: refuse inputs longer than 64 chars (entity names are short; an over-long string is a sentence, not an entity). A hard backstop so no caller can re-trigger the blow-up.
- **`beam.py` `_find_memories_by_entity`**: match entities *extracted* from the query (`extract_entities_regex`) instead of the raw query string, falling back to the raw string only when it is already short enough to be an entity name. Removes the `len(query)` factor and is semantically correct — entity-aware recall should match entities, not whole sentences.

## Test plan

- A 3.2 KB query × 2,711 entities: **~38 s → ~0.001 ms**.
- Short real entities still match (threshold-equivalent behavior unchanged).
- `extract_entities_regex` still pulls `Andy`/`Jessi`/`New York` out of a sentence, so real entity recall is preserved.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1340,19 +1340,36 @@ def _find_memories_by_entity(beam: "BeamMemory", entity_name: str, threshold: fl
     because of auto-invalidation on (subject, predicate).
     """
     try:
-        from mnemosyne.core.entities import find_similar_entities
+        from mnemosyne.core.entities import find_similar_entities, extract_entities_regex
 
         # Get all known entities (uses BeamMemory's cached AnnotationStore)
         known_entities = beam.annotations.get_distinct_values("mentions")
         if not known_entities:
             return []
 
-        # Find similar entities
-        matches = find_similar_entities(entity_name, known_entities, threshold=threshold)
+        # Bound the fuzzy-match cost. entity_name is the raw recall query here, and
+        # Levenshtein-matching the WHOLE query against every known entity is
+        # O(len(query) * N) -- pathological for a long query against a large
+        # mentions store (a multi-KB query vs a few thousand entities can pin a CPU
+        # core for minutes, and because the loop holds the GIL it stalls the whole
+        # event loop). Match against entities EXTRACTED from the query instead; only
+        # fall back to the raw string when it is already short enough to be an
+        # entity name.
+        candidates = extract_entities_regex(entity_name)
+        if not candidates and len(entity_name) <= 64:
+            candidates = [entity_name]
+
+        # Find similar entities for each candidate (dedup by matched name)
+        matched_names: Set[str] = set()
+        for cand in candidates:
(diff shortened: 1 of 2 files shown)


Milgauss:
**CI note — the failing check is environmental, not from this change.**

This diff is limited to `mnemosyne/core/entities.py` and `mnemosyne/core/beam.py` (the `find_similar_entities` length guard + `_find_memories_by_entity` extract-then-match). It doesn't touch embeddings, consolidation, or recall scoring tables.

`test (3.11)` ran for ~6h and hit the job timeout; every failure in it is the embedding-dependent cluster (`test_ab_toggles`, `test_beam` episodic/sleep) where `fastembed` couldn't load `BAAI/bge-small-en-v1.5`, and individual tests were taking ~78s each (the whole suite runs in ~2.5 min on the other versions). `test (3.10)`, `test (3.12)`, and `test (3.13)` all passed the same code. This is a degraded/slow runner + HF model-download issue, not a code fault — a re-run should clear it (I don't have re-run permission here). Thanks!

AxDSan (approved):
## Review: PR #247 — bound entity-aware recall cost

Fixes a real perf issue: `_find_memories_by_entity` was passing the entire recall query (potentially multi-KB) into `find_similar_entities`, which Levenshtein-compares against every known mentions entity. Observed: 38s for a 3.2KB query against 2.7k entities.

**Code:**
1. `entities.py`: refuse inputs >64 chars in `find_similar_entities` (hard backstop)
2. `beam.py`: match entities *extracted* from the query instead of raw query string, fall back to raw string only when it's already short enough

**Tests:** No new tests for the 64-char guard (minor — the guard is self-evident). Existing tests pass.

**Risk:** Two edge cases to consider:
- A query that's all one short entity name (e.g. "ComfyUI") still matches the same way via the fallback path — behavior is equivalent for short inputs.
- A query with zero extracted entities and a length >64 chars — was previously doing the expensive scan, now does nothing (drops content). Is this correct? Yes — if the query has no extractable entities, entity-aware recall has nothing to work with. The fallback `recall()` paths cover standard semantic search.
- A query with zero extracted entities but <64 chars still falls back to raw string match. Correct.

**Verdict:** LGTM. Well-motivated, correct fix.

PR #232, opened 2026-06-03
Filter pronoun/demonstrative/possessive subjects in regex fact extraction

Milgauss:
Fixes #231.

`EpisodicGraph.extract_facts()` is regex-based and captured any capitalized `"X is/has/uses/works at Y"` as a triple, gated only by `len > 2`. On conversational text this yields triples with pronoun/demonstrative subjects (`This`, `It`, `There`) or possessive-led phrases (`My report`). All such triples share the same `(subject, predicate)`, so `VeracityConsolidator` flags every differing object as a `contradiction` — flooding the `conflicts` table with unresolved noise and burying genuine contradictions.

**Change:** adds a `_LOW_QUALITY_SUBJECT_LEADERS` stoplist + `_is_low_quality_subject()` helper, applied to all four `extract_facts()` patterns (`is` / `has` / `uses` / `works_at`). Degenerate-subject triples are dropped at the source; named-entity facts are preserved.

**Verified:**
- Rejected: `This`, `It`, `There`, `My report`, `Her team`, `They`
- Kept: `Alice`, `Bob`, `Acme Corp`, `The build server`
- `"Alice is a developer"` / `"Bob works at Acme"` still extract; narration-style input yields no facts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #248, opened 2026-06-07
fix(extraction): article-led subjects, object-side guard, truncation fix

Milgauss:
## Summary

Follow-up to #232 (pronoun/demonstrative/possessive subject guard in regex fact extraction). An audit of the regex fact tier on conversational/narrative traffic found it still ~98% noise — #232 caught pronoun subjects, but three large junk classes remained:

1. **Article-led subjects** ("The silence is different", "A frame has X"). A named entity never opens with an article, so these are the same transient-state noise as the pronoun class — added `"the"`/`"a"`/`"an"` to `_LOW_QUALITY_SUBJECT_LEADERS`.
2. **Unguarded object side.** New `_is_low_quality_object()` drops a lone object token unless it is multi-word or a proper noun (lone capitalized token kept: "ComfyUI", "Rust"); applied to all four extraction patterns.
3. **Truncation bug.** The optional-article alternation `(?:a|an|the)?\s*` matched a bare "a" *inside the next word* ("is already" → article "a" + object "lready"). Changed to `(?:(?:a|an|the)\s+)?` (article must be a whole word) on the is/has/uses patterns.

Named-entity facts still pass; narrative prose is dropped at the source.

## Test plan

- "The silence is different" / "A frame has X" → no longer extracted (article subjects dropped).
- "Andy has lready" style truncation artifacts → object-side guard + whole-word article alternation drop them.
- "ComfyUI" / "Rust" as lone objects → still kept (proper-noun rule).
- Multi-word proper-noun facts ("Andy uses ComfyUI") → still extracted.

### DIFF
--- a/mnemosyne/core/episodic_graph.py
+++ b/mnemosyne/core/episodic_graph.py
@@ -54,21 +54,37 @@ def _is_low_quality_subject(subject: str) -> bool:
     return first in _LOW_QUALITY_SUBJECT_LEADERS
 
 
+# Lone lowercase tokens that carry no nameable value — transient-state
+# adjectives and fillers. A common-noun object ("developer", "engineer") is
+# legitimate, so we cannot reject all lone lowercase words; we reject only
+# this curated state/filler set plus -ly adverbs. Keyed lowercased.
+_LOW_QUALITY_OBJECT_WORDS = frozenset({
+    "different", "same", "similar", "fine", "okay", "ok", "good", "bad",
+    "here", "there", "ready", "already", "done", "gone", "back",
+    "nothing", "something", "anything", "everything",
+})
+
+
 def _is_low_quality_object(obj: str) -> bool:
-    """True if `obj` is a lone, value-free object token — an adverb/adjective
-    fragment ("different", "definitely"), an -ly adverb, or a truncation artifact
-    ("lready", "pparently"). A real object is a multi-word phrase or a proper
-    noun (kept: capitalized tokens like "ComfyUI", "Rust"). Mirrors
-    _is_low_quality_subject for the object side, which the four extraction
-    patterns previously left unguarded."""
+    """True if `obj` is a lone, value-free object token — an -ly adverb
+    ("definitely", "apparently") or a transient-state/filler word
+    ("different", "already"); see _LOW_QUALITY_OBJECT_WORDS. A real object is
+    a multi-word phrase, a proper noun ("ComfyUI", "Rust"), or a lone common
+    noun ("developer"). Mirrors _is_low_quality_subject for the object side,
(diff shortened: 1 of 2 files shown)


Milgauss:
**CI note — the failing check is environmental, not from this change.**

This diff is limited to `mnemosyne/core/episodic_graph.py` (article-led subject leaders, the object-side guard, and the whole-word article truncation fix in `extract_facts`). The failing tests are unrelated to extraction.

`test (3.11)` ran ~3h with every failure in the embedding-dependent cluster (`test_ab_toggles`, `test_beam` episodic/sleep) where `fastembed` couldn't load `BAAI/bge-small-en-v1.5`, individual tests taking ~78s each (vs ~2.5 min for the whole suite on the other versions). `test (3.10)`, `test (3.12)`, and `test (3.13)` all passed the same code. That's a degraded/slow runner + HF model-download issue, not a code fault — a re-run should clear it (I don't have re-run permission here). Thanks!

AxDSan (approved):
## Review: PR #248 — extraction fixes (article subjects, object guard, truncation)

Three fixes for the regex fact tier:

1. **Article-led subjects** ("The silence...", "A frame...") — added `"the"`/`"a"`/`"an"` to `_LOW_QUALITY_SUBJECT_LEADERS`. Named entities never open with an article, correct.
2. **Object-side guard** — new `_is_low_quality_object()` drops lone non-proper-noun object tokens. Same logic as the prefetch capitalization guard in #249.
3. **Truncation bug** — the optional-article alternation `(?:a|an|the)?\s*` was matching bare "a" *inside* the next word. Changed to `(?:(?:a|an|the)\s+)?`. Correct fix.

**Code:** All three are in `mnemosyne/core/episodic_graph.py`. +48/-9. New test file (+68 lines) tests all three fix classes.

**Verdict:** LGTM. Each fix is well-scoped, independently verifiable, and the test file covers before/after behavior. No new dependencies or risky refactors.

Real talk: this is Milgauss's third PR in this area (#232, then #249/#247, now this). The fact extraction tier has been a recurring noise source, and they're systematically working through it. Good engineering discipline.

ekinnee:
Heads-up: this merged fix is not present on the current `main` branch.

I verified that `mnemosyne/core/episodic_graph.py` on current `main` still lacks `_is_low_quality_object()` and still uses the pre-fix lazy object patterns. The file is byte-identical to our `6c3e0d8` container snapshot, which is producing the same single-token regex facts this PR was intended to prevent.

GitHub's compare endpoint reports this PR's merge commit (`f9cc53b`) and current `main` as diverged, with `main` behind that merge. Could this fix be restored/cherry-picked onto current `main`?

@AxDSan

PR #249, opened 2026-06-07
fix(prefetch): drop lone non-proper-noun tokens regardless of length

Milgauss:
## Summary

Follow-up to #243 (drop bare single-token extraction fragments from prefetch). The length+stopword gate let lone tokens ≥9 chars that were not stopwords through ("basically", "different", "physical"), and because the fact source merges first, even one importance-0 fragment could **lead** the injected prefetch block.

Replace the length+stopword test in `_is_low_quality_prefetch` with a **capitalization rule**: a lone token is kept only if it is capitalized (a proper noun like "ComfyUI"/"Rust"), otherwise dropped — regardless of length. Multi-word phrases still always pass. `_PREFETCH_MIN_FRAGMENT_CHARS` / `_PREFETCH_FRAGMENT_STOPWORDS` are retained as documented intent; the capitalization rule subsumes them.

## Test plan

- Lone "basically" / "different" (≥9 chars, non-stopword) → now dropped (previously leaked).
- Lone "ComfyUI" / "Rust" → kept (proper noun).
- Multi-word phrases → unaffected (always pass).

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -131,20 +131,26 @@ _PREFETCH_FRAGMENT_STOPWORDS = frozenset({
     "even", "very", "really", "again", "away", "off", "out", "up",
     "down", "over", "that", "this", "it", "so",
 })
-_PREFETCH_MIN_FRAGMENT_CHARS = 8   # lone tokens shorter than this are dropped
+_PREFETCH_MIN_FRAGMENT_CHARS = 12  # lone tokens shorter than this are dropped
 _PREFETCH_OVERFETCH = 16           # recall more, then filter junk and cap
 _PREFETCH_TOP_K = 8                # final injected count (unchanged behavior)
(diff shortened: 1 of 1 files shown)


AxDSan (approved):
## Review: PR #249 — prefetch capitalization guard

Follow-up to #243. The length+stopword gate let lone tokens ≥9 chars through ("basically", "different"). This replaces it with a capitalization rule: a lone token is kept only if capitalized (proper noun). Multi-word phrases unaffected.

**Code:** One expression change in `_is_low_quality_prefetch`. Clean.

**Tests:** Explicit test cases for the before/after behavior. Existing 1400+ tests pass.

**Risk:** Near zero. The capitalization rule is strictly stricter than the old length+stopword gate — it drops everything the old gate dropped plus capitalizes on the new cases. No false positives.

**Verdict:** LGTM. Straight improvement.

PR #250, opened 2026-06-07
fix(annotations): cascade-delete annotations when memories are removed

Milgauss:
## Summary

The `annotations` table (kinds: `mentions`, `occurred_on`, `has_source`) has no foreign key or trigger linking it to the memory tables, so a memory's annotations have to be deleted explicitly when the memory is. `forget_working()` does this correctly, but two other deletion paths do not — leaving annotations whose `memory_id` no longer exists (orphans) that no later write reclaims:

- **`_trim_working_memory()`** — the recurring TTL/size trim. This is the main accumulator, since it runs on an ongoing basis.
- **`import_from_dict(force=True)`** — force-overwrite of an existing `working_memory` or `episodic_memory` row before re-insert.

Over time the orphans inflate `SELECT DISTINCT value FROM annotations WHERE kind='mentions'`, which the entity-aware recall path scans.

## Changes

Apply the same cascade these two paths were missing, mirroring `forget_working()`:

- **Trim:** the trim selection is expressed once, and annotations for the matched `memory_id`s are deleted *before* the rows (while they still exist to resolve the subquery), wrapped in a transaction with rollback. Both statements share one predicate + params, so they operate on an identical row set.
- **Import force-overwrite:** `DELETE FROM annotations WHERE memory_id = ?` after the `working_memory` and `episodic_memory` deletes (a no-op when the row had none).

Scope is limited to **preventing new orphans**. Pre-existing orphans are left untouched and can be swept separately (e.g. `DELETE FROM annotations WHERE memory_id NOT IN (SELECT id FROM working_memory UNION SELECT id FROM episodic_memory)`).

## Test plan

- Unit-level SQL check of the trim cascade: an aged row is trimmed and its annotation is removed with it; a fresh row and a consolidated (trim-exempt) row, plus their annotations, are kept; no orphan remains.
- `forget_working()` behavior is unchanged (it already cascaded).

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -2827,9 +2827,13 @@ class BeamMemory:
         the experiment Arm B's "ADD-only" guarantee collapses at 24h.
         """
         cutoff = (datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS)).isoformat()
-        self.conn.execute("""
-            DELETE FROM working_memory
-            WHERE session_id = ?
+        # The trim selection, expressed once so the annotation cascade and the
+        # working_memory DELETE target an identical set. A memory's annotations
+        # must not outlive the memory: like forget_working(), the trim has to
+        # cascade, or the annotations table accumulates orphans (rows whose
+        # memory_id no longer exists) that no later write reclaims.
+        trim_where = """
+            session_id = ?
               AND consolidated_at IS NULL
               AND (
                 timestamp < ? OR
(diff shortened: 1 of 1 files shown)


Milgauss:
**CI note — the red checks are environmental, not from this change.**

The diff here is limited to `mnemosyne/core/beam.py`: the annotation cascade in `_trim_working_memory` and `import_from_dict(force=True)`, mirroring the existing `forget_working` cascade. It doesn't touch embeddings, consolidation, delta-sync, or the CLI.

Across two runs the failures are infrastructure flakiness, and each run failed on a *different* job:

- **Run 1** — only `test (3.12)` failed; every failure was `ValueError: Could not load model BAAI/bge-small-en-v1.5 from any source.` (fastembed couldn't fetch the model from HF on that runner). `3.10/3.11/3.13` passed.
- **Run 2** (after a no-op re-push) — `test (3.12)` now passes; instead `test (3.10)` ran ~6h and hit the job timeout, with individual tests taking ~78s each (the whole suite runs in ~2.5 min on the other versions). The errors/failures there span unrelated subsystems (`test_c25_deltasync_allowlist`, `test_cli_errors`, `test_entity_integration`) — the signature of a degraded/slow runner, not a code fault.

The change passes consistently on 3.11/3.12/3.13. Could you re-run the failed job when convenient? I don't have re-run permission on this repo. Thanks!

AxDSan (approved):
## Review: PR #250 — cascade-delete annotations

This is a clean bug fix. Three deletion paths in the codebase delete memories without cleaning up their annotations; these two (trim + import force-overwrite) were the remaining gaps. The fix mirrors `forget_working()`'s existing cascade.

**Code:** Adds `DELETE FROM annotations WHERE memory_id IN (SELECT ...)` before each delete, same predicate. Straightforward. Transaction-wrapped in trim path.

**Tests:** New test file verifies the trim cascade. Existing `forget_working()` behavior unchanged (already cascaded).

**Risk:** Minimal. Deleting annotations for a memory that had none is a no-op. The only risk is if the subquery and the delete see different row sets between execution — but both use the same params in one transaction, so that's handled.

**Verdict:** LGTM. Clear, correct, well-factored.

PR #260, opened 2026-06-09
feat(core): owner-scoped canonical (single-source-of-truth) facts (#256)

### LINKED ISSUE #256: Feature: per-profile canonical (single-source-of-truth) facts — a new owner-scoped memory layer
### Vantage point

We build **long-running conversational companion personas** on top of Hermes + Mnemosyne. What makes such a persona feel real is twofold: *associative recall* (which Mnemosyne's vector+FTS hybrid already does well) and *identity consistency over long timelines* — the persona must not contradict its own stable self-facts months apart. This proposal targets the second axis.

### Context

We currently repurpose the **global shared surface** (`mnemosyne_shared_*`) as a curated identity layer — a stand-in for Hermes Agent's built-in file memory (`MEMORY.md` / `USER.md`). That works, but it **overloads the shared surface for the wrong job**: the shared surface is designed to share memory *between different agent profiles*, whereas a persona's identity is single-profile and should not be visible to other profiles.

This proposal introduces the missing piece as a **new concept**, so the shared surface can go back to being used exactly as intended (cross-profile sharing), untouched.

### The gap

A persona's stable self-facts are stored today as ordinary working/episodic entries, with **no uniqueness or canonicalization guarantee**. Over a long-lived conversation:

- Restating a stable self-fact accumulates near-duplicate entries.
- A later change coexists with the stale value rather than superseding it.
- Recall then surfaces **conflicting copies**, with no single authoritative value to read back — the persona can describe itself two different ways.

Mnemosyne already solves this for clean relational facts via the **TripleStore** (`valid_until` supersession on `(subject, predicate)`). But that doesn't cover free-text identity "cards" that don't reduce to a single S-P-O triple, and it isn't the right home for a per-profile authoritative value store. There is currently no owner-scoped single-source-of-truth layer.

### Two orthogonal axes

The key insight is that "canonical-ness" (a uniqueness/upsert discipline) is **independent** of "shared-ness" (who can read it). Today only one of these cells exists at the canonical level:

|                              | Free-text / episodic                | **Canonical (single source of truth)**     |
|------------------------------|-------------------------------------|---------------------------------------------|
| **Private (one owner)**      | per-profile lived-experience recall | **← proposed: a profile's canonical self**  |
| **Shared (all profiles)**    | the existing shared surface         | (out of scope; optional future)             |

This proposal adds the **top-right** cell only. The shared surface (bottom row) is explicitly left as-is.

### Prior art / inspiration

[Sibyl-Memory](https://github.com/Sibyl-Labs/Sibyl-Memory) (a separate, lexical/FTS-only system) enforces *"single source of truth per entity"* via `UNIQUE (tenant_id, category, name)`. We don't think a lexical-only store is the right foundation for persona recall — semantic retrieval matters too much — but that invariant is embedding-agnostic and clean, and its `tenant_id` is precisely the **per-owner isolation** that keeps one profile's canonical facts out of another's.

### Proposal

Add an **owner-scoped canonical-fact layer** to `BeamMemory`:

- A canonical store keyed by **`UNIQUE(owner_id, category, name)`** with **upsert-in-place** semantics — each `(owner, category, name)` slot holds exactly one current value. `owner_id` is the profile/persona (Mnemosyne already has the hook for this via `profile_isolation` / scope namespacing).
- On update, the prior value is **versioned to history** (mirroring the TripleStore's `valid_until` pattern) rather than lingering as a live duplicate, preserving auditability.
- New owner-scoped tools (operating on the calling profile's own bank, **not** the shared surface):
  - `mnemosyne_remember_canonical(category, name, body)` → upsert the canonical slot
  - `mnemosyne_recall_canonical(category, name)` → read the single authoritative value
  - existing full-text `search` continues to span these entries
- Isolation is automatic: two profiles (e.g. two distinct personas) each get their own canonical namespace via `owner_id`; a non-persona profile (e.g. a research agent) simply has none. They can still exchange memories through the existing shared surface, which is unchanged.
- Canonical facts complement, not replace, episodic memory and the TripleStore: relational facts still belong in triples; free-text identity cards get a deduped authoritative slot.

### Why it fits Mnemosyne's principles

- **Zero new dependencies** — one SQLite table + a unique index.
- **Local-first / zero-config** — opt-in and additive; default behavior unchanged.
- **Fast** — an indexed point lookup on `(owner_id, category, name)` is cheaper than hybrid vector search for known-key identity reads.

### Backward compatibility

Fully opt-in and additive. The **shared surface is untouched** and keeps working in its intended cross-profile role. Existing private/episodic behavior is unchanged; the canonical layer is a new, separate table reached only through the new tools.

### Alternatives considered

- **Put canonical facts on the global shared surface** — rejected: "global" means every profile sees it, which leaks one persona's identity to others. Canonical and shared are orthogonal.
- **Rely solely on the TripleStore** — great for relational S-P-O facts, but not free-text identity cards, and not an owner-scoped authoritative value store.
- **Client-side dedup in the provider** — brittle and re-implemented by every consumer; a first-class store guarantee is cleaner and benefits all clients.

---

Implementation would span core `BeamMemory` (table + upsert/versioning) and the Hermes provider (new owner-scoped tools), so it may need routing across both layers. Happy to follow up with a PR (schema + upsert/versioning path + owner-scoped tool wiring + tests) if the direction looks good.

Milgauss:
Closes #256.

## Summary

Adds an **owner-scoped canonical-fact layer** — the missing *private + canonical* cell from the matrix in #256. Each `(owner_id, category, name)` slot holds exactly **one** current value, giving long-running personas identity consistency over time: restating a stable self-fact is a no-op (no duplicate accumulation), and a new value supersedes the old one rather than coexisting with it.

This is the piece that lets the **shared surface go back to being used exactly as intended** (cross-profile sharing) — canonical-ness and shared-ness are orthogonal, and this PR adds only the top-right cell.

## Design

Per the issue's own constraint — *"zero new dependencies — one SQLite table + a unique index"* — this is implemented with **one table + a partial unique index**, no FTS table:

- `canonical_facts(id, owner_id, category, name, body, source, confidence, version, valid_from, valid_until, created_at)`.
- `CREATE UNIQUE INDEX … ON canonical_facts(owner_id, category, name) WHERE valid_until IS NULL` — the partial index reconciles *"single source of truth"* with *"keep history in the same table"*: only the live row participates in the constraint, superseded rows accumulate unconstrained. This mirrors `TripleStore`'s `valid_until` supersession, with an `owner_id` dimension added.
- `remember()` is upsert-in-place: identical body → no-op; new body → stamp `valid_until` on the current row, insert the new value with `version + 1`.
- `recall()` is an indexed point lookup of the single current row — cheaper than hybrid vector search for a known-key identity read.

`CanonicalStore` follows the `AnnotationStore` template exactly (shared-connection or standalone construction, `init_*` / `_init_*_with_conn`, `export_all`/`import_all` parity with the sibling stores, module-level convenience functions).

## What's included

- **`mnemosyne/core/canonical.py`** — `CanonicalStore` with `remember`, `recall`, `list`, `history`, `search` (owner-scoped substring), `forget` (retire-as-history), and export/import.
- **`BeamMemory`** exposes `self.canonical` on its shared thread-local connection (no extra file descriptor), alongside `self.annotations`.
- **Two new tools** — `mnemosyne_remember_canonical` and `mnemosyne_recall_canonical` — on both the Hermes provider (`hermes_memory_provider`) and the MCP surface (`mnemosyne.mcp_tools` / `mnemosyne_hermes.tools`). Tool count 23 → 25. `recall_canonical` covers exact-slot read, category / whole-bank listing, version history, and owner-scoped search, so full-text lookups span canonical entries **without touching the default `mnemosyne_recall` path** (zero behavior change).
- **Owner isolation by construction:** the provider derives `owner_id` from the active profile identity and never reads it from tool args, so a profile cannot read or write another profile's canonical bank. (Over MCP the owner is a deployment setting, `MNEMOSYNE_DEFAULT_OWNER`.)

## Backward compatibility

Fully additive and opt-in. `canonical_facts` is created lazily on first init; existing tables, tools, and recall output are unchanged. The shared surface is untouched.

## Tests

`tests/test_canonical.py` (27 cases): upsert / no-op / supersession + versioning, partial-unique-index enforcement (incl. a hand-inserted duplicate live row rejected), owner isolation across `recall`/`list`/`search`, list/search/history read paths, `forget` semantics, export/import round-trip + idempotent re-import, `BeamMemory` wiring (shared connection), and provider tool round-trip incl. owner defaulting/isolation. Existing provider + MCP tool-registration counts updated 23 → 25.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [3.5.0] — 2026-06-09
+
+### Added
+
+- **Owner-scoped canonical (single-source-of-truth) facts** (issue #256). A new
+  `CanonicalStore` (`mnemosyne/core/canonical.py`) gives long-running personas an
+  identity layer where each `(owner_id, category, name)` slot holds exactly one
+  current value. Restating a stable self-fact is a no-op (no duplicate
+  accumulation); a new value supersedes the old one, which is preserved as
+  history — the TripleStore `valid_until` pattern, extended with an owner
+  dimension. Implemented as **one SQLite table plus a partial unique index**
+  (`… WHERE valid_until IS NULL`); no new dependency, no FTS table.
+  - Two new tools, `mnemosyne_remember_canonical` and `mnemosyne_recall_canonical`
+    (the latter covers exact-slot read, category/whole-bank listing, version
+    history, and owner-scoped substring search). Exposed on both the Hermes
+    provider and the MCP surface — total tool count 23 → 25.
+  - `BeamMemory` now exposes `self.canonical`, sharing its thread-local
+    connection (no extra file descriptor), mirroring `self.annotations`.
+  - Owner isolation is enforced by construction: the provider derives `owner_id`
+    from the active profile identity and never reads it from tool args, so one
+    profile cannot read or write another's canonical bank. The shared surface is
+    untouched and keeps its cross-profile role.
+  - Fully additive and opt-in: the `canonical_facts` table is created lazily on
+    first init; existing tables, tools, and recall output are unchanged.
+
 ## [3.4.0] — 2026-06-01
 
 ### Added
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -538,6 +538,46 @@ triples = store.get_triples_for_subject("memory_id_abc")
 
 ---
 
+## CanonicalStore (Single Source of Truth)
+
+**Module:** `mnemosyne.core.canonical`
+
+Owner-scoped canonical facts. Each `(owner_id, category, name)` slot holds
+exactly **one** current value — the right home for a persona's stable identity
+cards (name, voice, durable preferences) that must not contradict themselves
+over time. Restating a value is a no-op; a new value supersedes the old one,
+which is preserved as history (the TripleStore `valid_until` pattern, with an
+owner dimension). Backed by one SQLite table plus a partial unique index — no
+new dependency.
+
+```python
+from mnemosyne.core.canonical import CanonicalStore
+
+store = CanonicalStore(db_path)
+
+# Upsert the canonical value for a slot (returns the current row + status).
+store.remember("jessi", "identity", "name", "My name is Jessi.")   # status="created"
+store.remember("jessi", "identity", "name", "My name is Jessi.")   # status="unchanged" (no-op)
+store.remember("jessi", "identity", "name", "I go by Jess now.")    # status="updated"
+
+store.recall("jessi", "identity", "name")["body"]      # "I go by Jess now."
+store.list("jessi")                                    # all current slots for the owner
+store.list("jessi", category="identity")               # one category
+store.history("jessi", "identity", "name")             # every version, newest first
+store.search("jessi", "Jess")                          # owner-scoped substring search
+store.forget("jessi", "identity", "name")              # retire a slot (kept as history)
+```
+
+When used through `BeamMemory` the store is available as `beam.canonical`,
+sharing the beam's connection. Two profiles each get an isolated namespace via
+`owner_id`; the shared surface remains the place for *cross-profile* sharing.
+
+Exposed as the `mnemosyne_remember_canonical` and `mnemosyne_recall_canonical`
+tools (see below). The provider derives `owner_id` from the active profile, so a
+profile cannot reach another profile's canonical bank.
+
+---
+
 ## MCP Server
 
 **Module:** `mnemosyne.mcp_server`
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -531,6 +531,51 @@ TRIPLE_QUERY_SCHEMA = {
     },
 }
 
+REMEMBER_CANONICAL_SCHEMA = {
+    "name": "mnemosyne_remember_canonical",
+    "description": (
+        "Store a CANONICAL (single-source-of-truth) self-fact for the current "
+        "profile. Each (category, name) slot holds exactly one current value: "
+        "restating the same body is a no-op, and a new body supersedes the old "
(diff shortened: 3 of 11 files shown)


AxDSan (commented):
## Code Review — PR #260: Canonical Facts

### Verdict: Approve (pending one minor fix)

**Author:** @Milgauss
**Scope:** 1259 + 6 -, 11 files, 27 new tests
**Status:** All 82 tests pass (canonical + MCP + provider)

### What This Does

Adds a CanonicalStore — owner-scoped single-source-of-truth facts. Each (owner_id, category, name) slot holds exactly one current value via a partial unique index. Restating the same body is a no-op; a new body supersedes the old one (preserved as history). No new dependencies (one SQLite table + one partial unique index).

### Architecture

- `mnemosyne/core/canonical.py` — CanonicalStore with remember/recall/list/history/search/forget + export/import
- `BeamMemory` wires it as `self.canonical` sharing the beam's thread-local connection (7 lines in beam.py)
- Both Hermes provider (`mnemosyne_remember_canonical` / `mnemosyne_recall_canonical`) and MCP surface
- Owner isolation enforced by construction (derived from profile identity / MNEMOSYNE_DEFAULT_OWNER, never from tool args)
- Version bumped 3.4.0 → 3.5.0

### What I Like

- Follows the AnnotationStore template exactly — consistent patterns make the codebase easier to maintain
- Partial unique index is elegant: reconciles "single source of truth" with "keep history in same table"
- Owner isolation is correct — provider derives owner from internal state, not tool args
- Fully additive/opt-in — nothing changes for existing users
- Docs in api-reference.md and CHANGELOG.md updated

### One Nit

The MCP tool handlers in `mcp_tools.py` construct a CanonicalStore with `db_path=db_path, conn=mem.beam.conn` when `mem.beam.canonical` isn't yet set. This is belt-and-suspenders safety, but the 23→25 count mismatch in the MCP tests revealed an install-ordering issue (the editable install of `integrations/hermes/` was stale). Not a bug in the PR — the schemas and handler wiring are correct.

Clean work. Merging.

AxDSan (approved):
Clean architecture, 27 tests pass, fully additive. LGTM.

PR #263, opened 2026-06-10
fix(prefetch): always inject per-contact identity, independent of recall query

aldoeliacim:
## The bug: semantic-recall-gated identity → wrong identity on short openers

Per-contact identity memories (`source='identity'`, session-scoped, high
importance ~0.95, one silo per session) answer the question *"who am I talking
to?"*. The Claude Code memory provider surfaces context via
`MnemosyneMemoryProvider.prefetch()`, which is profile-driven and ultimately
calls `_prefetch_bank()` → `self._beam.recall(query=user_message, top_k=...)`.

The problem: identity memories **only surface if they win SEMANTIC recall**
against the user's message. When a user sends a short/generic message ("Hi",
"Hola", "thanks", or just a nickname), there is no semantic match to the
identity text (e.g. "Contact A is the lead engineer on the payments team"), so:

1. the identity memory never enters recall's `top_k` window,
2. the importance filter (which runs **after** recall) never sees it,
3. the agent loses track of who it's talking to and defaults to the wrong
   identity.

Reproduced deterministically: a generic opener yields a prefetch context with
no identity, even though a high-importance identity row exists for the active
session.

## The fix: deterministic, query-independent, session-scoped injection

Add a query-**independent** identity injection to `prefetch()`:

- **`_identity_fichas()`** — reads ALL identity rows for the **active**
  `session_id` straight from `working_memory` with a direct SQL read
  (`WHERE source='identity' AND session_id=?`), ordered by importance then
  recency. Strictly session-scoped, so there is **zero cross-session leakage**.
  Wrapped in try/except → non-fatal on any error.
- **`_prefetch_identity()`** — renders those rows in the **same format as the
  memory bank**, tagged `[IDENTITY]`:
  `  [{ts}] (importance {imp:.2f}) [IDENTITY] {content}`, deduplicated against
  whatever recall already surfaced (so a query that *does* match the identity
  never produces a duplicate).
- In `prefetch()`, the identity block is merged to the **FRONT** of the
  returned context, kept behind the existing
  `if not self._beam or self._agent_context in self._skip_contexts: return ""`
  guard.

The owner/empty-identity case (no identity rows) is a strict **no-op** —
existing behavior is preserved exactly. The change is additive and minimal.

## Tests

`tests/test_prefetch_identity_always_inject.py` proves:

- **(a)** an identity row surfaces on a generic query (`"Hi"`) that does NOT
  semantically match it;
- **(b)** identity stored under session A does **not** leak into session B's
  prefetch (and session A still sees its own);
- **(c)** no duplicate when a query DOES match the identity (recall + always-inject);
- **(d)** the no-identity-rows case is a no-op (no `[IDENTITY]` block).

Local run (new + existing prefetch/provider/identity/isolation suites):

```
tests/test_prefetch_identity_always_inject.py ....                 [ 4 passed ]
tests/test_prefetch_profiles.py / _content_truncation / _low_quality   [27 passed total]
tests/test_hermes_memory_provider.py + identity + multi_agent_identity
  + thread_isolation                                                [81 passed]
```

`python -c "import ast; ast.parse(open('hermes_memory_provider/__init__.py').read())"` → OK.

No private user data, real contact names, phone numbers, or chat IDs are
included — generic placeholders ("Contact A", "session-A") are used throughout.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1331,9 +1331,90 @@ class MnemosyneMemoryProvider(MemoryProvider):
                 block = ""
             if block:
                 blocks.append(block)
+        # Per-contact identity memories must surface on EVERY turn, independent
+        # of the semantic recall query. Routing them through recall is a latent
+        # bug: a short/generic opener ("Hi", a nickname) does not match the
+        # identity text, so it never enters recall's top_k window and the
+        # importance filter never sees it -- the agent then loses track of who
+        # it is talking to. Inject them deterministically at the FRONT, scoped
+        # strictly to the active session_id, deduplicated against whatever
+        # recall already surfaced. No identity rows == no-op (legacy behavior).
+        identity_block = self._prefetch_identity(blocks, profile)
+        if identity_block:
+            blocks.insert(0, identity_block)
         if profile.dedup:
             blocks = _dedup_blocks(blocks)
         return "\n\n".join(b for b in blocks if b)
+
+    def _prefetch_identity(self, existing_blocks: List[str], profile: "PrefetchProfile") -> str:
+        """Render the always-inject identity block for the active session.
+
+        Pulls identity memories straight from the active ``session_id`` (see
+        ``_identity_fichas``) and renders them in the same format as the memory
+        bank, tagged ``[IDENTITY]``. Rows whose content already appears in the
+        blocks recall produced are dropped, so a query that *does* match the
+        identity never yields a duplicate. Returns ``""`` when there is nothing
+        to inject.
+        """
+        rows = self._identity_fichas()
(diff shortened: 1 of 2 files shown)


AxDSan:
@aldoeliacim nice work. The bug analysis is spot on - routing "who am I" through semantic recall was a latent footgun. The fix is clean.

Verified locally on my end:
- All 4 new tests pass
- 59 existing tests still pass
- CI green

The design choices I checked:
- parameterized SQL, no injection risk
- session-scoped, no cross-session leak
- beam=None / empty-identity both no-op correctly
- dedup against recall output prevents duplication

Solid testing with the shared-DB cross-session isolation fixture. Merging now.

PR #268, opened 2026-06-10
feat(hermes): add bank_id config key for explicit bank selection

### LINKED ISSUE #267: Explicit bank selection via config key
With `profile_isolation: true`, the bank name is always derived from the profile identity through `_resolve_profile_bank()`. When the profile name maps to "default" (or one of the other filtered values like "primary", "none"), theres no way to route to a specific bank.

Hermes doesnt support renaming the default profile ([hermes-agent#21105](https://github.com/NousResearch/hermes-agent/issues/21105)), and the codebase has hardcoded assumptions around "default" being the profile name. So users stuck on the default profile cant work around this by renaming it.

This came up in #214 where `bank_id` was mentioned as a working config key for this, but its not documented and I couldnt confirm whether its still present in the current codebase.

**What Im looking for:** a config key under `memory.mnemosyne` (lets call it `bank_id`) that takes priority over the auto-derived bank name in `_resolve_profile_bank()`. Something like:

```python
def _resolve_profile_bank(self) -> str:
    # explicit override wins
    bank_id = self._read_config_key("bank_id")
    if bank_id:
        return self._sanitize_bank_name(bank_id)
    # existing derivation logic unchanged
    ...
```

This would let the default profile (and any other profile with a generic name) target a specific bank without renaming the profile.

If `bank_id` already works and I missed it, just documenting it in `get_config_schema()` would be enough. Happy to open a PR if that is the preferred path.

chdlc:
## Description

Adds `memory.mnemosyne.bank_id` config key and `MNEMOSYNE_BANK_ID` env var to override the auto-derived bank name in `_resolve_profile_bank()`.

When `profile_isolation` is enabled, the bank name is always derived from `agent_identity` or `hermes_home`. If the profile name maps to "default" (or other filtered values), there is no way to route to a specific bank. This lets users explicitly set a bank name regardless of profile identity.

Precedence: kwargs > config.yaml > `MNEMOSYNE_BANK_ID` env var > existing derivation logic.

## Related Issue

Closes #267

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [ ] Manual verification

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [3.6.0] — 2026-06-10
+
+### Added
+
+- **Explicit bank selection via `bank_id` config key** (issue #267). New
+  `memory.mnemosyne.bank_id` config key and `MNEMOSYNE_BANK_ID` env var allow
+  overriding the auto-derived bank name from `agent_identity` / `hermes_home`.
+  Useful when the Hermes profile name cannot be changed (e.g. "default") but a
+  specific bank is needed. Precedence: kwargs > config.yaml > env var.
+  Applied to both `hermes_memory_provider` and `integrations/hermes` implementations.
+
 ## [3.5.0] — 2026-06-10
 
 ### Added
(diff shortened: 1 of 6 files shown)


chdlc:
@AxDSan quick check on #268: it shows as merged on my side (dab9717, Jun 10 20:22 UTC), but the changes aren't visible on main — current HEAD is 5127772 from later that day, and the `bank_id` references aren't in `hermes_memory_provider/__init__.py`. The head ref (3aa1a6b on `feat/bank-id-override`) is still there.

Should I rebase and reopen for the next release, or is this staying closed for now?

PR #271, opened 2026-06-11
fix: Russian FTS5 fallback for inflected Cyrillic recall

### LINKED ISSUE #168: [BUG] FTS5 full-text search produces zero results for CJK (Chinese/Japanese/Korean) text due to missing tokenizer configuration
## Summary

Mnemosyne's FTS5 tables (`fts_episodes`, `fts_working`, `fts_facts`) are created without specifying a tokenizer, defaulting to SQLite's `unicode61`. This tokenizer splits on whitespace and punctuation only. Since CJK text has no word-separating spaces, entire sentences are indexed as a single token, making keyword search completely ineffective for CJK users.

## Environment

- Mnemosyne version: v3.0.0
- Python: 3.x
- SQLite: built-in
- Language of conversation: Chinese (Simplified)

## Steps to Reproduce

```python
from mnemosyne import remember, recall

remember("用户服务器部署在腾讯云国内节点")
results = recall("腾讯云")
print(results)  # returns [] — expected non-empty
```

Direct SQLite verification:
```sql
SELECT * FROM fts_episodes WHERE fts_episodes MATCH '腾讯云';
-- 0 rows, even though content exists in episodic_memory
```

## Expected Behavior

`recall("腾讯云")` should match memories containing "腾讯云", just as `recall("docker")` matches English content.

## Actual Behavior

FTS component returns 0 results for all CJK queries. The hybrid recall score (50% vector + **30% FTS** + 20% importance) silently degrades to ~71% vector + 29% importance for all CJK users.

## Root Cause

`beam.py` creates FTS5 tables without a tokenizer:

```sql
CREATE VIRTUAL TABLE IF NOT EXISTS fts_episodes USING fts5(
    content,
    content='episodic_memory',
    content_rowid='rowid'
    -- no tokenize= specified → defaults to unicode61
    -- "腾讯云VPS" is indexed as one token
    -- query "腾讯云" never matches
)
```

The `_fts_search()` sanitizer (`re.sub(r'[^\w\s]', ' ', query)`) correctly passes CJK characters through `\w`, but the indexing failure means no match is ever found.

## Proposed Fix — Bigram Pre-indexing

Pre-process text into character bigrams before indexing, and apply the same transform at query time. No external dependencies required.

```python
def _cjk_bigram(text: str) -> str:
    """Convert CJK text to bigram tokens for FTS5 indexing.
    CJK chars → unigram + bigram. Latin/digits → kept as-is.
    '腾讯云VPS' → '腾 腾讯 讯 讯云 云 VPS'
    """
    import re
    result = []
    buf = ''
    for i, c in enumerate(text):
        if '\u4e00' <= c <= '\u9fff' or '\u3400' <= c <= '\u4dbf':
            if buf:
                result.extend(re.findall(r'[a-zA-Z0-9_]+', buf))
                buf = ''
            result.append(c)  # unigram
            if i + 1 < len(text) and '\u4e00' <= text[i+1] <= '\u9fff':
                result.append(c + text[i+1])  # bigram
        else:
            buf += c
    if buf:
        result.extend(re.findall(r'[a-zA-Z0-9_]+', buf))
    return ' '.join(result) if result else text
```

Tested results:

| Query | Before | After |
|-------|--------|-------|
| `腾讯云` | ❌ 0 results | ✅ matches |
| `部署` | ❌ 0 results | ✅ matches |
| `Docker` | ✅ matches | ✅ matches |
| `tools模式` | ❌ 0 results | ✅ matches |

## Why Not `trigram` Tokenizer?

SQLite's built-in `trigram` tokenizer requires a minimum of 3 characters. Two-character Chinese words (腾讯, 部署, 然然) are extremely common and would still fail to match. Bigram pre-indexing handles all lengths with zero additional dependencies.

## Required Changes (all in `beam.py`)

1. Add `_cjk_bigram(text)` utility function
2. Register it as a SQLite UDF on each connection: `conn.create_function("cjk_bigram", 1, _cjk_bigram)`
3. Update FTS triggers to use `cjk_bigram(new.content)` instead of `new.content`
4. Apply `_cjk_bigram()` to query string in `_fts_search()` and `_fts_search_working()`
5. Add migration block to rebuild FTS index for existing databases

Happy to submit a PR if the approach looks good.

Whishp:
Like the CJK fallback added in issue #168, this routes Russian (and other `[а-яё]`-covered Cyrillic) queries to a LIKE-based candidate scan re-ranked by trigram Jaccard similarity when the default unicode61 FTS5 tokenizer returns zero rows.

The default tokenizer has no built-in stemmer for inflected Russian surface forms, so a single-lemma query would otherwise only match the exact surface form in storage.

**Structure** (mirrors the CJK fallback):
- `_has_cyrillic` detects the script — covers Russian and some Cyrillic neighbours, but **not** Serbian/Mongolian Cyrillic (ђ ћ џ њ љ ө ү). Documented for future extension.
- `_ngrams` + `_cyrillic_score` provide the trigram Jaccard ranking signal
- `_cyrillic_like_search` wires the LIKE-based candidate generation with a `_py_lower` UDF for Unicode-aware case folding (SQLite's default `LOWER()` is ASCII-only)
- Four integration points in `_fts_search` and `_fts_search_working` (empty-terms branch + post-MATCH zero-rows branch, for both episodic and working memory)

**Priority note:** CJK fallback is checked first. If a query contains both CJK and Cyrillic characters (unlikely in practice), the CJK path takes precedence.

**No new dependencies, no schema migration, no FTS5 schema changes, no version bump.**

Includes 27 unit tests covering detection, n-gram scoring, inflection overlap, ranking order, and end-to-end routing through the FTS5 entry points.

Closes #168 family — Cyrillic counterpart of the CJK fallback.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1846,6 +1846,167 @@ def _cjk_like_search(conn: sqlite3.Connection, query: str, k: int = 20, working:
     return [{result_key: rid, "rank": -score} for rid, score in scored[:k]]
 
 
+# ---------------------------------------------------------------------------
+# Cyrillic FTS5 fallback (Russian)
+#
+# Russian is morphologically rich: a single lemma produces many inflected
+# surface forms (тёмная / тёмную / тёмной, встреча / встречи / встречу).
+# The default unicode61 FTS5 tokenizer has no built-in stemmer for Russian,
+# so direct FTS5 queries match only the exact surface form.
+#
+# When a Russian query produces zero FTS5 results, we fall back to a
+# LIKE-based scan and rank candidates by trigram Jaccard similarity
+# between query and content words. This mirrors _cjk_like_search
+# (issue #168) but with n-gram scoring, which is significantly more
+# discriminative for inflected languages than the single-character
+# coverage used for CJK.
+#
+# NOTE: the regex [а-яёА-ЯЁ] covers Russian and some Cyrillic neighbours
+# (Ukrainian, Bulgarian, etc.) but does NOT include Serbian/Mongolian
+# Cyrillic characters (ђ ћ џ њ љ ө ү). These can be added later if needed.
+#
+# No new dependencies, no schema migration, no FTS5 changes. The fallback
+# activates only when FTS5 returns zero rows AND the query contains
+# Cyrillic characters.
+#
+# Priority: if a query contains both CJK and Cyrillic, the CJK fallback
+# takes precedence (checked first). This is unlikely in practice but
+# documented here for clarity.
+# ---------------------------------------------------------------------------
+
+_CYRILLIC_CHAR_RE = re.compile(r"[а-яёА-ЯЁ]")
+
+
+def _has_cyrillic(text: str) -> bool:
+    """True if text contains at least one Russian/Cyrillic character.
+
+    Matches [а-яёА-ЯЁ] which covers Russian and some neighbours but
+    not Serbian/Mongolian Cyrillic — see module-level note.
+    """
+    return bool(_CYRILLIC_CHAR_RE.search(text))
+
+
+def _ngrams(s: str, n: int = 3) -> set:
+    """Return the set of length-n sliding n-grams of s.
+
+    For strings shorter than n, returns the whole string as a single
+    n-gram to keep the Jaccard math well-defined.
+    """
+    if len(s) < n:
+        return {s}
+    return {s[i:i + n] for i in range(len(s) - n + 1)}
+
+
+def _cyrillic_score(query: str, content: str, n: int = 3) -> float:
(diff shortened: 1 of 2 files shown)


AxDSan:
@Whishp solid work on the Cyrillic FTS5 fallback.

Verified locally: 27/27 new tests pass, 150/150 existing recall/search/beam tests pass. CI green across all 6 checks. No regressions, no schema migration, no new deps.

The structure mirrors the CJK fallback cleanly. Good call documenting the Serbian/Mongolian Cyrillic gap for future extension. Merging now.

One thing worth considering for a follow-up: the `_py_lower` UDF is registered on-demand at fallback time. If someone runs a mix of Cyrillic + Latin queries in the same session, that registration path gets entered once and sticks. Thats fine for correctness, but making it unconditional at connection init would simplify the code path slightly. Not blocking at all, just a thought.

PR #276, opened 2026-06-11
fix(triples): restore temporal-triple lifecycle regressed after #246

Milgauss:
## Summary

The temporal-triple lifecycle added in #246 (merged 2026-06-08, merge commit `63fbb48`) is no longer present on `main` — `mnemosyne/core/triples.py` and `hermes_memory_provider/__init__.py` are back to their pre-#246 base. A later merge on `main` appears to have reverted it (likely a branch cut before #246 that overwrote `triples.py` on merge), so the lifecycle is **absent from the v3.5.0 and v3.6.0 releases** even though #246 shows as merged.

This re-applies #246; the diff applies cleanly to current `main`.

### `TripleStore` (`mnemosyne/core/triples.py`)
- `add(..., valid_until=None, supersede=True)` — `supersede=False` skips the auto-close so a `(subject, predicate)` can hold multiple simultaneous values (e.g. `('user','speaks','English')` + `('user','speaks','Spanish')`); `valid_until` sets an explicit expiry on insert.
- new `end(subject, predicate, object=None, valid_until=None)` — expire open triples **without** replacing them; returns the count closed.
- `query` subject match is case-insensitive (`COLLATE NOCASE`).
- module-level `add_triple` gains the `valid_until`/`supersede` passthrough; new `end_triple`.

### Provider (`hermes_memory_provider/__init__.py`)
- `mnemosyne_triple_add` gains `valid_until` / `source` / `confidence` / `supersede`.
- new `mnemosyne_triple_end` tool.
- `mnemosyne_triple_query` gains `as_of`.

**Default behavior is unchanged** — `add` still supersedes by default (single-valued); everything new is additive / opt-in.

### Tests

Re-adds `tests/test_triples_lifecycle.py` from #246 — covers multi-valued coexistence, `end` expiry, default supersede, explicit `valid_until`, `as_of` history, and case-insensitive subject. All pass:

```
7 passed
```

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -623,7 +623,10 @@ TRIPLE_ADD_SCHEMA = {
     "name": "mnemosyne_triple_add",
     "description": (
         "Add a temporal fact triple (subject, predicate, object) to the knowledge graph. "
-        "Example: ('user', 'prefers', 'neovim'). Use for structured relationships."
+        "Example: ('user', 'prefers', 'neovim'). Use for structured relationships. "
+        "By default a new triple supersedes any prior fact with the same subject+predicate; "
+        "set supersede=false for multi-valued facts that should coexist "
+        "(e.g. ('user','speaks','English') and ('user','speaks','Spanish'))."
     ),
     "parameters": {
         "type": "object",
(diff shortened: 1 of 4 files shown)


AxDSan:
@Milgauss the PR looks clean. I verified the diff matches what #246 shipped; the regression is real and your re-application is exact. CI failure was the tool counter (assert expects 25, but `mnemosyne_triple_end` makes 26). Patched that and pushed to your branch.

Merging now. Thanks for catching this and doing the restore.

PR #282, opened 2026-06-12
fix(hermes): scan full __init__.py in is_installed() instead of first 4096 bytes

### LINKED ISSUE #281: [BUG] mnemosyne-hermes status reports "Not installed" on a working install: is_installed() truncates the __init__.py scan to 4096 bytes
## Summary

On a correctly installed and **functionally working** Hermes integration, `mnemosyne-hermes status` prints `Not installed`, while `mnemosyne-hermes install` on the same machine reports the symlink `already exists` / `Installed`. The two subcommands disagree.

Root cause: `is_installed()` validates the plugin by string-scanning only the **first 4096 bytes** of `__init__.py`, but in 0.1.7 the marker symbols live far past that offset, so the check always returns a false negative.

This is cosmetic (the provider loads and works), but it actively misleads users into thinking the integration failed. It is the same class of confusion reported in #240 and #253 (both now closed), so this looks like a residual instance in the current release.

## Environment

| Item | Detail |
| -- | -- |
| OS | Ubuntu 24.04, aarch64 (Oracle Cloud Ampere A1) |
| Python (Hermes venv) | 3.11.15 |
| venv manager | `uv` |
| mnemosyne-hermes | **0.1.7** |
| mnemosyne-memory | 3.6.0 |
| fastembed | 0.8.0 |

## Reproduction

```bash
# install into the Hermes venv
VIRTUAL_ENV=~/hermes-agent/venv uv pip install mnemosyne-hermes
export HERMES_HOME=~/.hermes

mnemosyne-hermes install            # -> "Installed. Symlink at ~/.hermes/plugins/mnemosyne -> .../mnemosyne_hermes"
ls -la ~/.hermes/plugins/mnemosyne  # -> symlink present, resolves to the package
mnemosyne-hermes status             # -> "Not installed (expected symlink at ~/.hermes/plugins/mnemosyne)"   <-- WRONG
```

Independent confirmation that the provider is genuinely usable (Hermes' own discovery path):

```python
from plugins.memory import load_memory_provider
mp = load_memory_provider("mnemosyne")
print(type(mp).__name__, mp.is_available())   # -> MnemosyneMemoryProvider True
```

So `load_memory_provider("mnemosyne")` succeeds while `is_installed()` claims the plugin is absent.

## Root cause

`mnemosyne_hermes/install.py`, `is_installed()` (def at line 212), line **225**:

```python
def is_installed(*, hermes_home_path: str | Path | None = None) -> bool:
    target = plugin_target_dir(hermes_home_path)
    if not target.exists():
        return False
    init_file = target / "__init__.py"
    if not init_file.is_file():
        return False
    try:
        source = init_file.read_text(errors="replace")[:4096]      # truncated to 4096 bytes
        return "register_memory_provider" in source or "MnemosyneMemoryProvider" in source
    except Exception:
        return False
```

Measured on the installed 0.1.7 package:

```
$ wc -c mnemosyne_hermes/__init__.py
80344  mnemosyne_hermes/__init__.py

$ grep -boE 'register_memory_provider|MnemosyneMemoryProvider' mnemosyne_hermes/__init__.py | head
9255:MnemosyneMemoryProvider
79437:register_memory_provider

$ head -c 4096 mnemosyne_hermes/__init__.py | grep -oE 'register_memory_provider|MnemosyneMemoryProvider'
(no output)
```

The first marker is at byte **9255** and `register_memory_provider` at **79437**; the file is **80 344** bytes. The `[:4096]` window covers only the module docstring and imports, so neither symbol is ever seen and `is_installed()` returns `False`.

## Impact

- `status` reports a healthy install as broken (false negative).
- `install` post-message tells the user to run `... status` to verify, which then contradicts the install that just succeeded.
- No functional impact: the provider loads and `is_available()` is `True`.

## Suggested fix

Minimal, one line (scan the whole file instead of the first 4 KB):

```diff
-        source = init_file.read_text(errors="replace")[:4096]
+        source = init_file.read_text(errors="replace")
```

(`__init__.py` is ~80 KB; reading it fully for a CLI status check is negligible.)

More robust alternative: stop string-scanning entirely and validate by resolving the symlink and importing the module / checking for the `register_memory_provider` attribute, which is the same source of truth Hermes uses via `load_memory_provider`. That would also keep `status` and the real loader from ever diverging again.

## Notes

The byte offset will vary as `__init__.py` grows, but since it is already ~20x the 4096-byte window, any new symbol added near the top would be the only thing that could accidentally mask this. Reading the full file removes the size coupling.

greg00code:
## What

`is_installed()` in the Hermes integration only scanned the first 4096 bytes of the plugin `__init__.py` when looking for the `register_memory_provider` / `MnemosyneMemoryProvider` markers. In 0.1.7 those symbols are at byte 9255 and 79437 of an ~80 KB file, so the check never finds them and `mnemosyne-hermes status` reports `Not installed` on a working install, while `install` and Hermes' own `load_memory_provider` both succeed.

## Fix

Read the whole file instead of the first 4 KB (one line).

## Validation

Ubuntu aarch64, Python 3.11.15, mnemosyne-hermes 0.1.7, no other change:

- before: `mnemosyne-hermes status` -> `Not installed (expected symlink at ~/.hermes/plugins/mnemosyne)`
- after: `Installed. Symlink at ~/.hermes/plugins/mnemosyne -> .../mnemosyne_hermes` / `mnemosyne-memory 3.6.0 installed` / `Core library: OK`

The provider already loaded correctly throughout (`load_memory_provider("mnemosyne")` returns `MnemosyneMemoryProvider`, `is_available() == True`); only the status self-check was wrong.

Fixes #281

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -222,7 +222,7 @@ def is_installed(*, hermes_home_path: str | Path | None = None) -> bool:
     if not init_file.is_file():
         return False
     try:
-        source = init_file.read_text(errors="replace")[:4096]
+        source = init_file.read_text(errors="replace")
         return "register_memory_provider" in source or "MnemosyneMemoryProvider" in source
     except Exception:
         return False


AxDSan:
@greg00code - this fix is already on main in abaa7ea. I landed the same change independently while you were opening the PR (tight window - your commit at 09:06, mine at 09:07 UTC).

Your analysis in #281 was outstanding. The `wc -c` + `grep -bo` breakdown was the kind of hard data that makes a bug report instantly actionable. And your PR proved the fix independently - two people arriving at the same one-line change from opposite directions means it was the right call.

Closing this since the fix is merged. Thanks for the sharp work on both the report and the PR.

PR #284, opened 2026-06-13
fix(hermes): wire canonical fact tools in active adapter

WXBR:
## Summary

- Route `mnemosyne_remember_canonical` and `mnemosyne_recall_canonical` through the active `mnemosyne_hermes` adapter.
- Reuse the Beam-owned `CanonicalStore` when present, or attach one to the shared Beam connection when missing.
- Scope canonical facts to the active Hermes profile via provider state, not tool arguments.
- Add adapter-level regression tests for roundtrip, owner isolation, search/history/list modes, required-field validation, and schema exposure.

## Why

Canonical facts were present in core storage and tool schemas, but the active Hermes integration adapter did not dispatch the two canonical tools. That made the tools fail at runtime even though the schema advertised them.

## Verification

```bash
PYTHONPATH="$PWD:$PWD/integrations/hermes/src:${PYTHONPATH:-}" ~/.hermes/hermes-agent/venv/bin/python -m pytest \
  tests/test_canonical.py \
  integrations/hermes/tests/test_canonical_tools.py \
  tests/test_mcp_server.py::TestToolSchemas::test_all_tools_present \
  -q -o 'addopts='
# 33 passed in 0.47s

PYTHONPATH="$PWD:$PWD/integrations/hermes/src:${PYTHONPATH:-}" ~/.hermes/hermes-agent/venv/bin/python -m pytest \
  integrations/hermes/tests \
  tests/test_mcp_server.py::TestToolSchemas::test_all_tools_present \
  -q -o 'addopts='
# 9 passed in 0.33s
```

Also smoke-tested the active adapter against a temporary Hermes home/SQLite DB: remember returned `created`, recall returned `found=True` with the expected body.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/__init__.py
+++ b/integrations/hermes/src/mnemosyne_hermes/__init__.py
@@ -970,6 +970,10 @@ class MnemosyneMemoryProvider(MemoryProvider):
                 return self._handle_triple_add(args)
             elif tool_name == "mnemosyne_triple_query":
                 return self._handle_triple_query(args)
+            elif tool_name == "mnemosyne_remember_canonical":
+                return self._handle_remember_canonical(args)
+            elif tool_name == "mnemosyne_recall_canonical":
+                return self._handle_recall_canonical(args)
             elif tool_name == "mnemosyne_scratchpad_write":
                 return self._handle_scratchpad_write(args)
             elif tool_name == "mnemosyne_scratchpad_read":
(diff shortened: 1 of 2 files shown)


AxDSan:
@WXBR - solid PR. I can see the gap you caught: the canonical tool schemas were in the registry but handle_tool_call() had no dispatch for them. Your fix addresses exactly that.

Code review notes:
- _canonical_owner() scoping via _agent_identity (not tool args) is the right security model
- Lazy CanonicalStore attachment is clean but the getattr+create pattern is duplicated in both handlers. Worth extracting to a helper as a follow-up
- Tests cover roundtrip, isolation, search/history/list, validation, and schema exposure. Thorough coverage

CI status: 5/6 jobs passed (3.10, 3.12, 3.13, docs-check, build). The 3.11 runner is still in progress - looks like a transient infra hang, not a code issue. I'll merge as soon as CI clears or we decide to re-run the 3.11 leg. Expect merge within the hour.

AxDSan:
Merged in a6ce704. Thanks @WXBR for the sharp catch and the thorough tests. The canonical fact tools are live now.

One small follow-up: the lazy CanonicalStore getattr+create pattern is duplicated in both handlers. Dead simple extract to a helper if you feel like it, but the PR as-is is solid. Appreciate the contribution.

PR #285, opened 2026-06-13
fix(diagnose): downgrade optional local LLM fallback

WXBR:
## Summary
- fall back to package distribution metadata when `mnemosyne.__version__` is not exposed by the active import path
- classify `ctransformers` as an optional local-GGUF fallback dependency instead of a failed core dependency
- count optional dependency absence as non-failing diagnostic status
- add regression coverage for both cases

## Validation
- `PYTHONPATH="$PWD:$PWD/integrations/hermes/src:${PYTHONPATH:-}" ~/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_diagnose_optional_deps.py tests/test_hermes_memory_provider_diagnose_active_db.py -q -o 'addopts='`

### DIFF
--- a/mnemosyne/diagnose.py
+++ b/mnemosyne/diagnose.py
@@ -9,6 +9,7 @@ Never includes memory content, user queries, or API keys.
 Supports --fix mode: auto-installs missing dependencies.
 """
 
+import importlib.metadata
 import json
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


AxDSan:
@WXBR merged in 7bb21ca. Two birds with one PR, the ctransformers false alarm was my own pet peeve. The `__version__` fallback is a nice catch too. Solid as always.

Labels: bug, core.

PR #286, opened 2026-06-13
fix(hermes): route prompt to structured memory tools

WXBR:
## Summary

- Expand the Hermes adapter system prompt from basic remember/recall guidance into explicit Mnemosyne memory-layer routing.
- Nudge agents to use canonical facts, triples, graph links, validation/invalidation, scratchpad, and shared surface appropriately.
- Add a regression test so the structured-memory routing prompt does not silently narrow again.

## Why

The old prompt mentioned `mnemosyne_remember`, `mnemosyne_recall`, and `mnemosyne_shared_*`, which biases Hermes agents toward the simplest memory operations even though the adapter exposes richer structured memory tools.

This keeps the prompt concise while making the intended routing explicit.

## Tests

```bash
cd integrations/hermes
~/.hermes/hermes-agent/venv/bin/python -m py_compile src/mnemosyne_hermes/__init__.py tests/test_system_prompt.py
~/.hermes/hermes-agent/venv/bin/python -m pytest tests/test_system_prompt.py tests/test_canonical_tools.py tests/test_prefetch_hardening.py -q -o 'addopts='
```

Result: `9 passed`.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/__init__.py
+++ b/integrations/hermes/src/mnemosyne_hermes/__init__.py
@@ -752,10 +752,17 @@ class MnemosyneMemoryProvider(MemoryProvider):
             # (working / init-failed-visible / skip-context-silent).
             return (
                 "# Mnemosyne Memory\n"
-                "Active (native local memory). Use mnemosyne_remember to store ANY "
-                "durable fact, preference, identity, or insight. Use mnemosyne_recall to search. "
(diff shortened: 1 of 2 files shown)


AxDSan:
@WXBR clean PR. The new prompt is more explicit about routing to the right memory layer instead of defaulting to remember/recall. The regression test is a nice touch - prevents accidental narrowing.

Merging now.

PR #296, opened 2026-06-13
perf: split get_context hot path queries

### LINKED ISSUE #291: Optimize working_memory hot paths: get_context() and sleep() query plans
# Optimize working_memory hot paths: get_context() and sleep() query plans

## Summary

While reviewing Mnemosyne on `main` (`e7f3509`), I noticed two SQLite hot paths that appear to miss existing indexes and can become expensive as `working_memory` grows:

1. `BeamMemory.get_context()` currently performs a full table scan and temporary sort.
2. `BeamMemory.sleep()` does not use the existing unconsolidated-session index because the query wraps `session_id` in `COALESCE()`.

These paths are important because `get_context()` can run in the prompt-injection path and `sleep()` is used for consolidation/maintenance.

## Evidence

Repository: `AxDSan/mnemosyne`
Commit reviewed: `e7f3509`

Relevant code:

- `mnemosyne/core/beam.py`, `get_context()` around lines `3045-3102`
- `mnemosyne/core/beam.py`, `sleep()` query around lines `6958-6967`
- existing partial index: `idx_wm_unconsolidated ON working_memory(session_id, timestamp) WHERE consolidated_at IS NULL`

Observed query plans on a local Mnemosyne DB:

```text
get_context:
SCAN working_memory
USE TEMP B-TREE FOR ORDER BY

sleep:
SEARCH working_memory USING INDEX idx_wm_timestamp (timestamp<?)
```

The sleep query shape includes:

```sql
COALESCE(session_id, 'default') = ?
```

which prevents the existing `(session_id, timestamp)` partial index from being used effectively.

### 1. Split `get_context()` into index-friendly queries

Instead of one query with:

```sql
(session_id = ? OR scope = 'global')
ORDER BY CASE WHEN scope = 'global' THEN 0 ELSE 1 END, importance DESC, timestamp DESC
```

run two simpler queries:

- one for `scope = 'global'`
- one for `session_id = ?`

Then merge the small result sets in Python while preserving the desired priority ordering.

Possible indexes:

```sql
CREATE INDEX IF NOT EXISTS idx_wm_context_session
ON working_memory(session_id, importance DESC, timestamp DESC)
WHERE superseded_by IS NULL;

CREATE INDEX IF NOT EXISTS idx_wm_context_global
ON working_memory(scope, importance DESC, timestamp DESC)
WHERE superseded_by IS NULL;
```

### 2. Make `sleep()` avoid `COALESCE()` on the indexed column

For normal sessions, use:

```sql
session_id = ?
```

For the legacy/default `NULL` compatibility case, handle it explicitly, e.g.:

```sql
(session_id = ? OR (? = 'default' AND session_id IS NULL))
```

Alternatively, add an expression index for the current query shape:

```sql
CREATE INDEX IF NOT EXISTS idx_wm_sleep_coalesced
ON working_memory(COALESCE(session_id, 'default'), timestamp)
WHERE consolidated_at IS NULL AND (pinned IS NULL OR pinned = 0);
```

## Expected benefit

- Less scanning and temporary sorting in a prompt-context hot path.
- Faster consolidation scans as `working_memory` grows.
- Lower SQLite CPU and lock pressure in long-running agent sessions.

## Notes

This should be a small, testable PR. A regression test could use `EXPLAIN QUERY PLAN` against a temporary DB and assert that the intended indexes are used for representative `get_context()` and `sleep()` queries.

dplush:
## Summary

- split `BeamMemory.get_context()` into separate global/session queries so SQLite can use straightforward indexes instead of the previous `OR` + `CASE` ordering shape
- add partial context indexes for global and session lookups
- keep the public ordering contract: global memories first, then session memories; both groups sorted by importance and recency
- avoid the per-row `SELECT last_recalled` by selecting it in the initial queries and removing it from returned rows before returning

Closes #291.

## Verification

```text
env PYTHONPATH=$PWD uv run --no-project \
  --with pytest --with pytest-asyncio --with pyyaml --with numpy --with python-dotenv \
  python -m pytest tests/test_beam.py tests/test_beam_e3_additive_sleep.py tests/test_pre_experiment_fidelity.py -q

121 passed in 94.51s (0:01:34)
```

Added tests cover:

- context indexes are created
- global-first/session ordering is preserved
- `limit=-1` keeps SQLite's unbounded-limit behavior
- representative query shapes use the new context indexes without a temp B-tree

### DIFF
--- a/mnemosyne/__init__.py
+++ b/mnemosyne/__init__.py
@@ -1,3 +1,48 @@
-"""Mnemosyne — local-first, zero-cloud memory for AI agents."""
+"""
+Mnemosyne - The Zero-Dependency AI Memory System
+
+A native, sub-millisecond memory system for AI agents using SQLite.
+No HTTP, no servers, no API keys — just Python and SQLite.
+
+Example:
+    >>> from mnemosyne import remember, recall
+    >>> remember("User prefers dark mode", importance=0.9)
+    >>> results = recall("user preferences")
+"""
 
 __version__ = "3.7.0"
+__author__ = "Abdias J"
+__license__ = "MIT"
+
+# Lazy imports to allow mnemosyne.install to run without heavy deps
+# (e.g. numpy is not yet installed during first-time setup)
+_lazy_exports = {
+    "Mnemosyne": (".core.memory", "Mnemosyne"),
+    "remember": (".core.memory", "remember"),
+    "recall": (".core.memory", "recall"),
+    "get_context": (".core.memory", "get_context"),
+    "get_stats": (".core.memory", "get_stats"),
+    "get": (".core.memory", "get"),
+    "forget": (".core.memory", "forget"),
+    "update": (".core.memory", "update"),
+}
+
+
+def __getattr__(name: str):
+    if name in _lazy_exports:
+        mod_path, attr_name = _lazy_exports[name]
+        mod = __import__(f"mnemosyne{mod_path}", fromlist=[attr_name])
+        return getattr(mod, attr_name)
+    raise AttributeError(f"module 'mnemosyne' has no attribute '{name}'")
+
+
+__all__ = list(_lazy_exports.keys())
+
+# Conditionally expose MCP server if mcp package is installed
+try:
+    import mcp  # noqa: F401
+
+    _lazy_exports["run_mcp_server"] = (".mcp_server", "run_mcp_server")
+    __all__.append("run_mcp_server")
+except ImportError:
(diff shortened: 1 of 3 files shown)


AxDSan:
@dplush this looks solid. 121 tests pass locally, the CI failure on 3.13 is a pre-existing infra issue (main has the same failure on 3.11). Build + docs-check both pass. Merging now.

Appreciate the thorough analysis across all 3 issues. You found real optimization opportunities.

PR #297, opened 2026-06-13
fix: reclaim stale consolidation orphan claims

### LINKED ISSUE #293: Improve consolidation robustness: avoid permanent orphan claims in sleep()
# Improve consolidation robustness: avoid permanent orphan claims in sleep()

## Summary

`BeamMemory.sleep()` currently marks working-memory rows as consolidated before the episodic summary is written. The code comments acknowledge that a crash between the claim and the summary insert can leave an “orphan claim”: rows have `consolidated_at` set, but no corresponding episodic summary was created.

The originals remain recallable, so this is not data loss, but it can permanently skip expected consolidation for those rows.

Reviewed on `main` at commit `e7f3509`.

## Evidence

Relevant code: `mnemosyne/core/beam.py`, `sleep()` around lines `6972-7018`.

The code intentionally claims rows first:

```python
UPDATE working_memory SET consolidated_at = ?
WHERE id IN (...) AND consolidated_at IS NULL
```

The nearby comment notes that if the process dies after the claim but before episodic insert, the rows remain marked and future sleep cycles skip them.

## Why this matters

In long-running agents, container restarts, process kills, or local LLM failures can happen during consolidation. If the claim is final, the maintenance state can drift:

- `working_memory.consolidated_at` says processed
- `episodic_memory` has no summary
- future `sleep()` calls skip those rows

This can make consolidation stats misleading and reduce long-term summarization quality.

## Suggested changes

Use a two-phase claim/finalize model:

1. Add transient claim metadata:
   - `consolidation_claim_id`
   - `consolidation_claimed_at`
2. During sleep, claim rows with a unique claim id.
3. Write the episodic summary.
4. Only after the summary is written, set final `consolidated_at` and clear claim metadata.
5. On later sleep runs, reclaim stale claims older than a timeout.

A smaller alternative would be a maintenance helper/tool to find and clear stale consolidated rows that have no matching episodic summary.

## Expected benefit

- More robust consolidation across crashes/restarts.
- Clearer operational stats.
- Ability to recover from interrupted sleep runs without manual DB surgery.

## Suggested tests

- Simulate a sleep crash after claiming rows but before summary insert.
- Verify a later sleep run can reclaim and summarize those rows.
- Verify concurrent sleep still avoids duplicate summaries.

dplush:
## Summary

- add a crash-recovery marker for sleep claims: `working_memory.consolidation_claimed_at`
- add `reclaim_orphans()` to clear stale sleep claims that never produced an episodic summary
- expose the helper via `BeamMemory`, `Mnemosyne`, module-level API, and top-level `mnemosyne.reclaim_orphans`
- keep reclaim manual/explicit rather than automatically running it from `sleep_all_sessions()`
- preserve legacy E3 backfill behavior by only reclaiming rows with an explicit claim marker

Closes #293.

## Safety notes

`reclaim_orphans()` only clears rows where:

- `consolidated_at IS NOT NULL`
- `consolidation_claimed_at IS NOT NULL`
- the claim is older than `stale_after_seconds` (default: 3600)
- no `episodic_memory.summary_of` CSV token references the working-memory id

Successful `sleep()` runs now clear `consolidation_claimed_at` after the episodic summary is written, so normal consolidated rows are not candidates for reclaim.

## Verification

```text
python -m pytest tests/test_beam_e3_additive_sleep.py -q
21 passed in 16.09s
```

```text
python -m pytest \
  tests/test_beam.py \
  tests/test_beam_e3_additive_sleep.py \
  tests/test_hermes_memory_provider.py \
  tests/test_hermes_memory_provider_shared_multi_agent.py \
  tests/test_hermes_memory_provider_unified_recall.py \
  tests/test_hermes_memory_provider_validation.py \
  tests/test_mcp_server.py \
  tests/test_outer_package_version.py \
  tests/test_pre_experiment_fidelity.py \
  -q --tb=short
235 passed in 137.53s (0:02:17)
```

Also split the broader local suite into chunks. The only local failures observed were pre-existing/environmental:

- `tests/test_hermes_llm_adapter.py::test_complete_returns_none_when_agent_import_unavailable` sees this Hermes host's real `agent` package
- two `tests/test_mnemosyne_stats.py` assertions open the default home DB directly and fail when no local schema exists

Neither failure is in the reclaim path; the relevant Beam/provider/fidelity suites above pass.

### DIFF
--- a/mnemosyne/__init__.py
+++ b/mnemosyne/__init__.py
@@ -25,6 +25,7 @@ _lazy_exports = {
     "get": (".core.memory", "get"),
     "forget": (".core.memory", "forget"),
     "update": (".core.memory", "update"),
+    "reclaim_orphans": (".core.memory", "reclaim_orphans"),
 }
 
 
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -599,6 +599,12 @@ def init_beam(db_path: Path = None):
             (datetime.now().isoformat(),),
         )
 
+    try:
+        cursor.execute("ALTER TABLE working_memory ADD COLUMN consolidation_claimed_at TEXT")
+    except sqlite3.OperationalError as exc:
+        if "duplicate column" not in str(exc).lower():
+            raise
+
     # Partial index for the sleep eligibility predicate. Sleep scans
     # WHERE session_id = ? AND timestamp < ? AND consolidated_at IS NULL
     # on every cycle; once consolidated rows accumulate the predicate
@@ -608,6 +614,11 @@ def init_beam(db_path: Path = None):
         "CREATE INDEX IF NOT EXISTS idx_wm_unconsolidated "
         "ON working_memory(session_id, timestamp) "
         "WHERE consolidated_at IS NULL"
+    )
+    cursor.execute(
+        "CREATE INDEX IF NOT EXISTS idx_wm_consolidation_claims "
+        "ON working_memory(consolidation_claimed_at) "
+        "WHERE consolidation_claimed_at IS NOT NULL"
     )
 
     # --- SCRATCHPAD ---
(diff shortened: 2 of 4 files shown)


AxDSan:
@dplush this is clean. 235 local tests passed, mergeStateStatus CLEAN, the reclaim logic is tight. The explicit `reclaim_orphans()` approach keeps the sleep path simple without overcomplicating things.

Merging now.

Issue #292, opened 2026-06-13
Reduce repeated work in recall(): cache query embeddings and narrow working-memory vector scans


dplush:
# Reduce repeated work in recall(): cache query embeddings and narrow working-memory vector scans

## Summary

The current recall path appears to do repeated embedding work and broad scans over `memory_embeddings` for working-memory vector search. This can become expensive as the memory store grows.

Reviewed on `main` at commit `e7f3509`.

## Evidence

Relevant areas in `mnemosyne/core/beam.py`:

- Main `recall()` function starts around line `4509` and is currently very large.
- Query embedding appears to be computed/used in multiple places in the recall path, including around lines `4638`, `5076`, and `5085`.
- Working-memory vector search appears to scan broadly through `memory_embeddings` before later filtering.

During a local review, the working-memory vector path showed a broad scan pattern over `memory_embeddings`, which could become costly when many embeddings exist.

### 1. Compute `query_embedding` once per recall call

At the start of `recall()`, compute the query embedding once when any vector-backed path needs it, then pass it into:

- working-memory vector search
- episodic vector search
- binary/vector bonus scoring

This avoids duplicated embedding calls and makes the recall flow easier to reason about.

### 2. Filter working-memory vector candidates earlier

Allow the working-memory vector search helper to accept recall filters such as:

- `session_id`
- `scope`
- `channel_id`
- `author_id`
- `author_type`
- `source`
- `memory_type`
- `veracity`

Then join or preselect candidate memory IDs before decoding JSON vectors / doing NumPy distance work.

### 3. Longer-term: use a dedicated sqlite-vec table for working memory

Episodic memory already benefits from vector infrastructure. Working memory currently appears to rely more heavily on `memory_embeddings` JSON/vector scanning. A dedicated working-memory vector table could avoid Python-side broad scans for larger installs.

## Expected benefit

- Fewer embedding calls per recall.
- Less Python-side vector decoding and distance work.
- Better scalability when `working_memory` and `memory_embeddings` grow.
- Cleaner separation of recall phases.

## Suggested tests

- Add a test that monkeypatches the embedding function and asserts one query embedding computation per recall call.
- Add a test DB with multiple sessions/channels and assert the vector candidate set is filtered before scoring.
- Add a query-plan/benchmark smoke test for representative recall calls.

AxDSan:
[...]

PR #298, opened 2026-06-14
Cache recall query embedding per call

dplush:
## Summary

- Cache the query embedding once per `recall()` call.
- Reuse that embedding for working-memory vector search, binary-vector scoring, and episodic vector search.
- Add a regression test that asserts `embed_query()` is called only once during recall.

## Scope

This is the low-risk first PR from #292. The working-memory vector filter pushdown remains a separate follow-up.

## Tests

- `pytest tests/test_beam.py::test_recall_computes_query_embedding_once_per_call -q`
- `pytest tests/test_beam.py -q`
- `pytest tests/test_configurable_scoring.py tests/test_beam_e3_additive_sleep.py -q`
- Combined: `121 passed, 1 skipped`

Fixes part of #292.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -4625,6 +4625,27 @@ class BeamMemory:
         # ---- Configurable hybrid scoring setup (Phase 4) ----
         vw, fw, iw = _normalize_weights(vec_weight, fts_weight, importance_weight)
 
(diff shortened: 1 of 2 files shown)


AxDSan:
@dplush clean change. CI green, 49-line diff, zero regressions on the 121-test suite. Merging now.

Cache is the right first step. The wm_vec_search filter pushdown can follow when you're ready.

Issue #308, opened 2026-06-14
Feature: Phase 2 — async `reindex` with progress reporting + cancellation


Milgauss:
**Phase 1 (synchronous `mnemosyne reindex`) shipped in #311 (merged in `60a3207`).** This issue now tracks **Phase 2**.

## Phase 2 — async / monitored reindex

Phase 1 re-embeds and rebuilds all five vector representations synchronously: it blocks until done and is meant to be run offline (provider/gateway stopped). That covers the recovery case after a deliberate embedding-model swap.

Phase 2 adds an async / monitored path for when blocking isn't acceptable:

- **Progress reporting** surfaced to the caller (per-store counts / percentage), beyond the current CLI progress lines.
- **Cancellation** — safely abort a long-running reindex.
- **Reduced/eliminated downtime** — re-embed in the background (e.g. into staging tables) and swap atomically, so the provider can stay responsive during the operation.

### Notes / open questions

- Build on the Phase 1 mechanism (`reindex_vectors` reusing the existing write helpers); the main additions are job control + progress surface + staged-swap atomicity.
- Whether to expose this as a long-running CLI mode, a provider-level job, or both.

**Status:** deferred — to implement when demand justifies it (per the Phase 1 discussion). Not started.

AxDSan:
@Milgauss Solid research. Confirmed from source:

- vec0 tables at beam.py:681-687 (vec_episodes, vec_working) and beam.py:1009-1010 (vec_facts) all use IF NOT EXISTS with EMBEDDING_DIM baked in.
- EMBEDDING_DIM is computed once at import in embeddings.py:253.
- No existing reindex for vector tables (only FTS5 at beam.py:3671).

One note: there are 5 embedding stores now, not 4. vec_working was added in a recent PR as a dedicated vec0 mirror for working memory. The reindex will need to cover that too.

This is a well-scoped problem but the implementation is heavy. Re-embedding through a new model on a loaded DB could take minutes, which means the Hermes provider goes unresponsive during the operation.

Would a phased approach work for you?

Phase 1: CLI flag + manual re-embed (single-shot, synchronous, with a warning about duration). You swap the model, run the reindex, it blocks until done.
Phase 2: Async/monitored version with progress reporting and cancellation.

If Phase 1 meets your use case, I'd review that same-day. Phase 2 is a separate PR when the demand justifies it.

Milgauss:
[...]

PR #310, opened 2026-06-14
fix(beam): score fact_recall hits by query relevance, return full triple as content

### LINKED ISSUE #309: `fact_recall` ranks by stored confidence instead of query relevance, and returns the bare object as content
**Repo:** AxDSan/mnemosyne · **Version:** 3.7.0 · **Component:** `core/beam.py::fact_recall` + the recall integration in `recall()`
**Type:** bug (recall relevance) · **Feature gate:** `MNEMOSYNE_FACT_RECALL_ENABLED=1` (opt-in)

## Summary

When fact recall is enabled, fact hits are (1) scored by their stored
**confidence** rather than how well they match the **query**, and (2) surfaced
with only the fact's **object** as `content`. The FTS relevance (`rank`) is
computed and then discarded. The result: for a given query, every fact hit scores
near-identically and **outranks the relevance-scored episodic/working memories**,
while the merged content (e.g. just `"hiking"`) is uninformative. This makes the
opt-in fact-recall feature actively *reduce* recall quality.

## Repro

```python
# facts stored with the default confidence (e.g. 0.7), MNEMOSYNE_FACT_RECALL_ENABLED=1
#   e.g. (user, hobby, hiking), (user, hobby, chess)
mem.recall("what does the user do for fun?", top_k=5)
```

Observed (top_k=5): the `tier=fact` results' `content` is a single token
("hiking", "chess", …), all at `score≈0.63` (= 0.7 confidence × the integration's
0.9 discount), ranked **above** an episodic memory that directly describes the
relevant events. The episodic memory is pushed out of the top-k.

## Root cause

`core/beam.py::fact_recall()` — Source 1 (raw `facts` table):

```python
fts_rows = cursor.execute(
    "SELECT rowid, rank FROM fts_facts WHERE fts_facts MATCH ? ORDER BY rank, rowid LIMIT ?",
    (query, top_k * 3)).fetchall()
...
fact_ids = [r["rowid"] for r in fts_rows[:top_k]]
cursor.execute(f"""SELECT fact_id, subject, predicate, object, timestamp, confidence
                   FROM facts WHERE rowid IN ({placeholders})
                   ORDER BY confidence DESC LIMIT ?""", (*fact_ids, top_k))   # <-- FTS rank dropped
...
fact_text = obj if obj else f"{subject} {predicate} {obj}"                     # <-- bare object
results.append({"content": fact_text,
                "score": confidence if confidence is not None else 0.5, ...})  # <-- confidence, not relevance
```

Two issues, both **general** (independent of any application):

1. **Relevance is discarded.** `fts_facts` is queried `ORDER BY rank` (bm25), but
   the rows are then re-selected `ORDER BY confidence DESC` and scored by
   `confidence`. Since facts written by the same path share a confidence
   (`_store_facts_in_table` uses a constant 0.7), all hits for a query collapse to
   one score — query match quality never enters the ranking.
2. **Content is the bare object.** `fact_text = obj …` drops the
   subject/predicate, so a triple like `("user","hobby","hiking")` surfaces as
   `"hiking"`. (Source 2, `consolidated_facts`, already uses the full triple — the
   two sources are inconsistent.)

Downstream, `recall()` merges these with `"score": fr["score"] * 0.9` and re-sorts,
so the flat-confidence facts sit at the top of the merged list.

## Impact

- Fact hits dominate the top-k with query-independent scores, demoting genuinely
  relevant episodic/working memories.
- Merged `content` for facts is uninformative (object only).
- Affects **any** consumer that enables `MNEMOSYNE_FACT_RECALL_ENABLED`, not a
  specific application.

## Proposed fix (see PR)

In `fact_recall()` Source 1: preserve the FTS `rank` order, derive a relevance
score from the rank, combine it with confidence (`relevance * confidence`), and set
`content` to the full `subject predicate object` triple. This makes fact hits rank
by *both* query match and trust, so they integrate into the hybrid ranking instead
of flooding it, and read meaningfully when merged. No public API change; the
feature stays opt-in and off by default.

## Related / follow-up

- Source 2 (`consolidated_facts`) also scores by confidence only (`× 0.85`); it
  already uses full-triple content. A relevance-aware score there (e.g. weight by
  query/subject overlap) is a natural follow-up, kept out of the initial PR to keep
  it focused.

Milgauss:
## What

`fact_recall()` (the raw `facts` source) queries `fts_facts ORDER BY rank` (bm25
relevance) but then re-selects the rows `ORDER BY confidence DESC` and scores them by
the stored `confidence` — discarding query relevance. Because facts written by the
same path share a confidence, every fact for a query collapses to one score and
**outranks the relevance-scored episodic/working memories** in merged recall. The
merged `content` is also just the fact's object (e.g. `"hiking"`), meaningless out of
context.

General issue — affects any consumer that enables `MNEMOSYNE_FACT_RECALL_ENABLED`.

## Change

In `fact_recall()` Source 1: preserve the FTS rank order and each fact's rank
position, derive a bounded relevance from the position (top hit → ~1.0, decaying),
combine it with confidence (`relevance * confidence`), and build `content` from the
full subject-predicate-object triple. The existing `results.sort(...score)[:top_k]`
then orders by the relevance-aware score; the `recall()` integration's ×0.9 discount
still applies.

## Notes

- **Rank-position decay** (`1 - pos/n`) rather than the raw bm25 value: bounded to
  `[0, 1]` so it composes with confidence, dependency-free, and preserves the FTS
  ordering (the part that matters).
- **LIKE fallback** keeps working — those rows carry `rank=0` and are scored by
  position, a strict improvement over the flat score.
- **Source 2 (`consolidated_facts`)** left as-is here (already full-triple content;
  its confidence-only score is a smaller follow-up) to keep this focused.
- **No API/signature change**; behavior unchanged when the feature is disabled (default).

## Tests

Adds `tests/test_fact_recall_ranking.py`: two equal-confidence facts matching the same
query now return full-triple content (not the bare object) and get
relevance-differentiated scores instead of collapsing to one value.

## Risk

Low. Confined to the opt-in `fact_recall` path; no schema or API change; default-off
behavior unchanged.

Fixes #309

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -6076,34 +6076,48 @@ class BeamMemory:
                         fts_rows.append({"rowid": row["rowid"], "rank": 0})
 
         if fts_rows:
-            fact_ids = [r["rowid"] for r in fts_rows[:top_k]]
-            placeholders = ",".join("?" * len(fact_ids))
+            # Keep the FTS relevance ORDER and each fact's rank position so the
+            # score can reflect how well the fact matches the *query*, not just how
+            # confident we are it is true. (Previously the rows were re-ordered by
+            # confidence and the FTS rank was discarded, so every fact for a query
+            # scored at its stored confidence and outranked relevance-scored
+            # memories in the merged result.)
+            ranked_ids = [r["rowid"] for r in fts_rows[:top_k]]
+            rank_pos = {rid: i for i, rid in enumerate(ranked_ids)}
+            placeholders = ",".join("?" * len(ranked_ids))
             try:
                 cursor.execute(f"""
-                    SELECT fact_id, subject, predicate, object,
+                    SELECT rowid, fact_id, subject, predicate, object,
                            timestamp, confidence
                     FROM facts
                     WHERE rowid IN ({placeholders})
-                    ORDER BY confidence DESC
-                    LIMIT ?
-                """, (*fact_ids, top_k))
+                """, (*ranked_ids,))
                 fact_rows = cursor.fetchall()
             except Exception:
                 fact_rows = []
 
+            n = max(len(ranked_ids), 1)
             for raw_row in fact_rows:
                 row = dict(raw_row)
                 confidence = row.get("confidence")
-                subject = row.get("subject")
-                predicate = row.get("predicate")
-                obj = row.get("object")
-                fact_text = obj if obj else f"{subject} {predicate} {obj}"
+                confidence = confidence if confidence is not None else 0.5
+                subject = row.get("subject") or ""
+                predicate = row.get("predicate") or ""
+                obj = row.get("object") or ""
+                # Full subject-predicate-object triple: a bare object is
+                # meaningless once merged into general recall output.
+                fact_text = " ".join(p for p in (subject, predicate, obj) if p).strip() or obj
+                # Relevance from the FTS rank position (top hit -> ~1.0, decaying),
+                # combined with confidence. Replaces the flat per-fact confidence
+                # that made every hit score identically.
+                pos = rank_pos.get(row["rowid"], n - 1)
+                relevance = 1.0 - (pos / n)
                 results.append({
                     "content": fact_text,
-                    "score": confidence if confidence is not None else 0.5,
+                    "score": relevance * confidence,
                     "fact_id": row["fact_id"],
-                    "subject": subject if subject is not None else "",
-                    "predicate": predicate if predicate is not None else "",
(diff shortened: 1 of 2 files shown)


AxDSan:
@Milgauss Solid RCA and a clean fix. Confirmed the outdated code at beam.py:6568-6596 from source. The rank-position approach is the right simplicity for this scope. CI green, merging now.

Also, #308 (reindex) is a well-researched proposal. The vec0 hard-failure path is the right place to start. Happy to review when you draft. Same-day turn as this one.

PR #311, opened 2026-06-14
feat(reindex): Phase 1 — synchronous `mnemosyne reindex` to rebuild vectors after a model/dim change

Milgauss:
Phase 1 of #308 — a synchronous `mnemosyne reindex` that rebuilds every vector representation from source text with the active embedding model. Opened to save you a step; happy to adjust per your review.

## What it does

`reindex_vectors()` (in `core/beam.py`, alongside `repair_vec_working`) re-embeds `working_memory` and `episodic_memory` with the current model and recreates the sqlite-vec tables at the active dimension. It reuses the existing write helpers so encodings stay identical to the normal store path:

- **working:** `memory_embeddings` (float JSON) + `vec_working` — via `_store_working_embedding`
- **episodic:** `episodic_memory.binary_vector` + `vec_episodes` — via `_mib` + `_vec_table_insert`
- **vec_facts:** recreated empty (no writer yet) so its declared dim can't mismatch a query

Good catch on `vec_working` — that's covered (this is the full 5-store set). Exposed as `Mnemosyne.reindex_vectors(...)` and a CLI command:

```
mnemosyne reindex [--model NAME] [--dry-run] [--yes] [--no-backup]
```

Synchronous and blocking, with an up-front "may take several minutes — run with the gateway stopped" warning, auto-backup first (via `dr.recovery.create_backup`), and a per-store progress line. `--dry-run` prints the plan (model, dim, per-store counts) and writes nothing.

## Your design questions

- **Shape / auto-backup / dry-run:** as above — `--model`/`--dry-run`/`--yes`/`--no-backup`, backup-first by default.
- **Atomicity:** currently backup-first + per-batch commits — a crash mid-run is recoverable from the backup, and re-running is idempotent (drops/recreates the vec tables, overwrites the float/binary stores). Glad to switch to staging into temp tables + atomic swap if you'd rather not lean on the backup for Phase 1.
- **Idempotency:** yes — re-running with the same model re-embeds to the same vectors.
- **diagnose/verify dim-mismatch check:** left out here to keep this focused; happy to add a follow-up that flags *configured dim ≠ stored vec-table dim* and points at `reindex`.

## `--model` note

`--model` sets `MNEMOSYNE_EMBEDDING_MODEL` before the embedding module is imported, so `EMBEDDING_DIM` (and the binary packer) pick up the new dimension without a reload. Documented flow is still: swap the model, then `reindex`.

## Tests

`tests/test_reindex_vectors.py` simulates the bug (vec0 tables recreated at a wrong dim), then asserts reindex recreates them at the active dim, repopulates working + episodic vectors, refreshes `binary_vector`, and leaves recall working; plus a `--dry-run` writes-nothing check.

## Scope

Phase 1 only (synchronous). Phase 2 (async + progress/cancellation) is a separate PR per your note.

Refs #308

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -585,6 +585,81 @@ def cmd_bank(args):
         _fail(str(e))
 
 
+def cmd_reindex(args):
+    """Rebuild vector indexes from source text with the active embedding model.
+
+    Usage: mnemosyne reindex [--model NAME] [--dry-run] [--yes] [--no-backup]
+
+    Use after changing the embedding model/dimension. Synchronous and blocking —
+    re-embeds working + episodic memory, so it can take minutes on a large DB.
+    Run it with any provider/gateway stopped.
+    """
+    dry_run = "--dry-run" in args
+    assume_yes = "--yes" in args or "-y" in args
+    no_backup = "--no-backup" in args
+
+    # --model has to win before the embedding module is imported: it freezes the
+    # model + dimension from the env at import time.
+    if "--model" in args:
+        try:
+            os.environ["MNEMOSYNE_EMBEDDING_MODEL"] = args[args.index("--model") + 1]
+        except IndexError:
+            _usage("Usage: mnemosyne reindex [--model NAME] [--dry-run] [--yes] [--no-backup]")
+
+    from mnemosyne.core import embeddings as _emb
+
+    mem = _get_memory()
+
+    if dry_run:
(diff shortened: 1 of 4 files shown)


AxDSan:
@Milgauss Merged in 60a3207. Great PR — clean scope, proper test coverage, and the backup-first + idempotent design makes this safe to ship.

The 5-store coverage (including vec_working) and the per-store progress lines were smart touches. Phase 2 can wait until demand justifies it.

Solid work.

PR #313, opened 2026-06-15
fix: use k=? constraint instead of LIMIT for sqlite-vec 0.1.9+

ether-btc:
## Problem

sqlite-vec 0.1.9+ removed support for the `LIMIT N` clause when querying vec0 KNN tables. The 7 KNN call sites in `mnemosyne/core/beam.py` and `mnemosyne/core/polyphonic_recall.py` were silently returning empty result sets because the `LIMIT` clause was being rejected.

## Fix

Replace `WHERE ... LIMIT N` with `WHERE k=N` (the documented vec0 KNN constraint) on all 7 call sites. `k` is the vec0-native KNN cap; `LIMIT` is not supported for vec0 KNN.

## Test plan

- All `tests/test_beam.py` and `tests/test_beam_e5_polyphonic_recall.py` tests pass locally
- Manual verification: `mnemosyne_remember` and `mnemosyne_recall` work end-to-end after the fix
- Live database: `vec_episodes` has 422 rows and is queryable via `mnemosyne_recall`

## Affected files

- `mnemosyne/core/beam.py` (5 call sites)
- `mnemosyne/core/polyphonic_recall.py` (2 call sites)

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1926,7 +1926,7 @@ def _vec_search(conn: sqlite3.Connection, embedding: List[float], k: int = 20) -
     k = int(k)
     if vec_type == "bit":
         rows = conn.execute(
-            f"SELECT rowid, distance FROM vec_episodes WHERE embedding MATCH vec_quantize_binary(?) ORDER BY distance LIMIT {k}",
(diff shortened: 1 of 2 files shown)


ether-btc:
Superseded by #315 (combined fix).

PR #314, opened 2026-06-15
fix: make _effective_vec_type table-aware for mixed vec0 schemas

ether-btc:
## Problem

`_effective_vec_type()` in `mnemosyne/core/beam.py` only read the schema of `vec_episodes` and assumed all vec0 tables (`vec_episodes`, `vec_working`, `vec_facts`) shared the same vector type. On databases created or migrated under different `MNEMOSYNE_VEC_TYPE` settings, this caused:

- `_vec_table_insert()`: float32 INSERT into int8 table -> silent failure (0 rows)
- `_wm_vec_search_sqlite()`: float32 MATCH against int8 table -> wrong results

In our production DB, `vec_episodes` is float32[384] but `vec_working` is int8[384], leaving `vec_working` permanently empty (0 rows) despite 1000+ working memory records.

## Fix

1. `_effective_vec_type()` now takes a `table` parameter (defaults to `vec_episodes` for backward compatibility)
2. `_vec_table_insert()` passes the actual target table name
3. `_wm_vec_search_sqlite()` passes `"vec_working"` explicitly
4. SQL query is parameterized to prevent injection on the table name

## Test plan

- `test_beam.py` KNN/vec tests all pass (10/10)
- `test_beam_e5_polyphonic_recall.py` passes (17/17)
- Real DB backfill: `vec_working` went from 0 rows to 75 rows after the fix
- `mnemosyne_remember` + `mnemosyne_recall` work end-to-end on the patched code

## Affected files

- `mnemosyne/core/beam.py` (3 sites)

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1758,13 +1758,21 @@ def _in_memory_vec_search(conn: sqlite3.Connection, query_embedding: np.ndarray,
     return results[:k]
 
 
-def _effective_vec_type(conn: sqlite3.Connection) -> str:
-    """Re-detect the actual vector type used by vec_episodes."""
+def _effective_vec_type(conn: sqlite3.Connection, table: str = "vec_episodes") -> str:
+    """Re-detect the actual vector type used by a vec0 table.
+
+    Each vec0 virtual table (vec_episodes, vec_working, vec_facts) can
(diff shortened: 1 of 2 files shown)


ether-btc:
Superseded by #315 (combined fix).

PR #320, opened 2026-06-15
fix(cli): respect HERMES_HOME for data directory

dplush:
## Summary

This makes the standalone CLI use the same default Mnemosyne data directory as the Hermes integration.

Before this, `mnemosyne stats` ignored `HERMES_HOME` unless `MNEMOSYNE_DATA_DIR` was also set. In a Hermes install with `HERMES_HOME=/opt/data`, that meant the CLI could read an empty `~/.hermes/mnemosyne/data` database while the active provider was using `/opt/data/mnemosyne/data`.

The new order is:

1. `MNEMOSYNE_DATA_DIR`
2. `HERMES_HOME/mnemosyne/data`
3. `~/.hermes/mnemosyne/data`

I applied the same fallback to `scripts/mnemosyne-stats.py`, since it had the same default-path behavior.

## Testing

```bash
uv run --no-project \
  --with pytest --with pyyaml --with numpy --with sqlite-vec --with fastembed \
  --with python-dotenv --with cryptography --with requests \
  python -m pytest -q \
  tests/test_cli_data_dir_resolution.py \
  tests/test_cli_stats.py \
  tests/test_cli_errors.py \
  tests/test_cli_operation_failures.py \
  tests/test_cli_usage_errors.py \
  tests/test_mnemosyne_stats.py
```

Result:

```text
58 passed in 79.27s
```

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -12,10 +12,21 @@ import json
 from pathlib import Path
 from typing import NoReturn
 
-# Data directory - respects MNEMOSYNE_DATA_DIR env var
-DATA_DIR = os.environ.get("MNEMOSYNE_DATA_DIR") or str(
-    Path.home() / ".hermes" / "mnemosyne" / "data"
-)
+def _default_data_dir() -> str:
+    """Resolve the default data directory used by the CLI.
+
+    Keep the standalone CLI aligned with Hermes integrations:
+    MNEMOSYNE_DATA_DIR wins, then HERMES_HOME/mnemosyne/data, then the
+    historical ~/.hermes/mnemosyne/data fallback.
+    """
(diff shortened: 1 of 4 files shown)


dplush:
Small bit of context from where I hit this: this showed up in a Hermes Docker setup where `HERMES_HOME=/opt/data`.

The active Hermes provider was using `/opt/data/mnemosyne/data`, but running the standalone CLI without `MNEMOSYNE_DATA_DIR` looked under `~/.hermes/mnemosyne/data` instead. So the CLI reported an empty/different DB even though Hermes itself was using Mnemosyne correctly.

That made this feel less like a local-only quirk and more like a small default-path mismatch between the standalone CLI and the Hermes integration.

PR #338, opened 2026-06-16
fix(hermes): add reflection cron and budget guardrails

### LINKED ISSUE #337: Add reflection budget and cron guardrails
## Problem

Automatic reflection / consolidation can be useful in interactive sessions, but it is risky for scheduled jobs and other high-frequency automation.

Cron-style workflows often produce repetitive, low-value context: daily briefs, watchdog output, status pings, market scans, etc. If those runs trigger expensive reflection calls or write noisy memories, users pay extra and the memory store slowly drifts toward summaries of routine automation.

## Proposal

Add an explicit reflection budget / guardrail layer.

Example config:

```yaml
reflect:
  max_calls_per_session: 3
  max_calls_per_day: 20
  disabled_for_cron: true
  min_turns_before_reflect: 5
  cost_budget_usd: 0.05
```

Or under the Hermes integration namespace if that is a better fit:

```yaml
memory:
  mnemosyne:
    reflect:
      disabled_for_cron: true
      max_calls_per_session: 3
```

When reflection is skipped, return a structured reason instead of silently doing nothing:

```json
{
  "status": "skipped",
  "reason": "reflect_disabled_for_cron"
}
```

Other useful skip reasons:

- `reflect_budget_exhausted`
- `min_turns_not_reached`
- `low_value_automation_context`
- `cost_budget_exhausted`

## Why this helps

This keeps memory useful and predictable. Interactive sessions can still benefit from reflection, while cron/watchdog/briefing jobs stay cheap and do not flood memory with routine output.

It also gives operators a clear knob for production setups: memory quality first, cost second, no surprise background reflection.

dplush:
## Summary

Implements the v1 reflection guardrails discussed in #337:

- add `reflect.disabled_for_cron` (default `true`)
- add `reflect.max_calls_per_session` (default `3`, negative/env parsed as unlimited)
- return structured skip payloads for:
  - `reflect_disabled_for_cron`
  - `reflect_budget_exhausted`
- apply the guardrails to explicit `mnemosyne_sleep`, auto-sleep, and session-end sleep
- keep `cost_budget_usd` out of scope for v2, per maintainer guidance

The change is mirrored in both Hermes provider entry points:

- `hermes_memory_provider/__init__.py`
- `integrations/hermes/src/mnemosyne_hermes/__init__.py`

## Tests

```bash
uv run --with pytest pytest tests/test_reflection_guardrails.py tests/test_hermes_memory_provider_audit.py -q
python3 -m py_compile hermes_memory_provider/__init__.py integrations/hermes/src/mnemosyne_hermes/__init__.py tests/test_reflection_guardrails.py
```

Result:

```text
18 passed
```

Closes #337

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -951,6 +951,43 @@ def _parse_env_float(key: str, default: float) -> float:
         return default
 
 
+def _coerce_bool(value: Any, default: bool) -> bool:
+    """Coerce config/env values to bool while preserving a safe default."""
+    if value is None:
+        return default
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)):
+        return bool(value)
+    raw = str(value).strip().lower()
+    if raw in ("1", "true", "yes", "on"):
+        return True
+    if raw in ("0", "false", "no", "off"):
+        return False
+    return default
+
+
+def _parse_env_bool(key: str, default: bool) -> bool:
+    """Read a boolean env var, falling back to default on missing/invalid values."""
+    return _coerce_bool(os.environ.get(key), default)
+
+
+def _coerce_optional_int(value: Any, default: Optional[int]) -> Optional[int]:
+    """Coerce config/env values to a non-negative int; negative means unlimited."""
+    if value is None:
+        return default
+    try:
+        parsed = int(value)
+    except (TypeError, ValueError):
+        return default
+    return parsed if parsed >= 0 else None
+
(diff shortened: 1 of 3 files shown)


AxDSan:
Good stuff @dplush. Spec-faithful implementation of all three guardrails. Conservative defaults are the right call, cron does not need reflection by default and a hard cap is exactly the kind of guardrail that prevents runaway sessions. Clean fix, #337 closed with this merge.

PR #340, opened 2026-06-16
fix(export): include canonical_facts in export/import (fixes #339)

### LINKED ISSUE #339: export_to_file silently omits canonical_facts — JSON restore is lossy
## Summary

`export_to_file()` is documented and used as the portable backup / cross-machine restore source, but it does not export several persisted surfaces. The most serious is **`canonical_facts`** — the owner-scoped, single-source-of-truth store (authored identity facts *with history*). A `CanonicalStore` row is **not** written to the JSON and therefore **silently disappears on a restore from that JSON**, with no warning.

`canonical_facts` is *authored* data (not derived/reproducible), so its loss is unrecoverable from the export. This is a correctness bug, not just a completeness gap.

## What is and isn't exported today

Exported: `working_memory`, `episodic_memory` (+ `episodic_embeddings`), `scratchpad`, `consolidation_log`, `legacy_memories` (+ `legacy_embeddings`), `triples`, `annotations`, and optional `sync_events`.

**Not exported:**

| Surface | Authored or derived? | Consequence of omission |
|---------|----------------------|--------------------------|
| `canonical_facts` | **authored** (single-source-of-truth, with history) | **Unrecoverable** — silently lost on JSON restore |
| `graph_edges` | derived (episodic graph) | Lost until graph is rebuilt |
| `gists` | derived (episodic graph) | Lost until rebuilt |
| `consolidated_facts` | derived (veracity consolidation) | Lost until reconsolidation |

The derived surfaces can in principle be rebuilt by re-running the graph/consolidation passes, but the export gives no signal that a restore is incomplete, and a rebuild is not guaranteed to reproduce them identically.

## Reproduction (synthetic)

```python
from mnemosyne.core.memory import Mnemosyne
from mnemosyne.core.canonical import CanonicalStore

mem = Mnemosyne(db_path="bank.db")
CanonicalStore(db_path="bank.db").remember(
    "owner-1", "identity", "display_name", "Goes by Sam."
)

mem.export_to_file("backup.json")

import json
data = json.load(open("backup.json"))
print("canonical_facts" in data)            # -> False

# Round-trip into a fresh DB:
Mnemosyne(db_path="restored.db").import_from_file("backup.json")
print(CanonicalStore(db_path="restored.db").recall("owner-1", "identity", "display_name"))
# -> None   (the authored fact is gone, no error raised)
```

## Expected

- `export_to_file()` includes a `canonical_facts` section and `import_from_file()` restores it, so a round-trip is **lossless for all authored surfaces**.
- The derived surfaces (`graph_edges`, `gists`, `consolidated_facts`) are either exported too, or the export documents that a restore must re-run the graph / consolidation passes to rebuild them — so "restore" is never silently partial.

## Notes

`CanonicalStore` already implements `export_all()` / `import_all()` with the same idempotency semantics as `TripleStore` / `AnnotationStore`; they were simply never wired into the file export/import. Wiring them in is small and low-risk. A PR follows that scopes the concrete fix to `canonical_facts` and leaves the derived surfaces as a tracked follow-up (they need a design choice: export vs. rebuild-on-import, plus new `export_all` methods on the episodic-graph / consolidation stores).

Milgauss:
Fixes #339.

## Why

`export_to_file()` is the portable backup / restore source, but it never writes the **`canonical_facts`** store — the owner-scoped single-source-of-truth identity facts (with history). Those rows are *authored*, not derived, so a restore from the JSON **silently loses them** with no error.

`CanonicalStore` already exposes `export_all()` / `import_all()` with the same idempotency contract as `TripleStore` and `AnnotationStore` (both already wired into the file export). This change wires canonical facts in the same way, so a round-trip is lossless for all authored surfaces. Additive section; mirrors existing code.

## Change

- `export_to_file`: add a `canonical_facts` section alongside triples/annotations; bump schema `1.2` → `1.3` (canonical facts are always present); report `canonical_facts_count`.
- `import_from_file`: accept version `1.3` and restore the section via `CanonicalStore.import_all`.
- New regression test `tests/test_export_canonical.py`.

## Compatibility

- **Old export → new import:** 1.0/1.1/1.2 files lack `canonical_facts`; `data.get("canonical_facts", [])` makes the import a no-op. Unchanged behavior.
- **New export → old import:** a pre-patch importer rejects `version == "1.3"` (`Unsupported export version`). This matches how 1.1 and 1.2 were introduced — import with a matching-or-newer build. Called out so it's a deliberate bump, not a surprise.

## Tests

`tests/test_export_canonical.py` (all passing locally), covering:

1. **Round-trip** — a canonical fact survives export → import into a fresh DB.
2. **History survives** — superseded values are preserved across the round-trip.
3. **Idempotency** — importing the same export twice creates no duplicate live row.
4. **Backward compat** — a synthetic 1.2 payload with no `canonical_facts` key imports without error as a no-op.

The existing export test (`tests/test_e6a_followup_gaps.py`) still passes.

## Out of scope (follow-up)

`graph_edges`, `gists`, and `consolidated_facts` are *derived* surfaces owned by the episodic-graph and veracity-consolidation modules, which don't yet expose `export_all`/`import_all`. Covering them needs a design decision (export the derived rows vs. rebuild them on import by re-running the passes) plus new store methods, so they're left to a separate PR to keep this one focused and reviewable.

### DIFF
--- a/mnemosyne/core/memory.py
+++ b/mnemosyne/core/memory.py
@@ -605,19 +605,22 @@ class Mnemosyne:
     ) -> Dict:
         """
         Export all Mnemosyne data (legacy + BEAM + triples + annotations +
-        optional sync events) to a JSON file. Returns export metadata.
+        canonical facts + optional sync events) to a JSON file. Returns export
+        metadata.
 
-        Schema version 1.2 (post-sync) adds an optional ``sync_events``
-        section alongside memory data.  Previous versions (1.0, 1.1) are
-        still importable.
+        Schema version 1.3 adds the always-present ``canonical_facts`` section.
+        1.2 (post-sync) adds an optional ``sync_events`` section. Previous
+        versions (1.0, 1.1, 1.2) are still importable; ``sync_events`` presence
+        is keyed off the section itself on import, not the version number.
         """
         from mnemosyne.core.triples import TripleStore
         from mnemosyne.core.annotations import AnnotationStore
+        from mnemosyne.core.canonical import CanonicalStore
         import json as _json
 
         # Build export metadata with device_id when available
         meta = {
-            "version": "1.2" if include_sync_events else "1.1",
+            "version": "1.3",
             "export_date": datetime.now().isoformat(),
             "source_db": str(self.db_path),
         }
@@ -674,6 +677,14 @@ class Mnemosyne:
         annotations = AnnotationStore(db_path=self.db_path)
         export["annotations"] = annotations.export_all()
 
+        # Canonical facts: owner-scoped single-source-of-truth identity, with
+        # history. These are AUTHORED (not derived), so omitting them made a
+        # JSON restore silently lossy. CanonicalStore already exposes
+        # export_all/import_all (same contract as triples/annotations); they
+        # were simply never wired into the file export.
+        canonical = CanonicalStore(db_path=self.db_path)
+        export["canonical_facts"] = canonical.export_all()
+
         # Sync events (optional, schema 1.2)
         if include_sync_events:
             try:
@@ -701,6 +712,7 @@ class Mnemosyne:
             "legacy_memories_count": len(export["legacy_memories"]),
             "triples_count": len(export["triples"]),
             "annotations_count": len(export["annotations"]),
+            "canonical_facts_count": len(export["canonical_facts"]),
             "sync_events_count": len(export.get("sync_events", [])),
         }
 
@@ -711,12 +723,14 @@ class Mnemosyne:
         Set force=True to overwrite.
         Returns import statistics.
 
-        Accepts schema versions 1.0 (pre-E6), 1.1 (post-E6), and 1.2
-        (post-sync).  When a 1.2 export includes ``sync_events``, those
-        events are imported with idempotency based on ``event_hash``.
+        Accepts schema versions 1.0 (pre-E6), 1.1 (post-E6), 1.2 (post-sync),
+        and 1.3 (canonical_facts).  When an export includes ``sync_events``,
+        those events are imported with idempotency based on ``event_hash``.
+        Sections absent from older exports are treated as no-ops.
         """
         from mnemosyne.core.triples import TripleStore
         from mnemosyne.core.annotations import AnnotationStore
+        from mnemosyne.core.canonical import CanonicalStore
         import json as _json
 
         with open(input_path, "r", encoding="utf-8") as f:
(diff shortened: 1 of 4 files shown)


Milgauss:
⚠️ **Heads-up for maintainers: this PR also fixes a control-flow bug that currently lives on `main`.**

While getting this branch green I traced a `test_sync` failure (`test_push_routes_through_remember_pipeline` and `test_push_deduplicates_by_hash`, both `assert accepted == 2` instead of `1`) to commit `12b85b8` ("fix(sync): progress indicator + graceful Ctrl+C handling"), **not** to anything in this PR.

**The bug:** `12b85b8` wrapped the `remember()` call in `push_changes()` with `try/except KeyboardInterrupt`, but the pre-existing DeltaSync-fallback `else:` ended up attached to the `try` instead of the `if self._beam … has remember` branch — i.e. a `try/except/else`:

```python
try:
    if self._beam is not None and hasattr(self._beam, "remember"):
        self._beam.remember(...)
        stats["accepted"] += 1
except KeyboardInterrupt:
    stats["interrupted"] = True
    break
else:                      # <-- runs whenever NO exception fired
    # DeltaSync fallback ... stats["accepted"] += 1
```

A `try`'s `else` executes whenever the body raises no exception, so on **every successful CREATE** the fallback also ran — double-counting `accepted` and **double-writing the memory through DeltaSync**. It's a real correctness bug, not just a test artifact.

**Why CI looked intermittent:** `pull_request` runs test *head merged with base*. Before `12b85b8` landed on `main` the tests passed; after it landed, every PR whose merge base includes `12b85b8` fails these two tests. `main`'s own push run for `12b85b8` is also marked failed — **`main` is currently red from this independently of this PR.**

**Fix (included here):** reattach the fallback `else` to the `if/else` *inside* the `try`, so it only runs when no `remember()` is available. `KeyboardInterrupt` handling is preserved. Full suite is green locally (1626 passed) and all `test_sync` tests pass.

Recommend cherry-picking this `sync.py` fix straight onto `main` so `main` goes green without waiting on this PR.

dplush:
Thanks @Milgauss for catching this. I opened a minimal PR with just the sync fix (#344) so @AxDSan  can merge it first and get main green again. Then we can continue with #340 cleanly.

Milgauss:
Confirmed — #344 is byte-identical to the fix in this branch's merge commit (`07e586a`) and is green across all four Python versions. 👍 Once it's on `main` I'll re-sync this branch (no conflict expected since the `sync.py` change matches) and we can proceed with #340. Thanks for the quick turnaround.

AxDSan:
Good stuff @Milgauss. Tiny diff, big impact. canonical_facts export was the silent loss nobody noticed until you flagged it. Clean fix, #339 closed with this merge.

PR #345, opened 2026-06-17
fix(recall): admit inflected Cyrillic forms past the lexical relevance gate

### LINKED ISSUE #341: Russian recall: lexical relevance gate discards Cyrillic fallback candidates for inflected forms
## Summary

Russian/Cyrillic recall fails for short queries (1-2 tokens) with inflected word forms, even though the Cyrillic FTS5 fallback (`_cyrillic_like_search`) correctly identifies the candidates. The lexical relevance gate (`_lexical_relevance`) in the scoring pipeline discards the candidates because it does exact/substring token matching without Russian stemming, returning 0.0 for different case endings (тёмная vs тёмную, резервная vs резервное).

The existing `_cyrillic_score` (trigram Jaccard) already solves this problem at the candidate-ranking level, but its score is not used as the relevance gate for the final scoring pass.

## Reproduction

```python
from mnemosyne import remember, recall
from mnemosyne.core.beam import _lexical_relevance, _cyrillic_score, _minimum_recall_relevance, _recall_tokens

remember("Пользователь предпочитает тёмную тему интерфейса", importance=0.9)

q = "тёмная тема"
content = "Пользователь предпочитает тёмную тему интерфейса"
tokens = _recall_tokens(q)

print(_minimum_recall_relevance(tokens))   # 0.15 (threshold for 2-token queries)
print(_lexical_relevance(tokens, content))  # 0.0  ← gate blocks this
print(_cyrillic_score(q, content))          # 0.333 ← would pass

recall("тёмная тема")   # → 0 results (missed)
recall("тёмную тему")   # → 1 result, score=0.630 (exact form works)
```

## Expected behavior

`recall("тёмная тема")` should find `"Пользователь предпочитает тёмную тему интерфейса"` — the words are different case forms of the same lemma, and `_cyrillic_score` correctly rates them as similar (0.333).

## Actual behavior

Returns 0 results. The candidate is found by `_cyrillic_like_search` (LIKE `%тёмн%` matches `тёмную`), but the scoring gate at line 5289 of `beam.py` filters it out because `_lexical_relevance` returns 0.0.

## Root cause

The scoring pipeline calls `_lexical_relevance` for all working-memory candidates (line 5287):

```python
relevance = _lexical_relevance(query_words, row["content"], query_lower)
row_min_relevance = single_token_relevance if broad_multi_hit_query else min_relevance
if relevance >= row_min_relevance or (wm_ranks and len(query_words) <= 1 and relevance > 0):
    # ... score and append
```

`_lexical_relevance` does exact token matching and substring matching for tokens ≥ 4 chars. Russian inflection changes word endings significantly (тёмная/тёмную/тёмной, резервная/резервное), so neither exact nor substring matching works.

The code already has the right tool: `_cyrillic_score` (trigram Jaccard, line 2313) and `_has_cyrillic` (line 2293). These are used in `_cyrillic_like_search` for candidate generation but not in the final relevance gate.

## Suggested fix

When `_has_cyrillic(query)` is true, use `_cyrillic_score` as a fallback or replacement for `_lexical_relevance` in the scoring gate:

```python
if _has_cyrillic(query_lower):
    relevance = max(
        _lexical_relevance(query_words, row["content"], query_lower),
        _cyrillic_score(query_lower, row["content"])
    )
else:
    relevance = _lexical_relevance(query_words, row["content"], query_lower)
```

This mirrors the existing pattern where `_cyrillic_like_search` already uses `_cyrillic_score` for candidate ranking.

## Test cases

| Query | Stored content | Current | Expected |
|-------|---------------|---------|----------|
| `тёмная тема` | `...тёмную тему...` | 0 results | 1 result |
| `резервная копия Яндекс` | `Резервное копирование...` | 0 results | 1 result |
| `тёмную тему` (exact) | `...тёмную тему...` | 1 result ✓ | 1 result |
| `настройки интерфейса` (semantic) | `...тёмную тему интерфейса` | 1 result ✓ | 1 result |

## Environment

- Mnemosyne: 3.8.0 (pipx)
- Embedding model: `sentence-transformers/paraphrase-multilingual-mpnet-base-v2` (768-dim)
- Python: 3.14.4
- OS: Linux (Ubuntu)

AxDSan:
Fixes #341.

## Problem

The lexical relevance gate in `_lexical_relevance` returned 0.0 for short inflected Cyrillic queries (тёмная vs тёмную, резервная vs резервное) because exact token matching cannot span Russian case endings.

The scoring pipeline already generated the candidate via `_cyrillic_like_search` (LIKE + trigram Jaccard), but the gate then dropped the candidate, so `recall()` returned 0 results for the very queries Cyrillic users actually type.

## Fix

Add a Cyrillic fallback inside `_lexical_relevance`, mirroring the existing CJK fallback that already lives at the bottom of the function. When the query contains Cyrillic and the exact-token path returns 0, use `_cyrillic_score` (trigram Jaccard) so the relevance gate stays consistent with `_cyrillic_like_search`.

The change is contained: one `elif` branch inside the existing fallback block. All 5 call sites of `_lexical_relevance` benefit (working memory gate, episodic memory gate, and the MEMORIA relevance path).

## Reproduction before the fix

```python
recall("тёмная тема")   # → 0 results (missed)
recall("тёмную тему")   # → 1 result
```

## After

```python
recall("тёмная тема")   # → 1 result, _cyrillic_score = 0.333, passes 0.15 threshold
recall("тёмную тему")   # → 1 result, exact match, score = 1.0
```

## Tests

Added a new `TestLexicalRelevanceCyrillicGate` class in `tests/test_cyrillic_fts.py`:

- inflected query passes the gate
- exact form still returns full score (no dilution)
- English query unaffected
- Latin query against Cyrillic content returns zero (no false positives)
- second-declension inflection (резервная копия vs Резервное копирование) passes

All 32 tests in `test_cyrillic_fts.py` pass, plus 99 recall precision regression tests.

## Credit

Diagnosis and suggested fix path from @rvboris in #341. I picked this up because the issue was filed without a follow-up PR after the offer, and the fix is contained enough to ship fast. rvboris is credited here for the root cause analysis and the suggested gate shape.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1694,6 +1694,13 @@ def _lexical_relevance(query_tokens: List[str], content: str, query_lower: str =
                 or "\uac00" <= ch <= "\ud7af"
             }
             score = len(query_cjk & content_cjk) / len(query_cjk)
+        elif _has_cyrillic(query_lower):
+            # Cyrillic inflected forms (тёмная/тёмную, резервная/резервное)
+            # defeat exact token matching. _cyrillic_score uses trigram
+            # Jaccard, which handles inflection correctly. The scoring
+            # pipeline already uses _cyrillic_like_search for candidate
+            # generation; this keeps the relevance gate consistent.
+            score = _cyrillic_score(query_lower, content_lower)
 
     return min(score, 1.0)
 
--- a/tests/test_cyrillic_fts.py
+++ b/tests/test_cyrillic_fts.py
@@ -38,6 +38,8 @@ from mnemosyne.core.beam import (
     _cyrillic_like_search,
     _fts_search,
     _fts_search_working,
+    _lexical_relevance,
+    _recall_tokens,
 )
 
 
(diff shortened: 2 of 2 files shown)


AxDSan:
Good stuff @rvboris. Your root cause analysis was right on the money: trigram Jaccard via _cyrillic_score was the correct signal, the gate just was not wired to use it. Closed the loop in #341 with this merge.

The fix is contained: one elif branch in the existing _lexical_relevance fallback block, mirroring the CJK path that already lived there. All 5 call sites of _lexical_relevance benefit from the same change. The new TestLexicalRelevanceCyrillicGate class pins the contract: inflected forms pass the gate, exact forms still return full score, English stays unaffected, and Latin-vs-Cyrillic returns zero (no false positives).

If you want to keep going on Cyrillic recall, the natural follow-up is a real Russian stemmer (Snowball has one) so we can stop relying on trigram Jaccard for short queries. Not blocking anything.

PR #346, opened 2026-06-17
fix(hermes): tell agent to read Mnemosyne context before retrieval

### LINKED ISSUE #321: [Hermes Provider] Agent bypasses injected memory context — reaches for session_search before reading Mnemosyne block
### Problem

When Mnemosyne injects context into the Hermes agent system prompt, the agent still habitually reaches for `session_search()` before reading what Mnemosyne already provided. The injected context block is present and sufficient — but the agent defaults to retrieval tools out of habit.

**Observed behavior:**
1. User asks a question Mnemosyne context can answer
2. Agent fires `session_search()` → unnecessary tool call
3. Agent acknowledges the answer was already in the `## Mnemosyne Context` block

**Expected behavior:**
1. Agent reads the injected Mnemosyne context block first
2. If sufficient → answers directly
3. If insufficient → falls back to retrieval tools

### Why this matters

Memory providers are only useful if agents actually use them. If agents bypass injected context by default, it undermines trust in the provider and wastes tokens/latency on redundant retrieval.

### What has already been fixed

- **Hermes #12081** — Synchronous prefetch was implemented (COMPLETED May 10). The context IS being injected in time for the current turn.
- The technical plumbing works. This is a **behavioral** issue on the agent side.

### What could Mnemosyne do?

Possible mitigations from the Mnemosyne side:

1. **Post-injection hint**: Append a short instruction to the injected block (e.g., `Note: answer from above memory context when possible; only use session_search if insufficient.`)
2. **Priority signal**: Tag the injected context with a priority/urgency marker that the agent system prompt can reference
3. **Documentation**: Explicitly call out this pattern in the Hermes integration guide so users know to add a behavioral instruction

This is a cross-cutting concern between Hermes (system prompt) and Mnemosyne (context format) — filing here so both sides can address it.

### Related

- Hermes issue filed: https://github.com/NousResearch/hermes-agent/issues/46955
- Reddit user report: https://www.reddit.com/r/hermesagent/comments/1t4ujs5/ ("agent does not always check Mnemosyne at every turn")

AxDSan:
Fixes #321.

## Problem

The Mnemosyne context block is injected into the turn and is often sufficient to answer the user's question. The agent defaults to retrieval tools out of habit, fires `session_search()`, and only then acknowledges the answer was already in the injected block. Wasted turn, wasted tokens.

## Fix

Add a short standing rule to `MnemosyneMemoryProvider.system_prompt_block()` on both provider paths:

```
When a ## Mnemosyne Context block is injected into the current turn,
read it before calling retrieval tools. If it answers the user's
question, answer directly. Use session_search only when the injected
Mnemosyne context is missing, stale, or insufficient.
```

Wording per dplush's suggested shape in #321. Short enough to keep in the standing prompt without inflating per-turn tokens, and leaves the `prefetch()` output as raw context (no Hermes-specific tags emitted from Mnemosyne).

## Files changed

- `integrations/hermes/src/mnemosyne_hermes/__init__.py` — new provider path
- `hermes_memory_provider/__init__.py` — legacy provider path
- `integrations/hermes/tests/test_system_prompt.py` — pin the rule on the new path
- `tests/test_provider_all_15_tools.py` — pin the rule on the legacy path

The init-failed and skip-context branches are unchanged. The rule only applies when the provider is active.

## Tests

Two new tests pin the rule on each provider path. All 33 tests across `test_system_prompt.py` and `test_provider_all_15_tools.py` pass.

## Credit

Diagnosis and suggested wording from @dplush in #321. I picked this up because I committed to shipping it in the original issue thread and the dplush-shaped fix is small and contained. dplush is credited here for the implementation shape.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1560,7 +1560,12 @@ class MnemosyneMemoryProvider(MemoryProvider):
                 "Active (native local memory). Use mnemosyne_remember to store ANY "
                 "durable fact, preference, identity, or insight. Use mnemosyne_recall to search. "
                 "Use mnemosyne_shared_* tools for manual shared surface CRUD. "
-                "The legacy memory tool is deprecated for durable storage — Mnemosyne is primary."
+                "The legacy memory tool is deprecated for durable storage — Mnemosyne is primary.\n"
+                "\n"
+                "When a `## Mnemosyne Context` block is injected into the current turn, "
+                "read it before calling retrieval tools. If it answers the user's question, "
+                "answer directly. Use session_search only when the injected Mnemosyne "
+                "context is missing, stale, or insufficient."
             )
         # C27: when init failed (as opposed to a deliberate skip-context),
         # surface the failure in the system prompt so the agent -- and through
--- a/integrations/hermes/src/mnemosyne_hermes/__init__.py
+++ b/integrations/hermes/src/mnemosyne_hermes/__init__.py
@@ -881,7 +881,12 @@ class MnemosyneMemoryProvider(MemoryProvider):
                 "mnemosyne_validate/invalidate/update/forget for provenance, corrections, stale facts, and cleanup; "
                 "mnemosyne_scratchpad_* for temporary working notes; "
                 "mnemosyne_shared_* only for compact cross-agent stable metadata, never raw conversation.\n"
-                "Prefer compact, declarative, non-imperative memories. Do not save one-off task progress."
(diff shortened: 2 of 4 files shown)


AxDSan:
Shipped. dplush's suggested wording went in on both provider paths (new integrations/hermes path and the legacy hermes_memory_provider path), with focused tests pinning the rule on each. Closed #321 with this merge.

Thanks @dplush for the suggested shape, the wording was ready to go. Thanks @michaeltanyk for the diagnosis.

PR #347, opened 2026-06-17
fix(hermes): detect broken plugin symlinks in status (#343)

### LINKED ISSUE #343: mnemosyne-hermes should detect broken Hermes plugin symlinks
## Problem

`mnemosyne-hermes install` creates a plugin symlink under Hermes home:

```text
$HERMES_HOME/plugins/mnemosyne -> <hermes-venv>/site-packages/mnemosyne_hermes
```

That works until the package disappears from that exact path. If the Hermes venv is rebuilt, cleaned up, or the package is removed/reinstalled, the symlink can be left dangling.

When that happens, Hermes reports Mnemosyne as not installed, but it does not say the real problem is a broken symlink.

## What I saw

On a Hermes install with `HERMES_HOME=/opt/data`, the plugin path existed but pointed to a missing target:

```text
/opt/data/plugins/mnemosyne -> /opt/hermes/.venv/lib/python3.13/site-packages/mnemosyne_hermes
```

The target directory no longer existed.

`hermes memory status` showed:

```text
Provider:  mnemosyne
Plugin:    NOT installed ✗
Install the 'mnemosyne' memory plugin to ~/.hermes/plugins/
```

Direct provider discovery failed too:

```python
from plugins.memory import load_memory_provider
p = load_memory_provider("mnemosyne")
print(bool(p))  # False
```

The fix was to reinstall the packages into the Hermes venv and refresh the symlink:

```bash
uv pip install -p /opt/hermes/.venv \
  'mnemosyne-memory[embeddings]==3.8.0' \
  'mnemosyne-hermes==0.2.0'

/opt/hermes/.venv/bin/mnemosyne-hermes \
  --hermes-home /opt/data install --force --no-bootstrap
```

After that:

```text
Plugin:    installed ✓
Status:    available ✓
provider_loaded: True
```

## Expected behavior

`mnemosyne-hermes status` should call out this state directly:

```text
Broken symlink:
  /opt/data/plugins/mnemosyne -> /opt/hermes/.venv/.../mnemosyne_hermes
  target does not exist

Suggested fix:
  mnemosyne-hermes --hermes-home /opt/data install --force --no-bootstrap
```

`install --force` already fixes the case once you know to run it. The missing piece is the diagnosis.

## Small fix option

Teach `status` / `is_installed()` to distinguish these cases:

- no plugin path
- valid plugin symlink or directory
- broken symlink
- path exists but does not look like a memory provider

That would have made this a two-minute fix instead of a small debugging session.

## More robust option

Instead of symlinking directly into `site-packages`, `mnemosyne-hermes install` could write a tiny shim plugin directory under:

```text
$HERMES_HOME/plugins/mnemosyne/
```

with an `__init__.py` that imports the installed package:

```python
from mnemosyne_hermes import *
```

That keeps the Hermes plugin path stable. It still requires `mnemosyne-hermes` to be installed in Hermes' Python environment, but it avoids a stale symlink to an old `site-packages` location.

The minimal diagnostic fix is probably enough for a first pass.

## Environment

```text
Hermes home: /opt/data
Hermes venv: /opt/hermes/.venv
Python: 3.13
mnemosyne-memory after repair: 3.8.0
mnemosyne-hermes after repair: 0.2.0
```

dplush:
## Summary

Add `PluginState` enum and `plugin_state()` helper in `mnemosyne-hermes install` to distinguish:

- `OK` – valid plugin symlink
- `BROKEN_SYMLINK` – symlink exists but target is missing (venv rebuild, reinstall, etc.)
- `MISSING` – no symlink

`mnemosyne-hermes status` now reports broken symlinks explicitly with a remediation hint.

`is_installed()` remains unchanged for backwards compatibility.

Fixes #343.

## Test plan

- Existing installs continue to report `installed ✓`
- Broken symlink scenario now reports `broken symlink (target missing) ✗` + hint
- New `plugin_state()` is additive only

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -8,8 +8,17 @@ import os
 import shutil
 import subprocess
 import sys
+from enum import Enum
 from pathlib import Path
 from typing import Optional
+
+
+class PluginState(Enum):
+    """Possible states of the Hermes plugin symlink."""
+    OK = "ok"
+    BROKEN_SYMLINK = "broken_symlink"
+    MISSING = "missing"
+
 
 PLUGIN_NAME = "mnemosyne"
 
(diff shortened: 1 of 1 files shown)


AxDSan:
Good stuff @dplush. Tiny diff, real diagnostic value, ships first in the installer chain so #354 and #356 can build on it cleanly. Closed #343 with this merge.

PR #330, opened 2026-06-16
feat: add automated model refresh and recall tooling

WXBR:
## Summary

This PR adds automated sleep-time model refresh plus opt-in observability/continuity surfaces for Mnemosyne:

- add sleep-time model update inference, validation, deterministic auto-apply/reject, and diagnostic listing
- add `mnemosyne_model_card` and `mnemosyne_model_refresh` for canonical model-card inspection and model-refresh diagnostics
- add `MNEMOSYNE_QUERY_INTENT=1` gated query-intent weight adjustment in `BeamMemory.recall()`
- expose `mnemosyne_recall_diagnostics` for recall tier/fallback counters, gated by `MNEMOSYNE_RECALL_DIAGNOSTICS=1`
- add `mnemosyne_task_progress`, an owner-scoped canonical `task:progress` tool for curated “where did we leave off?” state
- replace the dead `mnemosyne.core.orchestrator` stub with a thin compatibility wrapper over `BeamMemory.recall()`

## Behavior changes

- Hermes `sync_roles` now defaults to user-turn autosave only. The old default was `["user", "assistant"]`; the new default is `["user"]`.
- Existing deployments that want to keep assistant-turn autosave should set this explicitly in `config.yaml`:

```yaml
memory:
  mnemosyne:
    sync_roles: ["user", "assistant"]
```

- This changes the release target to `3.11.0` because the default autosave behavior changes silently for deployments that did not already override `sync_roles`.

## Notes

- `mnemosyne_recall_diagnostics` returns disabled unless explicitly enabled via env var.
- `mnemosyne_model_refresh` is diagnostic-only; the schema exposes `action: "list"` only.
- `mnemosyne_task_progress` uses the existing `CanonicalStore` owner scoping instead of raw global keys.
- Normal recall remains retrieval-only. No LLM reranking or new network dependency is added.
- Query-intent weighting is opt-in and explicit caller weights still win.
- `MNEMOSYNE_RECALL_EXTRA_STOPWORDS` is read at import time through `_FACT_MATCH_STOPWORDS`, so changing that env var requires a process restart to affect recall stopword matching.

## Verification

Ran locally after rebasing on current `upstream/main` and applying the release-hygiene fixes:

```bash
PYTHONPATH=integrations/hermes/src:. python3 -m pytest tests/ -q -o 'addopts='
# 1712 passed, 28 skipped in 34.05s

cd integrations/hermes && PYTHONPATH=src:../.. python3 -m pytest tests/ -q -o 'addopts='
[...]

PR #348, opened 2026-06-17
test(hermes): add provider parity coverage

### LINKED ISSUE #323: Add parity tests for Hermes provider implementations to prevent behavior drift
Mnemosyne has had more than one Hermes-facing provider implementation or integration path over time. That is understandable, but it makes drift easy:

- one path exposes a different tool set
- one path applies config defaults differently
- one path uses different sync limits
- one path handles scope/session/global memory differently
- one path gets a bugfix and the other does not

Some of these issues have already shown up in real Hermes installs.

## Proposal

Add parity tests around the Hermes provider surface.

At minimum, the tests should compare:

- registered tool schemas
- tool names
- important schema defaults
- provider availability checks
- config key handling
- remember/recall behavior for session and global memory
- canonical fact behavior, if supported by both paths

Longer term, a shared provider base or compatibility layer would reduce the need to keep behavior aligned manually.

## Acceptance criteria

- tests fail if one Hermes provider path exposes different tools unexpectedly
- tests fail if important config defaults drift
- tests cover session-scoped and global remember/recall behavior
- duplicated provider logic is either reduced or explicitly covered by parity tests

This is mostly a maintenance issue. The goal is to stop integration bugs before they reach users.

dplush:
## Summary

- add a Hermes provider parity test module covering `hermes_memory_provider` and `integrations/hermes/src/mnemosyne_hermes`
- assert both provider copies expose the same tool set, tool schemas, config defaults, sync limits, and sync-turn scope/truncation behavior
- align the drift the new tests exposed: sync/persona tool schemas, `sync_roles`, `mnemosyne_triple_end`, sync/persona dispatch, triple metadata handling, and zero-limit sync-turn behavior

Closes #323.

## Test plan

- `uv run --no-sync --with pytest pytest tests/test_hermes_provider_parity.py -q`
- `uv run --with pytest pytest tests/test_hermes_provider_parity.py tests/test_sync_turn_content_limit.py -q`
- `PYTHONPATH=".:integrations/hermes/src" uv run --no-sync --with pytest pytest tests/test_hermes_provider_parity.py tests/test_sync_turn_content_limit.py tests/test_provider_all_15_tools.py tests/test_persona_adapter.py tests/test_persona_schema.py -q`

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -961,6 +961,22 @@ try:
 except Exception:  # pragma: no cover - sync extras are optional at import time
     ALL_SYNC_TOOL_SCHEMAS = []
 
+try:
+    from hermes_memory_provider.persona_tools import (
+        PERSONA_PROMOTE_SCHEMA,
+        PERSONA_DEMOTE_SCHEMA,
+        PERSONA_LIST_SCHEMA,
+        PERSONA_REINFORCE_SCHEMA,
+    )
+    ALL_PERSONA_TOOL_SCHEMAS = [
+        PERSONA_PROMOTE_SCHEMA,
+        PERSONA_DEMOTE_SCHEMA,
+        PERSONA_LIST_SCHEMA,
+        PERSONA_REINFORCE_SCHEMA,
+    ]
+except Exception:  # pragma: no cover - persona extras are optional at import time
+    ALL_PERSONA_TOOL_SCHEMAS = []
+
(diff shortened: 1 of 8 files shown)


AxDSan:
Good stuff @dplush. Parity coverage on the right surface, and you sequenced it right: this lands before the larger #353 and #330 that touch the same provider files, so the new tests catch any drift those introduce. Closed #323 with this merge.

Issue #318, opened 2026-06-15
[BUG] `sleep_all_sessions(force=True)` does not bypass age cutoff on per-session consolidation


pioneerAlone:
## Summary

`sleep_all_sessions(force=True)` only partially honors the `force` flag. The
outer loop sets `cutoff = datetime.max.isoformat()` correctly, so the
`SELECT session_id, COUNT(*) ... WHERE timestamp < ?` finds eligible rows
across all sessions. But when it then calls `beam.sleep(dry_run=dry_run)`
for each session, the `force` argument is **dropped on the floor**. The
per-session `sleep()` recomputes its own cutoff with
`datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS // 2)` (12h
default), so any session whose memories are all younger than 12h returns
`no_op`. The aggregated result therefore reports `items_consolidated: 0`
even when 100+ unconsolidated rows exist.

## Reproduction

```python
from mnemosyne import Mnemosyne
m = Mnemosyne()
# After a fresh batch of 100+ working-memory writes in the last 12h:
r = m.sleep_all_sessions(force=True)
# -> status='no_op', items_consolidated=0
```

Confirmed on mnemosyne-memory 3.7.0 (PyPI). `sleep_all_sessions(force=True)`
on a fresh DB with all rows < 12h old will always return `no_op`, while
`mnemosyne_stats` reports `unconsolidated=N` for N > 0.

## Root cause

`mnemosyne/core/beam.py`, `sleep_all_sessions()` calls:
```python
result = beam.sleep(dry_run=dry_run)   # force not forwarded
```
The fix is to forward `force`:
```python
result = beam.sleep(dry_run=dry_run, force=force)
```

This was introduced (or at least left in) after the --force feature was
shipped in commit b71a971b (issue #279), which only wired the flag through
`mnemosyne.cli.cmd_sleep` → `sleep_all_sessions`, not all the way down to
the per-session `sleep()` call.

## Suggested fix

In `sleep_all_sessions()`, change:
```python
result = beam.sleep(dry_run=dry_run)
```
to:
```python
result = beam.sleep(dry_run=dry_run, force=force)
```

## Environment

- mnemosyne-memory 3.7.0
- mnemosyne-hermes 0.1.8
- Python 3.11.15
- macOS arm64
- DB: /Users/wangbo/.hermes/mnemosyne/data/mnemosyne.db
- 104 unconsolidated working memories, all < 12h old
[...]

PR #352, opened 2026-06-18
fix: propagate force in sleep_all_sessions

WXBR:
## Summary

Fixes `BeamMemory.sleep_all_sessions(force=True)` so the `force` flag is passed into each per-session `sleep()` call.

Without this, `sleep_all_sessions(force=True)` selected fresh/non-age-eligible sessions in the outer query, but each delegated session still ran `sleep(force=False)`. The result was a misleading no-op for rows that should have been consolidated by a forced maintenance pass.

## What changed

- Pass `force=force` from `sleep_all_sessions()` to per-session `sleep()`.
- Add a regression test that seeds a fresh working-memory row and verifies `sleep_all_sessions(force=True)` consolidates it immediately.

Related to #318, but this PR includes the concrete fix plus regression coverage.

## Verification

- RED: the new regression failed before the fix with `assert 0 == 1` for `sessions_consolidated`.
- `python -m pytest tests/test_beam_e3_additive_sleep.py::TestE3AdditiveSleep::test_sleep_all_sessions_force_consolidates_fresh_rows tests/test_beam_e3_additive_sleep.py::TestE3AdditiveSleep::test_manual_reclaim_then_sleep_all_sessions_consolidates_orphan -q -o 'addopts='` -> `2 passed`
- `python -m pytest tests/test_beam_e3_additive_sleep.py -q -o 'addopts='` -> `22 passed`
- `python -m py_compile mnemosyne/core/beam.py` -> passed
- `git diff --check` -> passed

Full-suite note: with `PYTHONPATH=integrations/hermes/src`, the full suite reached `1681 passed` and one unrelated existing failure in `tests/test_llm_conflict_detector.py::test_beam_integration_with_llm_conflict`, where conflict candidates were empty before this PR's changed path is involved.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -8009,7 +8009,7 @@ class BeamMemory:
                     author_id=self.author_id,
                     author_type=self.author_type,
                 )
-                result = beam.sleep(dry_run=dry_run)
+                result = beam.sleep(dry_run=dry_run, force=force)
                 result = dict(result)
                 result["session_id"] = session_id
                 result["eligible"] = row["eligible"] if hasattr(row, "keys") else row[1]
--- a/tests/test_beam_e3_additive_sleep.py
+++ b/tests/test_beam_e3_additive_sleep.py
@@ -571,6 +571,29 @@ class TestE3AdditiveSleep:
         conn.close()
         assert marker is None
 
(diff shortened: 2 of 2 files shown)


AxDSan:
Good stuff @WXBR. Trivial diff, real bug, the lost force flag in delegation was an easy miss. The CI cancel-cascade was runner infra flake, not a regression, re-run came back fully green.

PR #356, opened 2026-06-18
fix(install): support named Hermes profiles in plugin install

tvinagre:
## What

The Mnemosyne installer (both the legacy `mnemosyne/install.py` and the
canonical `integrations/hermes/src/mnemosyne_hermes/install.py`) only
created a single plugin link under the default `~/.hermes`. Users with
named profiles under `~/.hermes/profiles/<name>/` had the `memory.provider:
mnemosyne` config set correctly, but the plugin link was never created
for their profile, so the provider failed to load.

## Fix

- New `_iter_mnemosyne_profiles()` discovers every profile whose
  `config.yaml` selects `memory.provider: mnemosyne`. Uses a real YAML
  parse (with a regex fallback only when PyYAML is unavailable, so the
  core package keeps its zero-runtime-dep claim).
- New `_link_all_profiles()` creates the per-profile plugin link,
  idempotent and with continue-on-failure semantics.
- New `_unlink_all_profiles()` mirrors the install pass on uninstall,
  scanning `profiles/` by *link* (not by config opt-in) so orphans
  after a provider switch are still cleaned up.
- New `_verify_links()` reports PASS/FAIL for the default home and
  every opted-in profile.
- Symlinked profile entries are skipped during discovery (avoids
  confused-deputy writes if the installer is ever run privileged).
- `--force` on the canonical installer now logs the old target before
  overwriting.

## Why a single profile is symlinked, not all of them at once

`install_plugin()` and `uninstall_plugin()` only ever target one
`HERMES_HOME` per call. With the `--hermes-home` flag set to a profile
home, the per-profile peer scan no-ops (correct: the caller is
explicitly targeting one profile). The bulk scan is the default when
`--hermes-home` is not overridden (i.e. the global install). The
canonical installer will accept an explicit `--all-profiles` flag as a
follow-up; open question noted in `/tmp/phase6-review.md`.

## Tests

30/30 pass when run from the repo root:

```
tests/test_install.py: 14 passed
integrations/hermes/tests/test_install_hermes.py: 16 passed
```

Covers: default install, opted-in profiles only, idempotent rerun,
missing `profiles/` dir, profile with no config, uninstall (with
regression test for the orphan-after-config-change case), foreign link
preservation, comment/whitespace in provider config, symlinked profile
dir skipped, malformed YAML, force-replaces-real-dir, OSError
isolation, home-link-also-removed-on-uninstall.

## Out of scope / acknowledged in this PR

- Legacy `mnemosyne/install.py`'s `force=True` kwarg is intentionally
  unreachable from any current entrypoint. The legacy installer is on
  the way out (see `integrations/hermes/`), and the safe default
  (never delete user data without an explicit override) is correct.
  The kwarg is kept for parity with the canonical installer and ready
  if the legacy entrypoint is later extended.
- The PyYAML regex fallback can match `provider: mnemosyne` at any
  nesting level (the YAML path requires it under `memory:`). The
  fallback is only reached when PyYAML is genuinely unavailable, and
  in practice the core package is run alongside Hermes where PyYAML is
  declared. Narrowing further is a follow-up.

## Out of scope / not addressed

- The "bulk-link-all vs. targeted" design choice (an explicit
  `--all-profiles` flag) is a maintainer call. The current behavior is
  correct for the global `mnemosyne-install` path; the targeted path
  (`mnemosyne-hermes install --hermes-home ~/.hermes/profiles/alice`)
  is correct for a per-profile explicit install. Both work; neither
  has a flag exposing the choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
+- **Named Hermes profiles now get the plugin link** (issue #365). `mnemosyne-install`
+  and `mnemosyne-hermes install` now scan `~/.hermes/profiles/*/config.yaml` for
+  `memory.provider: mnemosyne` and create (or remove, on uninstall) the plugin
+  symlink in each matching profile's `plugins/` directory. Previously the link was
+  only created under the default `~/.hermes/`, so the provider silently failed to
+  load for users with named profiles.
+
 - **Host LLM backend registration in skip-context sessions.**
   `register_hermes_host_llm()` was called at the end of
   `MnemosyneMemoryProvider.initialize()`, after the skip-context early
--- a/integrations/hermes/pyproject.toml
+++ b/integrations/hermes/pyproject.toml
@@ -13,6 +13,7 @@ requires-python = ">=3.10"
 keywords = ["hermes-agent", "memory", "mnemosyne", "ai-agents", "local-first", "long-term-memory"]
 dependencies = [
     "mnemosyne-memory[embeddings]>=3.1",
+    "PyYAML>=6.0",
 ]
 classifiers = [
     "Development Status :: 4 - Beta",
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -308,7 +308,186 @@ def install_plugin(
 
     target.parent.mkdir(parents=True, exist_ok=True)
     os.symlink(str(source), str(target))
+
+    # Link any named profiles that opt into Mnemosyne (no-op without profiles).
+    _link_all_profiles(source, hermes_home_path=hermes_home_path, force=force)
+
     return target
+
+
+def _config_selects_mnemosyne(text: str) -> bool:
+    """Return True when a profile config selects ``memory.provider: mnemosyne``.
+
+    Prefers a real YAML parse, which ignores comments and tolerates arbitrary
+    whitespace. The line-anchored regex is used **only** when PyYAML is genuinely
+    unavailable (``ImportError``). Malformed YAML is treated as "not opted in"
+    rather than falling through to the looser regex.
+    """
+    try:
+        import yaml
+    except ImportError:
+        import re
+        return re.search(
+            r"^\s*provider\s*:\s*mnemosyne\s*(#.*)?$", text, re.MULTILINE
+        ) is not None
+    try:
+        cfg = yaml.safe_load(text)
+    except yaml.YAMLError:
+        return False
+    if isinstance(cfg, dict):
+        memory = cfg.get("memory")
+        if isinstance(memory, dict):
+            return memory.get("provider") == "mnemosyne"
+    return False
+
+
+def _iter_mnemosyne_profiles(hermes_home_path: str | Path | None = None) -> list[Path]:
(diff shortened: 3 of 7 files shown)


dplush:
This looks useful, especially for setups that actually use several Hermes profiles.

One thing I would coordinate before taking it out of draft: this touches the same installer/status code as the broken-plugin-symlink work in #354 / #347. This branch does not seem to include the newer `plugin_state` / `broken_symlink` handling yet, so a rebase after the diagnostics PR is settled will probably save some conflicts and duplicated installer logic.

@AxDSan Suggested order from my side:

1. land or settle the broken symlink diagnostics first (#354),
2. close/supersede the older overlapping PR (#347),
3. rebase this profile-support PR on top (#356),
4. then check that profile install/uninstall and broken-symlink status still work together.

The profile support itself makes sense to me. It just sits on the same patch of installer code, so I would avoid reviewing it in isolation until the symlink-status PR is sorted.

tvinagre:
Created issue #365 to document the bug this PR fixes, per CONTRIBUTING.md

AxDSan (dismissed):

@tvinagre solid profile-aware install pass. Both legacy `mnemosyne/install.py` and canonical `integrations/hermes/src/mnemosyne_hermes/install.py` now use the same `_iter_mnemosyne_profiles` / `_link_all_profiles` / `_unlink_all_profiles` / `_verify_links` quartet, which is the right shape. PyYAML with regex fallback in the core package keeps the zero-runtime-dep claim intact, and the symlinked-profile-skip in discovery avoids the confused-deputy write case you called out.

Verified on `pr/356`:
- `py_compile` clean on all four changed Python files.
- `pytest tests/test_install.py -q` -> 14 passed.

LGTM with two minor asks before merge:

1. **CHANGELOG entry.** No `[Unreleased]` line for this fix. One line under the next version block.
2. **Version bump.** This is a bugfix (named profile symlink never landed) so PATCH. `mnemosyne/__init__.py` is still on `3.10.1` from #373; bump to `3.10.2` (or whatever the next patch target is).

Question on the `--all-profiles` flag note: that is a follow-up, fine to defer. One thing worth confirming in the code: when `_link_all_profiles()` runs and a profile's `config.yaml` selects a different provider (not `mnemosyne`), the link is skipped by the opt-in filter, but if a profile previously had a `mnemosyne` link and was switched away, the new unlink path's "scan by link, not by opt-in" claim handles the orphan. Good catch on that.

#365 is the companion issue filed per CONTRIBUTING.md. This PR closes it cleanly.

tvinagre:
No worries, thank you for the review @AxDSan !

Rebased onto main (now at v3.10.1 — picked up the JWT security release) and addressed both asks:
- CHANGELOG: ### Fixed entry under [Unreleased] (closes #365).
- Version: bumped to 3.10.2 (correct next patch now that main is 3.10.1).

No changes to the profile-install logic itself, the rebase was clean there; only CHANGELOG/version conflicted (trivial). Force-pushed due to the rebase.

AxDSan (approved):
Re-reviewed the rebase diff. Named profile support is merged cleanly with the wrapper install mode from the previous PR. Both features now coexist without conflicts. Approving.

AxDSan:
Hey @tvinagre, your PR content landed on main as part of the v3.11.0 release. Named Hermes profiles plugin link support. Solid work on this! Landed on main for v3.11.0. Closing this out. Appreciate the contribution!

AxDSan:

PR #361, opened 2026-06-19
fix(cli): fallback to absolute import for host LLM backend in standalone module loading

tml89:
## Problem

`register_hermes_host_llm()` in `mnemosyne_command()` uses a relative import:

```python
from .hermes_llm_adapter import register_hermes_host_llm
```

This silently fails when `cli.py` is loaded as a standalone module — which is what happens in Hermes' user-plugin discovery path. `plugins/memory/__init__.py:discover_plugin_cli_commands()` uses `importlib.util.spec_from_file_location()` to load `cli.py` directly from the plugin directory without registering the parent package. Relative imports break in this context.

The surrounding `try/except` swallows the `ImportError`, so `register_hermes_host_llm()` never runs and `MNEMOSYNE_HOST_LLM_ENABLED=true` is silently ignored when sleep is invoked via CLI (e.g. cron jobs calling `hermes mnemosyne sleep --all-sessions`). Consolidation falls back to AAAK with no LLM.

## Fix

Try the relative import first (preserves existing behavior when loaded as part of the package), then fall back to an absolute import on `ImportError`:

```python
try:
    from .hermes_llm_adapter import register_hermes_host_llm
except ImportError:
    from mnemosyne_hermes.hermes_llm_adapter import register_hermes_host_llm
```

Applied to both copies of `cli.py` (`integrations/hermes/src/mnemosyne_hermes/cli.py` and `hermes_memory_provider/cli.py`).

## Verification

Before fix:
```
$ hermes mnemosyne sleep --dry-run --all-sessions | jq '.llm_used'
0
```

After fix:
```
$ hermes mnemosyne sleep --dry-run --all-sessions | jq '.llm_used'
2
```

## Context

This builds on commit fdac938 ("fix: register Hermes host LLM backend in CLI path") which added the registration call but used a relative import that doesn't work in the standalone-module loading path.

### DIFF
--- a/hermes_memory_provider/cli.py
+++ b/hermes_memory_provider/cli.py
@@ -69,9 +69,17 @@ def mnemosyne_command(args):
         print("Usage: hermes mnemosyne {stats|sleep|version|inspect|clear|export|import}")
         return 1
 
-    # Register Hermes host LLM backend so sleep uses Hermes' provider
+    # Register Hermes host LLM backend so sleep uses Hermes' provider.
+    # Use a try/except fallback chain: the relative import works when loaded
+    # as part of the hermes_memory_provider package; the absolute import is
+    # needed when this module is loaded standalone (e.g. Hermes user-plugin
+    # discovery via importlib.util.spec_from_file_location, which does not
+    # set up the parent package — breaking relative imports silently).
     try:
-        from .hermes_llm_adapter import register_hermes_host_llm
+        try:
+            from .hermes_llm_adapter import register_hermes_host_llm
+        except ImportError:
+            from mnemosyne_hermes.hermes_llm_adapter import register_hermes_host_llm
         register_hermes_host_llm()
     except Exception:
         pass
--- a/hermes_memory_provider/cli.py
+++ b/hermes_memory_provider/cli.py
@@ -79,7 +79,7 @@ def mnemosyne_command(args):
         try:
             from .hermes_llm_adapter import register_hermes_host_llm
         except ImportError:
-            from mnemosyne_hermes.hermes_llm_adapter import register_hermes_host_llm
+            from hermes_memory_provider.hermes_llm_adapter import register_hermes_host_llm
         register_hermes_host_llm()
     except Exception:
         pass
--- a/integrations/hermes/src/mnemosyne_hermes/cli.py
+++ b/integrations/hermes/src/mnemosyne_hermes/cli.py
@@ -65,9 +65,17 @@ def mnemosyne_command(args):
         print("Usage: hermes mnemosyne {stats|sleep|version|inspect|clear|export|import}")
         return 1
 
-    # Register Hermes host LLM backend so sleep uses Hermes' provider
+    # Register Hermes host LLM backend so sleep uses Hermes' provider.
+    # Use a try/except fallback chain: the relative import works when loaded
+    # as part of the mnemosyne_hermes package; the absolute import is needed
+    # when this module is loaded standalone (e.g. Hermes user-plugin
+    # discovery via importlib.util.spec_from_file_location, which does not
+    # set up the parent package — breaking relative imports silently).
     try:
-        from .hermes_llm_adapter import register_hermes_host_llm
+        try:
+            from .hermes_llm_adapter import register_hermes_host_llm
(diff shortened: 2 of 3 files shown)


dplush:
Thanks for chasing this. The standalone loading case is the right one to fix.

I think there’s one small package-path detail left: in `hermes_memory_provider/cli.py`, the absolute fallback should probably use the root package:

```python
from hermes_memory_provider.hermes_llm_adapter import register_hermes_host_llm
```

rather than:

```python
from mnemosyne_hermes.hermes_llm_adapter import register_hermes_host_llm
```

The `mnemosyne_hermes` fallback looks right for `integrations/hermes/src/mnemosyne_hermes/cli.py`. For the root `hermes_memory_provider/cli.py` path, though, a standalone `spec_from_file_location()` load may only have the root package importable. In that setup the relative import fails, `mnemosyne_hermes` may not exist on `sys.path`, but `hermes_memory_provider.hermes_llm_adapter` does.

So I’d only change the fallback target in the root CLI copy, and add a small regression test that loads both CLI files via `spec_from_file_location()` and verifies the registration path is reached.

tml89:
Good catch on the package path. Updated:

- `hermes_memory_provider/cli.py` now falls back to `from hermes_memory_provider.hermes_llm_adapter import register_hermes_host_llm` (not `mnemosyne_hermes`)
- `integrations/hermes/src/mnemosyne_hermes/cli.py` keeps `from mnemosyne_hermes.hermes_llm_adapter` as fallback (correct for that package)
- Added `tests/test_cli_host_llm_standalone_loading.py` — loads both CLI files via `spec_from_file_location` and asserts `get_host_llm_backend() is not None` after calling `mnemosyne_command()`. Both tests pass.

AxDSan (approved):

Clean, scoped fix. The relative `from .hermes_llm_adapter import ...` was the right pattern when `cli.py` is loaded inside the package, but Hermes' plugin loader (`mnemosyne/core/plugins.py:_discover_plugins`) calls `importlib.util.spec_from_file_location(file_path.stem, ...)` and registers the module under just its filename, no parent package. The surrounding `try/except` then swallows the `ImportError` and `MNEMOSYNE_HOST_LLM_ENABLED=true` is silently dropped.

Verified locally on `pr/361`:
- `py_compile` clean on both `cli.py` copies and the new test.
- `pytest tests/test_cli_host_llm_standalone_loading.py` -> 2 passed (parametrized over both CLI copies).
- Reverted `cli.py` files to `main`, re-ran the same test -> both cases FAIL with `register_hermes_host_llm() was not reached ... the relative import likely failed silently`. Confirms the test actually guards the regression.
- CI green: build, docs-check, test on 3.10/3.11/3.12/3.13 all SUCCESS.

One documentation nit, not a blocker: the PR body references `plugins/memory/__init__.py:discover_plugin_cli_commands()`, but the actual loader in this repo is `_discover_plugins` in `mnemosyne/core/plugins.py` (no `discover_plugin_cli_commands` function exists). The fix and the test are correct because they exercise the same `spec_from_file_location` pattern; just the prose should be fixed before merge.

Overlap with #368 worth calling out: #368's branch includes this exact `cli.py` fallback plus a separate fix in `MnemosyneMemoryProvider.initialize()` that moves `register_hermes_host_llm()` before the skip-context early return, and a companion change in `shutdown()` to skip unregister when the session is a skip context. So #368 is a strict superset of #361. My recommendation: merge #368 first, close #361 as superseded. If you want #361 in as a smaller blast-radius commit, that is fine too, but the `__init__.py` ordering fix really should land alongside it.

No CHANGELOG entry, no version bump on either. Acceptable here since these are correctness fixes with no API change. I'd ask for a one-line `[Unreleased]` entry when the dust settles.

Approving with a request that the PR body be tightened to reference `_discover_plugins` correctly, or that #368 land first and #361 close as superseded.

tml89:
Closing as superseded by #368, which is a strict superset of this PR and includes the `__init__.py` ordering fix alongside the CLI fallback import fix. AxDSan recommended merging #368 first in his review.

The CHANGELOG entry and `_discover_plugins` doc-nit fix requested in the review have been pushed to #368 in commit 35627ea.

PR #363, opened 2026-06-19
fix(hermes): make mnemosyne CLI bank-aware under profile_isolation (#362)

### LINKED ISSUE #362: mnemosyne stats ignores profile_isolation banks — always reads the default bank, reports 0 while data lives in banks/<profile>/
## Bug

With `profile_isolation: true`, the Hermes provider **writes** captured memories to a per-profile bank (`<data_dir>/banks/<profile>/mnemosyne.db`), but `mnemosyne stats` **reads** the default/legacy bank (`<data_dir>/mnemosyne.db`). `stats` therefore reports `working.total: 0` no matter how much is stored in the profile bank, and there is no flag to point it at the right bank.

## Versions

- mnemosyne-hermes 0.2.0
- mnemosyne-memory 3.10.0
- Python 3.11

## Root cause

`stats` builds its beam with no bank, so it binds to the default bank.

`integrations/hermes/src/mnemosyne_hermes/cli.py` (stats handler):
```python
beam = BeamMemory(session_id="hermes_default")   # no bank -> default bank
...
working = beam.get_global_working_stats() if getattr(args, "global", False) else beam.get_working_stats()
```

The provider's write path resolves a per-profile bank when isolation is enabled.

`integrations/hermes/src/mnemosyne_hermes/__init__.py`:
```python
bank_name = self._resolve_profile_bank()   # e.g. "zedd"
Mnemosyne(bank=bank_name, ...)             # -> banks/zedd/mnemosyne.db
```

Default bank = legacy top-level db (`mnemosyne/core/banks.py`):
```python
# The 'default' bank uses the legacy path (data_dir/mnemosyne.db).
return self.data_dir / "mnemosyne.db"
```

So writes go to `banks/<profile>/mnemosyne.db` and `stats` reads `mnemosyne.db` — different files.

The `stats` subparser exposes only `--global` (still scoped to the default bank). Unlike `import` (`import_cmd.add_argument("--bank", ...)`), `stats` has no `--bank`, so there is no way to target the active bank.

## Reproduction

1. Set `profile_isolation: true` for a profile.
2. Capture memories in that profile (its `banks/<profile>/mnemosyne.db` gets `working_memory` rows).
3. `mnemosyne stats` -> `working.total: 0` (and `--global` too), despite rows in the profile bank.

## Impact

`stats` is unusable under `profile_isolation` — always `0`. The same pattern likely affects other handlers that build beams via `...(session_id="hermes_default")` without resolving the active bank (e.g. `inspect`, `sleep`).

## Suggested fix

- Add `--bank` to `stats` (and `inspect`/`sleep`), mirroring `import`.
- When `profile_isolation` is on, auto-resolve the active profile bank in these handlers (reuse `_resolve_profile_bank()`).
- Optionally make `--global` aggregate across all banks.

Happy to PR.

doziedotdev:
## Summary

`hermes mnemosyne <stats|inspect|sleep|export>` built its beam with a bare `BeamMemory(session_id="hermes_default")`, which always binds to the default/legacy bank (`<data_dir>/mnemosyne.db`). Under `profile_isolation` the provider writes to a per-profile bank (`<data_dir>/banks/<profile>/mnemosyne.db`), so these commands reported empty results even when the profile bank held data.

This makes the CLI resolve the active bank the same way the provider does:

- New `--bank` flag on `stats`, `sleep`, `inspect`, `export` for explicit targeting.
- When no `--bank` is given and `profile_isolation` is enabled, auto-resolve the profile bank from `HERMES_HOME` (mirroring the provider's `_resolve_profile_bank` fallback), reusing `MnemosyneMemoryProvider._sanitize_bank_name`.
- The beam is then built via `Mnemosyne(session_id="hermes_default", bank=...).beam`, matching how the provider builds its own beam.

`import --bank` is left untouched — it names the *source* provider bank (e.g. Hindsight), not the Mnemosyne target, so it is explicitly excluded from target resolution.

Backward-compatible: with isolation off, no `--bank`, or an unresolvable bank, the CLI uses the default bank exactly as before. Resolution never raises — any failure falls back to the default bank.

Fixes #362.

## Test plan

- New `integrations/hermes/tests/test_cli_bank.py` covers: explicit `--bank` (sanitized), profile bank resolved when isolation is on, default bank when isolation is off / no config / root `.hermes` home, and `import --bank` not redirecting the target.
- `pytest integrations/hermes/tests/` → **16 passed** (6 new), no regressions.
  - Note: `test_install_status.py` currently fails to *collect* on `main` due to a pre-existing missing `from enum import Enum` in `install.py` (`class PluginState(Enum)` at `install.py:16`) — unrelated to this change; excluded from the run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+### Fixed
+
+- **hermes integration:** `hermes mnemosyne <stats|sleep|inspect|export>` are now
+  bank-aware under `profile_isolation` — they resolve the active profile bank (or an
+  explicit `--bank`) instead of always reading the default bank, which reported empty
+  state when the profile bank held the data. (#362, #363)
+
 ## [3.10.0] — 2026-06-18
 
 ### Added
--- a/integrations/hermes/src/mnemosyne_hermes/cli.py
+++ b/integrations/hermes/src/mnemosyne_hermes/cli.py
@@ -6,8 +6,14 @@ Available via: hermes mnemosyne <subcommand>
 from __future__ import annotations
 
 import json
+import os
 import sys
 from pathlib import Path
+
+_BANK_HELP = (
+    "Mnemosyne bank to operate on. Defaults to the active Hermes profile's bank "
+    "when profile_isolation is enabled, otherwise the shared default bank."
+)
 
 
 def register_cli(subparser):
@@ -16,15 +22,18 @@ def register_cli(subparser):
 
     stats_cmd = mn_cmds.add_parser("stats", help="Show memory statistics")
     stats_cmd.add_argument("--global", "-g", action="store_true", help="Show global stats across all sessions")
+    stats_cmd.add_argument("--bank", type=str, help=_BANK_HELP)
 
     sleep_cmd = mn_cmds.add_parser("sleep", help="Run consolidation cycle")
     sleep_cmd.add_argument("--all-sessions", action="store_true", help="Consolidate eligible old working memories across all sessions")
     sleep_cmd.add_argument("--dry-run", action="store_true", help="Report what would be consolidated without writing changes")
+    sleep_cmd.add_argument("--bank", type=str, help=_BANK_HELP)
     mn_cmds.add_parser("version", help="Show Mnemosyne version")
 
     inspect_cmd = mn_cmds.add_parser("inspect", help="Search memories")
     inspect_cmd.add_argument("query", nargs="?", default="", help="Search query")
     inspect_cmd.add_argument("--limit", type=int, default=10, help="Max results")
+    inspect_cmd.add_argument("--bank", type=str, help=_BANK_HELP)
 
     mn_cmds.add_parser("clear", help="Clear scratchpad")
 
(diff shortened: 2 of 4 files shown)


AxDSan (dismissed):
Sorry for the delay, was IRL busy the last few days. Catching up on the board now.

LGTM, this is a clean targeted fix for issue 362.

What I verified:
- `_resolve_cli_bank` correctly mirrors the provider's `_resolve_profile_bank` fallback: HERMES_HOME basename, isolation guard, default-bank treatment of `.hermes`, `hermes`, and `default`. Diff is small and surgical.
- Explicit bank flag precedence over profile resolution is right. The `import` bank argument is correctly excluded since it names the source provider, not the target. Good catch calling that out.
- Backward compatible. With isolation off, no bank flag, or any resolution failure, the CLI falls through to the bare `BeamMemory(session_id="hermes_default")` path. No exceptions leak.
- The bank-aware path uses `Mnemosyne(session_id="hermes_default", bank=bank).beam`, matching how the provider itself constructs the beam under profile_isolation. Consistent surface.
- Local run: `pytest integrations/hermes/tests/` → 16 passed (including the 6 new in `test_cli_bank.py`). The `test_install_status.py` collection failure is pre-existing on `main` (missing `from enum import Enum` in `install.py:16`), confirmed unrelated to this PR.
- CI: all 6 checks green on the latest run.

Minor asks before merge:
- Please add a one-line entry under `[Unreleased]` in `CHANGELOG.md` noting the CLI bank awareness under profile_isolation. A 3.10.2 PATCH bump in `mnemosyne/__init__.py` would be appropriate when this lands, since this is a user-facing behavior fix in the integration layer.
- The `_profile_isolation_enabled` helper re-parses config.yaml here while the provider caches its parsed config. Not blocking, but worth a follow-up to share the parser if more CLI commands grow profile awareness.

Approve. Solid fix, well-scoped test coverage. Thanks for the clear writeup.

doziedotdev:
Thanks for the feedback @AxDSan 

I made the changes to both the CHANGELOG.md and the __init__.py files.

The PR should be good to merge once you eyeball it.

AxDSan:
Hey @doziedotdev, your PR content landed on main as part of the v3.11.0 release. CLI bank-aware fix for profile_isolation. Thanks for the clean fix! Your code landed on main as part of the v3.11.0 release. Closing this out. Appreciate the contribution!

AxDSan:

PR #367, opened 2026-06-20
Add wrapper install mode for Hermes plugin discovery

### LINKED ISSUE #355: Support stable wrapper install mode for Docker/read-only Hermes venvs
## Problem

`mnemosyne-hermes install` currently installs Hermes discovery by creating:

```text
$HERMES_HOME/plugins/mnemosyne -> <python-env>/site-packages/mnemosyne_hermes
```

That works for normal writable Python environments, but it is fragile for Docker/image-based Hermes installs:

- a new Hermes image can replace or rebuild `/opt/hermes/.venv`
- the Hermes venv may be root-owned/read-only, so bootstrap into it fails
- the plugin symlink can point at a package directory that no longer exists
- persistent Hermes state usually lives under `$HERMES_HOME` (for example `/opt/data`), while the image venv is runtime state

This is closely related to #343, but that issue is about diagnosing a broken symlink. This issue is about a more durable install mode that avoids creating fragile symlinks into image-owned venvs.

## Proposed v1

Add a stable wrapper install mode, for example:

```bash
mnemosyne-hermes install --mode wrapper
```

It would create a real persistent plugin directory:

```text
$HERMES_HOME/plugins/mnemosyne/__init__.py
$HERMES_HOME/plugins/mnemosyne/plugin.yaml
```

The wrapper would:

1. add a configured Mnemosyne package `site-packages` path to `sys.path`
2. import and re-export `mnemosyne_hermes`
3. keep the marker strings Hermes discovery expects (`MemoryProvider` / `register_memory_provider`)

A simple generated wrapper shape could be:

```python
from __future__ import annotations
import sys as _sys
_SITE = "/persistent/path/to/venv/lib/pythonX.Y/site-packages"
if _SITE not in _sys.path:
    _sys.path.insert(0, _SITE)
from mnemosyne_hermes import *  # noqa: F401,F403
```

## Expected behavior

- On writable normal installs, current symlink mode can remain the default.
- On read-only Hermes venv / Docker image installs, installer either:
  - auto-detects and recommends wrapper mode, or
  - supports explicit `--mode wrapper --python /path/to/venv/bin/python`.
- `mnemosyne-hermes status` should report whether the plugin is installed as `symlink` or `wrapper`.
- Package updates happen in the persistent side venv; Hermes discovery remains stable across image updates.

## Why this helps

This turns Docker/image-based Hermes updates from a memory outage into a normal package-update flow. The persistent Hermes home keeps the plugin discovery surface, while the selected Python env owns the actual Mnemosyne packages.

dplush:
## Summary
- add explicit `mnemosyne-hermes install --mode wrapper` support
- generate a persistent `$HERMES_HOME/plugins/mnemosyne/` shim that imports `mnemosyne_hermes` from a selected `--python` environment
- extend `mnemosyne-hermes status` to report install mode, wrapper target, and stale wrapper targets
- document wrapper mode for Docker/read-only Hermes venv setups

## Tests
- `uv run --with pytest --with-editable . pytest -q tests/test_install_status.py tests/test_system_prompt.py tests/test_prefetch_hardening.py tests/test_canonical_tools.py`
- `uv run --with-editable . python -m py_compile src/mnemosyne_hermes/install.py`
- smoke-tested `install --mode wrapper --python ...` plus `status`

Closes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+### Added
+
+- **Hermes wrapper install mode for read-only / Docker deployments.**
+  `mnemosyne-hermes install --mode wrapper --python <path>` now creates a
+  stable `$HERMES_HOME/plugins/mnemosyne/` shim that imports from the selected
+  Python environment instead of symlinking into a rebuildable Hermes venv.
+  `mnemosyne-hermes status` reports wrapper mode, target interpreter, import
+  health, and stale/broken targets.
+
 ## [3.10.1] — 2026-06-22
 
 ### Security
--- a/docs/hermes-integration.md
+++ b/docs/hermes-integration.md
@@ -41,15 +41,16 @@ cd mnemosyne
 pip install -e "integrations/hermes[dev]"
 ```
 
-> **Docker users:** Inside the official Hermes Docker container, the Hermes home directory is `/opt/data/` (the mounted volume), not `~/.hermes/`. After pip-installing `mnemosyne-hermes`, run the installer with the correct home path:
+> **Docker users:** Inside the official Hermes Docker container, the Hermes home directory is `/opt/data/` (the mounted volume), not `~/.hermes/`. For image-based or read-only Hermes venv installs, prefer the persistent wrapper mode so the plugin discovery directory survives image rebuilds:
 >
 > ```bash
-> # Inside the container
-> mnemosyne-hermes install --hermes-home /opt/data
+> # Inside the container, pointing at a side/persistent venv that has mnemosyne-hermes installed
+> mnemosyne-hermes install --hermes-home /opt/data --mode wrapper --python /path/to/venv/bin/python
+> mnemosyne-hermes status --hermes-home /opt/data
 > hermes gateway restart
 > ```
 >
-> The `install` CLI creates the plugin symlink that Hermes needs to discover the memory provider. Skip Steps 2-4 below; the installer handles everything. Run `hermes memory setup` after restarting to activate.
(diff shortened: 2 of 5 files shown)


AxDSan (dismissed):

@dplush wrapper install is the right shape for Docker/read-only Hermes venvs. The `$HERMES_HOME/plugins/mnemosyne/__init__.py` shim with `sys.path` injection keeps discovery stable when the image rebuilds the runtime venv, which is exactly the pain @bedpan called out in the issue thread.

What I verified:
- `py_compile` clean on the changed files in `integrations/hermes/src/mnemosyne_hermes/install.py` and the new test.
- `uv run --with-editable . python -m py_compile src/mnemosyne_hermes/install.py` -> clean.

LGTM with asks:

1. **CHANGELOG entry.** This is a new install mode (new behavior surface), so a one-line `[Unreleased]` entry under the next version block.
2. **Version bump.** New user-facing mode = MINOR bump. Current main is `3.10.1` from #373; this would land in `3.11.0`.

One quick sanity check that would be nice to see in the PR or tests: when `--mode wrapper` is used, `mnemosyne-hermes status` should report the wrapper target path and mark the link "stale" if the configured `--python` env no longer exists. The PR body says this is in the status output, but I want to confirm the stale-detection logic handles the case where the wrapper's `_SITE` path was moved or the venv was wiped between `install` and `status` calls.

Closing #355 cleanly. Bedpan's production confirmation in the issue thread is exactly the kind of evidence I want for new install modes.

dplush:
Added the CHANGELOG entry and bumped the target release to 3.11.0 after rebasing on current main / v3.10.1.

Local checks:
- `PYTHONPATH=.:integrations/hermes/src uv run --no-sync --with pytest pytest integrations/hermes/tests/test_install_status.py -q` → 8 passed
- `python3 -m py_compile integrations/hermes/src/mnemosyne_hermes/install.py mnemosyne/__init__.py` → passed

AxDSan (approved):
Re-reviewed the rebase diff. Clean — wrapper install mode is well-scoped, doesn't break existing install paths, and the status command makes it self-documenting. Approving.

AxDSan:
Hey @tvinagre, your PR content landed on main as part of the v3.11.0 release. Wrapper install mode for read-only / Docker deployments. Nice addition! Landed on main for v3.11.0. Closing this out. Appreciate the contribution!

AxDSan:

PR #368, opened 2026-06-20
fix: register host LLM backend before skip-context early return

tml89:
register_hermes_host_llm() was at the END of initialize(), after the skip-context early return. Cron/subagent/background sessions never reached it, so mnemosyne_sleep silently fell back to AAAK. Fix: move registration before the skip-context check. Also guard shutdown() to only unregister in non-skip contexts. Tested with simulated cron-context init: backend registered, llm_available()=True, _try_host_llm() returns text. Both __init__.py copies patched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Host LLM backend registration in skip-context sessions.**
+  `register_hermes_host_llm()` was called at the end of
+  `MnemosyneMemoryProvider.initialize()`, after the skip-context early
+  return. Cron, subagent, and background sessions never reached it, so
+  `mnemosyne_sleep` silently fell back to AAAK. Registration now fires
+  before the skip-context check; `shutdown()` only unregisters when the
+  session is not in a skip context. Also fixes the CLI standalone-loading
+  fallback import path for both `cli.py` copies (#368, supersedes #361).
+
 ## [3.10.0] — 2026-06-18
 
 ### Added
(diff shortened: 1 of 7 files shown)


AxDSan (approved):

@tml89 clean fix, and the superset relationship with #361 is exactly what I want to see on this board. Two related fixes from the same author, one strictly contains the other, no duplication, clean merge story.

What I verified on `pr/368`:
- `py_compile` clean on both provider inits and both cli.py copies.
- The ordering: `register_hermes_host_llm()` now fires before the skip-context early return in `MnemosyneMemoryProvider.initialize()`, so cron/subagent/background contexts (which previously fell back to AAAK) now get the host backend registered. Symmetric change in `shutdown()`: only unregister when the session is not in `self._skip_contexts`, so a skip-context session teardown does not kill the backend for the whole process.
- `pr/368` head includes `pr/361`'s commits (`3ec69e8` for the cli.py fallback and `f4d3738` for the regression test) plus the `__init__.py` ordering fix.

Recommendation: merge #368 first, close #361 as superseded. I already approved #361 on the assumption both could land, but since #368 is a strict superset with the cleaner merge commit, that path is better. The `_discover_plugins` doc nit I raised on #361 still applies: that prose should be tightened, but it lands via the merge of #368.

One ask before merge:

**CHANGELOG entry.** No `[Unreleased]` line for either fix. These are correctness fixes with no API change, so no version bump required, but a one-line note that host LLM registration now fires in cron/skip-context paths is worth recording.

LGTM on #368. Will close #361 as superseded after #368 lands.

tml89:
Thanks for the review. Both asks are addressed in commit 35627ea:

- **CHANGELOG**: Added an `[Unreleased] → Fixed` entry covering the host LLM registration ordering fix and the CLI standalone-loading fallback import path, with a note that #368 supersedes #361.
- **`_discover_plugins` doc nit**: Fixed both references in `tests/test_cli_host_llm_standalone_loading.py` — `discover_plugin_cli_commands` → `discover_plugins` (with path to `mnemosyne/core/plugins.py`), matching the actual loader method name.

#361 closed as superseded, per your recommendation.

AxDSan:
@tml89 Quick note — the host LLM registration fix from this PR shipped in #378. The close didn't leave an explanation, which I realize is frustrating especially coming after #361 being superseded.

Appreciate the thorough diagnostic work on both. The RCA was solid.

PR #395, opened 2026-07-02
fix(hermes): resolve profile bank when CLI is loaded standalone

TurgutKural:
## Problem

Hermes may load plugin CLI modules directly from a file path using `importlib.util.spec_from_file_location()`. In that context, the module may not have package metadata for relative imports.

`_resolve_cli_bank()` used a relative import for `MnemosyneMemoryProvider`. When loaded standalone, that import raised `ImportError`, was caught by the broad fallback, and returned `None`. CLI commands then used the default/legacy bank instead of the active profile bank.

This breaks the profile-isolation-aware CLI behavior added for Hermes profiles.

## Root Cause

`mnemosyne_hermes/cli.py` `_resolve_cli_bank()` calls `from . import MnemosyneMemoryProvider`. When loaded via `spec_from_file_location()` (as Hermes plugin loader does), `__package__` is not set, the import fails, and a broad `except Exception: return None` swallows the failure. Subsequent code falls back to `BeamMemory(session_id="hermes_default")`, which uses the default/legacy bank path instead of the active profile bank.

## Fix

Add an absolute-import fallback helper `_get_provider_class()` to match the existing fallback pattern used by `hermes_llm_adapter`. The helper is now called at the top of `_resolve_cli_bank`.

## Tests

- Package import path resolves profile bank (existing)
- Standalone `spec_from_file_location()` path resolves profile bank
- Explicit `--bank` still wins
- Root/default profile still falls back to default bank
- Import command keeps existing provider-bank semantics

## Compatibility

No config or migration changes. Runtime provider behavior is unchanged. Only CLI bank resolution changes when the plugin CLI is loaded standalone.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/cli.py
+++ b/integrations/hermes/src/mnemosyne_hermes/cli.py
@@ -82,6 +82,23 @@ def _profile_isolation_enabled(hermes_home: str) -> bool:
     return bool(val)
 
 
+def _get_provider_class():
+    """Return MnemosyneMemoryProvider, tolerating standalone plugin loads.
+
+    Hermes may load plugin CLI modules directly from a file path using
+    ``importlib.util.spec_from_file_location()``. In that context the module
+    has no parent package, so relative imports raise ``ImportError``. Fall
+    back to the absolute import — which works as long as the package is on
+    ``sys.path`` (true for both pip installs and Hermes-managed plugin
+    symlinks).
+    """
+    try:
+        from . import MnemosyneMemoryProvider
+    except ImportError:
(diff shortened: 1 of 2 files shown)


dplush:
Nice fix, thanks. I ran the focused CLI bank tests locally with the needed deps and they pass.

Tiny non-blocking test hygiene note: `test_standalone_load_via_spec_resolves_profile_bank` leaves `HERMES_HOME` set to the temp path if it was unset before the test. Might be worth using `monkeypatch.setenv(...)` there, or popping it again when the saved value is `None`.

Not a runtime concern, just avoids possible order-dependent test pollution.

PR #401, opened 2026-07-02
Add query/document embedding prompt prefix env vars

daefron-cmd:
## Why
Several recommended models require asymmetric prompt prefixes (multilingual-E5: `query: `/`passage: `; EmbeddingGemma: `task: search result | query: ` / `title: none | text: `). The client currently sends raw text on both paths, so these models run off-label. Also, `embed([single_text])` delegates to `embed_query()`, which would stamp a query prefix onto a stored document once prefixes exist.

## What
- `MNEMOSYNE_EMBEDDING_QUERY_PREFIX` / `MNEMOSYNE_EMBEDDING_DOC_PREFIX` (default empty = no behavior change), applied verbatim (trailing whitespace preserved) in `embed_query()` / `embed()` for both the API and fastembed paths.
- Removed the single-text delegation in `embed()`.
- Byte-exact tests against a stub HTTP server, including the single-document regression.

Note: tests cover both branches — byte-exact wire assertions for the API path and a fake-model test proving the prefixes reach the fastembed path's .embed() call.

### DIFF
--- a/mnemosyne/core/embeddings.py
+++ b/mnemosyne/core/embeddings.py
@@ -68,6 +68,28 @@ _DEFAULT_MODEL = os.environ.get("MNEMOSYNE_EMBEDDING_MODEL", "BAAI/bge-small-en-
 _embedding_model = None
 _API_CALL_COUNT = 0
 
+# (1) Prefix support — read at call time so env changes and test fixtures take effect
+# without a module reload. The _PREFIXES_LOGGED guard suppresses log spam.
+_PREFIXES_LOGGED = False
+
+
+def _get_prefix(kind: str) -> str:
+    """Model prompt prefixes (e.g. E5 'query: '/'passage: ', EmbeddingGemma retrieval
+    prompts). Applied VERBATIM — no trimming, no separator magic — because trailing
+    whitespace is part of the trained prompt for several models."""
+    var = ("MNEMOSYNE_EMBEDDING_QUERY_PREFIX" if kind == "query"
+           else "MNEMOSYNE_EMBEDDING_DOC_PREFIX")
+    prefix = os.environ.get(var, "")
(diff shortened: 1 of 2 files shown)


AxDSan:
Clean, focused PR. The verbatim-prefix approach is exactly right for asymmetric models where trailing whitespace is part of the trained prompt.

What I verified:
- Read at call time from env (not module-level frozen — hot-reloadable)
- Cached inner function keyed on prefixed text prevents stale vectors if prefix changes mid-process
- Dual-wired through both embed_query and embed_doc paths
- Log suppress guard (_PREFIXES_LOGGED) prevents spam

The right call. Merge-ready once CR checks pass and CHANGELOG entry is in.

AxDSan:
@daefron-cmd CI is failing on the fastembed prefix test. The issue is likely the 3.11 fastembed path check. Can you look at the failing test output? The design is approved (verbal LGTM from AxDSan), just needs the CI fix.

AxDSan:
@daefron-cmd good work on the embedding prefix feature. The CI failure is a single test issue.

**Root cause:** `test_fastembed_path_applies_prefixes` calls `embed(["local fact"])` which returns `None` in CI because the fastembed model (BAAI/bge-small-en-v1.5) cannot be downloaded in the CI environment. The `MNEMOSYNE_NO_EMBEDDINGS=1` flag is set in CI, and `embed()` returns `None` when no model is available. The test asserts `is not None`.

**Fix needed:** The test should either:
1. Skip when `MNEMOSYNE_NO_EMBEDDINGS=1` is set (the standard pattern for embedding-dependent tests in CI), or
2. Mock the embedding call to return a known vector

The API path tests in the same file likely work fine because they go through the API provider. Only the fastembed path needs the guard.

Want me to push the fix to your branch? Or prefer to handle it yourself?

PR #404, opened 2026-07-03
fix(mnemosyne): CLI bank selection, bank list, and 3.11.1 schema migration (Issues 1, 2, 3)

codxt:
## Summary

Three test-first fixes for three mnemosyne-oss issues found by
an external consumer of multi-tenant mnemosyne banks. All three
fixes have failing tests committed first; the fixes make them
pass.

### Issue 1: `mnemosyne store` ignores `MNEMOSYNE_BANK` env var
- **mnemosyne/cli.py**: `_get_memory()` now reads `MNEMOSYNE_BANK` and uses `BankManager.get_bank_db_path(bank)` when set, falling back to the default bank at `<DATA_DIR>/mnemosyne.db`. The MCP server honors `--bank` via the Dockerfile CMD; the CLI now matches.
- **tests/test_issue_1_cli_bank_selection.py**: subprocess-based failing test that pre-creates the bank DB at `<DATA_DIR>/banks/<bank>/mnemosyne.db`, sets `MNEMOSYNE_BANK=<bank>`, runs `mnemosyne store`, and asserts the row landed in the bank DB (not the default). Includes a control test that confirms the default behavior is preserved when `MNEMOSYNE_BANK` is unset.

### Issue 2: `mnemosyne bank list` reports phantom "default"
- **Final implementation:** leave `BankManager.list_banks()`'s virtual-`default` contract intact (existing core API/tests rely on it) and suppress the phantom `default` only in the operator-facing CLI output (`cmd_bank list`) when the default bank file does not exist on disk. This fixes the operator confusion without breaking the core `BankManager` API.
- **tests/test_issue_2_bank_list_disk_state.py**: three subprocess-based tests for the three cases (no file → no default in CLI output, file exists → default listed, bank subdirs exist → they are listed).
- **tests/test_cli_errors.py**: one assertion in the existing `test_bank_cli_list_create_delete_uses_configured_data_dir` was asserting the old buggy CLI behavior. Updated to assert the new correct CLI behavior. The rest of the test (create/list/delete cycle) is unchanged.
- **Core regression note:** the existing `BankManager` tests in `tests/test_memory_banks.py` continue to pass. The earlier CI failure was caused by a first implementation that changed the core API instead of just the CLI output; the fix is now correctly scoped to the CLI layer.

### Issue 3: existing banks at older mnemosyne schemas do not auto-migrate to the 3.11.1 schema (missing `memory_events` and `sync_meta`)
- **mnemosyne/migrations/e7_311_tables.py**: new migration module that adds the two tables introduced in mnemosyne-memory 3.11.1 to an existing bank at the older 54-table schema. The DDL is **copied verbatim from the canonical source at `mnemosyne/core/sync.py:641-677`** (`SyncManager._init_events_table`). This migration does **not** invent DDL. Idempotent (uses `CREATE TABLE IF NOT EXISTS` + try/except for indices, matching upstream behavior). Does not delete data or drop tables.
- **mnemosyne/cli.py**: new `cmd_migrate` subcommand that runs the migration against a bank. Bank selection: `--bank` flag, else `$MNEMOSYNE_BANK`, else the default bank. Reports tables added, tables already present, and indices added.
- **tests/test_issue_3_311_migration.py**: unit tests that
  - create a bank at the older 54-table schema (`init_db` + `DROP TABLE IF EXISTS` for the two new tables),
  - run `migrate_311_tables`; assert the two tables are added and the other tables are preserved (no data loss),
  - run the migration twice; assert idempotency,
  - on a bank that already has the two tables, assert the migration is a no-op,
  - assert the new tables have the canonical columns from the upstream source.

## Test plan

All 92 tests in the affected regression set pass on the feature branch:
- 3 new issue tests (Issues 1, 2, 3)
- 7 existing CLI/bank/migration/canonical suites, including `tests/test_memory_banks.py`
- The pre-existing relative-import issue in `tests/test_cli_host_llm_standalone_loading.py` is unrelated to this change and was not part of the affected regression set

## Background

The CLI's `_get_memory()` always opened the default bank regardless of `MNEMOSYNE_BANK`. The CLI's `bank list` always included `default` even when `<DATA_DIR>/mnemosyne.db` did not exist on disk, misleading operators about the bank state. An existing bank at the older 54-table mnemosyne schema was missing the two new tables `memory_events` and `sync_meta` introduced in 3.11.1, and the migration was not added by the upgrade.

The two new tables' DDL is **referenced from the canonical source** (`mnemosyne/core/sync.py:641-677`, `SyncManager._init_events_table`), not invented. If the upstream source changes its DDL, this migration should be updated to match.

## Checklist

- [x] Tests added (test-first)
- [x] Tests pass
- [x] No invented DDL (DDL copied from canonical upstream source)
- [x] Regression test updated to match the corrected behavior
- [x] Existing CLI/bank/migration/canonical tests still pass
- [ ] (upstream) Consider whether the migration should auto-run on `init_db` (like the e6 pattern) or remain explicit; the PR leaves it explicit per the test design.

This PR strengthens Mnemosyne’s **local-first, per-bank isolation** across CLI memory operations, sync, and operator-facing tooling—without changing the underlying tiered-memory contents or retrieval/consolidation logic.

## Core architecture & local-first guarantees
- **Per-bank database routing in CLI:** CLI commands that construct the Mnemosyne instance via `_get_memory()` now **honor `MNEMOSYNE_BANK`** and create `Mnemosyne(db_path=BankManager.get_bank_db_path(bank), bank=bank)` rather than always using the default database file. This applies to typical memory flows (e.g., `store`, `recall`, `update`, `delete`, `stats`, consolidation via `sleep`, plus `export/import`, and the sync commands that run SyncEngine against the local DB).
- **No misleading “default” bank in `bank list`:** `mnemosyne bank list` filters out the virtual `"default"` entry when the on-disk legacy DB (`<DATA_DIR>/mnemosyne.db`) doesn’t exist—so operators don’t get an output that doesn’t match actual local storage. The underlying `BankManager` behavior remains unchanged.

## Sync layer & long-term schema maintainability
- **3.11.1 migration for incremental sync primitives:** Adds an idempotent migration (`mnemosyne migrate [--bank <name>]`) that ensures existing banks have the sync-critical tables:
  - `memory_events` (event log used by the sync engine for tracking memory mutations)
  - `sync_meta` (persisted sync state such as device identity/cursors)
- **Canonical DDL + safe re-runs:** The migration uses upstream/canonical DDL (not bespoke definitions) and relies on `CREATE TABLE IF NOT EXISTS` (plus best-effort index creation), making it safe to run repeatedly and for future schema compatibility.
- **Sync server alignment with bank isolation:** `mnemosyne sync`, `sync-serve`, and `sync-status` all build their `SyncEngine`/server using the bank-aware Mnemosyne instance from `_get_memory()`, so the server’s push/pull bookkeeping occurs in the selected bank DB (not the default).

## Agent integration surfaces (Hermes/MCP/CLI)
- **CLI + sync surfaces are now bank-consistent:** Because `sync*` CLI commands route through the bank-aware Mnemosyne instance, any agent workflows that call into these CLI sync endpoints inherit correct per-bank isolation automatically.
- **MCP remains bank-selectable and aligned conceptually:** The MCP server supports a default bank via `--bank` (wired to `MNEMOSYNE_MCP_BANK`) and can resolve per-call `bank` from tool arguments; MCP tool handlers instantiate `Mnemosyne(bank=...)`, preserving the intended multi-tenant/domain-separation boundary for agent-driven memory access.

## Privacy posture
- **No new data sharing; reduced risk of cross-bank mixing:** The changes primarily prevent accidental reads/writes of sync metadata and memory event logs into the wrong local database by enforcing the intended bank at the storage boundary.
- **Migration is local-only:** The schema update only adds required sync tables to existing local SQLite files; it does not alter the content model or introduce outbound behavior.

## What did *not* change
- The PR does **not** redesign tiered memory (working vs episodic vs BEAM), retrieval strategies, consolidation semantics, veracity logic, or the export/retrieval algorithms. It focuses on correctness of **banked persistence** and the **sync schema** needed for stable incremental replication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+### Fixed
+
+- **CLI memory commands honor `MNEMOSYNE_BANK`.** `mnemosyne store`
+  and other `_get_memory()`-backed CLI commands now write to the selected
+  bank database instead of always using the default database.
+
+- **`mnemosyne bank list` no longer reports a phantom default bank.**
+  The CLI suppresses `default` when `<DATA_DIR>/mnemosyne.db` does not
+  exist, while preserving the core `BankManager` virtual-default contract.
+
+- **Existing 3.11.0-era banks can be migrated to the 3.11.1 sync schema.**
+  Added an explicit `mnemosyne migrate [--bank <name>]` command that
+  idempotently ensures `memory_events` and `sync_meta` exist.
+
+
 ### Changed
 
 - **Default prompt context excludes consolidated working-memory rows.**
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -58,10 +58,25 @@ def _parse_int(value: str, name: str) -> int:
         _fail(f"{name} must be an integer: {value}")
 
 
+def _resolve_bank_name(bank_override: str | None = None) -> str:
+    """Resolve the selected memory bank from an override or MNEMOSYNE_BANK."""
+    value = bank_override if bank_override is not None else os.environ.get("MNEMOSYNE_BANK", "")
+    return value.strip() or "default"
+
+
 def _get_memory():
-    """Get a Mnemosyne v2 instance."""
+    """Get a Mnemosyne v2 instance, honoring MNEMOSYNE_BANK."""
     from mnemosyne.core.memory import Mnemosyne
-    return Mnemosyne(db_path=os.path.join(DATA_DIR, "mnemosyne.db"))
+    from mnemosyne.core.banks import BankManager
+
+    bank = _resolve_bank_name()
+    bm = BankManager(Path(DATA_DIR))
+    try:
+        db_path = bm.get_bank_db_path(bank)
+    except ValueError as e:
+        _fail(str(e))
+
+    return Mnemosyne(db_path=str(db_path), bank=bank)
 
 
 def _resolve_default_scope() -> str:
@@ -612,6 +627,13 @@ def cmd_bank(args):
     try:
         if subcmd == "list":
             banks = bm.list_banks()
+            # CLI output should reflect on-disk state: if the default
+            # bank file does not exist yet, suppress the phantom
+            # virtual 'default' entry here while keeping BankManager's
+            # core virtual-default contract unchanged.
+            default_db = bm.get_bank_db_path("default")
+            if not default_db.exists():
+                banks = [b for b in banks if b != "default"]
             print("\nMemory Banks:\n")
             for b in banks:
                 print(f"  - {b}")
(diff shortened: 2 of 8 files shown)


AxDSan:
Solid writeup. Three scoped fixes, test-first, clean implementation.

What I verified:
- MNEMOSYNE_BANK env var routing in CLI matches BankManager contract
- Bank list phantom-default suppression is the right edge case handling
- e7 migration is idempotent (CREATE TABLE IF NOT EXISTS)

Two asks before merge:
1. Rebase on current main — conflicts from recent merges (batch tool, hermes config)
2. CHANGELOG entry under [Unreleased]

The right shape for all three issues.

codxt:
Rebased on current main and force-pushed the updated branch.

Resolved the cli.py conflict by preserving current main behavior and re-applying the scoped PR changes:
- MNEMOSYNE_BANK-aware CLI DB selection
- bank list phantom-default suppression
- explicit migrate command
- CHANGELOG entry under [Unreleased]

Also applied the useful CodeRabbit correctness point: CLI-created Mnemosyne instances now receive both db_path and bank, so self.bank reflects the selected bank.

Local validation:
python3 -m pytest \
  tests/test_issue_1_cli_bank_selection.py \
  tests/test_issue_2_bank_list_disk_state.py \
  tests/test_issue_3_311_migration.py \
  tests/test_cli_errors.py \
  tests/test_memory_banks.py

Result: 48 passed.

AxDSan:
@codxt CI is now fully green (all 4 Python versions passed, CodeRabbit approved). The conflict is from the 6 PRs that just landed on main. Can you rebase? The CLI bank selection and schema migration are ready to go once the conflicts are resolved.

codxt:
Rebased again on latest main and force-pushed.

Resolved the new cli.py conflict by preserving current main behavior and re-applying the scoped PR changes:
- MNEMOSYNE_BANK-aware CLI DB selection
- bank list phantom-default suppression
- explicit migrate command
- CHANGELOG entry under [Unreleased]

Local focused validation:
python3 -m pytest \
  tests/test_issue_1_cli_bank_selection.py \
  tests/test_issue_2_bank_list_disk_state.py \
  tests/test_issue_3_311_migration.py \
  tests/test_cli_errors.py \
  tests/test_memory_banks.py

AxDSan:
Clean work @codxt. Three fixes, all green, rebased twice. Merged.

CLI bank selection, bank list, and schema migration all landed. Thanks for the patience on the rebase treadmill.

PR #415, opened 2026-07-05
fix: validate bank names before path operations

dplush:
## Summary

- validate bank names before path-based BankManager operations resolve filesystem paths
- reject empty bank names consistently while preserving the explicit `default` special case
- cover delete/exists/path/stats/rename with invalid bank names and specific validation errors
- assert invalid traversal-shaped names do not affect sibling directories

## Tests

- `uv run --no-sync --with pytest pytest tests/test_memory_banks.py -q` → 30 passed
- `python3 -m py_compile mnemosyne/core/banks.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of bank names across deletion, existence checks, database path resolution, stats lookup, and renaming.
  * Prevents invalid or path-like names from being routed to legacy paths or triggering filesystem actions outside the managed data area.
* **Tests**
  * Added coverage to ensure invalid bank names/paths (e.g., empty, traversal, slashes, dotted names) are rejected with `ValueError` and do not alter or remove files outside the app’s data directory.

PR #416, opened 2026-07-05
fix: prefer current-state memories in now queries

dplush:
## Summary

- add a narrow current-state tie-breaker for recall queries that ask what is true now/currently
- only applies the bonus when the candidate memory itself carries an explicit current/latest/updated marker
- keeps the fix scoped to the failing current-state regression without changing the broader ranking pipeline

## Context

This addresses the CI failure seen on unrelated PRs (#414 and #415):

`tests/test_recall_precision_regressions.py::TestRecallPrecisionRegressions::test_current_state_query_prefers_newer_correction_over_stale_history`

The stale and corrected Project Atlas memories were nearly tied, and the stale legacy-cluster row could win due to a slightly higher lexical/FTS score. The added bonus is bounded and gated by both query intent and candidate content markers.

## Tests

- `uv run --with pytest --with numpy python -m pytest tests/test_recall_precision_regressions.py -q` → 15 passed, 4 subtests passed
- `uv run --with pytest --with numpy python -m pytest tests/test_query_intent_recall.py tests/test_temporal_recall.py tests/test_fact_recall_ranking.py tests/test_beam_e5_polyphonic_recall.py -q` → 45 passed
- `git diff --check`
- `python3 -m py_compile mnemosyne/core/beam.py`

## Summary by CodeRabbit

* **New Features**
  * Improved recall ranking for “current,” “now,” “latest,” and “updated” queries, helping more recent or state-related memories surface higher in results.
  * Added a small tie-breaker boost when matching content clearly indicates the current state, making retrieval more relevant for up-to-date questions.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1643,6 +1643,26 @@ def _minimum_recall_relevance(query_tokens: List[str]) -> float:
     return 0.15
 
 
+_CURRENT_STATE_QUERY_TOKENS = {"current", "currently", "latest", "now", "newer"}
+_CURRENT_STATE_CONTENT_TOKENS = {"current", "currently", "latest", "newer", "updated"}
+
+
+def _current_state_recency_bonus(query_tokens: List[str], content: str) -> float:
+    """Small tie-breaker for "what is true now/currently" queries.
+
+    Hybrid lexical scoring can leave stale and corrected facts nearly tied when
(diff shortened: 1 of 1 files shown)


dplush:
This should unblock the unrelated CI failures currently showing up on #414 and #415. The full matrix is green here, and the fix is scoped to the failing current-state recall regression.

PR #420, opened 2026-07-06
fix(plugins): enable auto_sleep by default in Hermes Mnemosyne plugin (#59836)

### LINKED ISSUE #59836: Hermes Mnemosyne plugin sets auto_sleep default to false, breaking cross-session memory consolidation
## Summary

The Hermes integration plugin for Mnemosyne (`integrations/hermes/src/mnemosyne_hermes/__init__.py`) sets `auto_sleep` to `false` by default (line 662), overriding the Mnemosyne core default of `true`. This means Mnemosyne never automatically consolidates working memory into episodic memory, effectively breaking cross-session memory retrieval for fresh Hermes+Mnemosyne installations.

A user can write memories, but recall across sessions never returns meaningful results because the episodic vector index stays empty.

## Root Cause

In `__init__.py`, the `get_config_schema()` method defines:
```python
{"key": "auto_sleep", "description": "...", "default": False},
```

Mnemosyne core documentation states the default should be `true` — consolidation runs on session start and end. But the Hermes plugin imposes `false` with no notice, no setup wizard prompt, and no mention in the installation docs.

## Impact

- Fresh install: memory appears to work but never consolidates
- Users discover it only when they notice the agent keeps forgetting cross-session context
- Workaround: manually set `memory.mnemosyne.auto_sleep: true` in config.yaml (and optionally `sleep_threshold: 20`)
- A cron job may still be needed for `mnemosyne sleep --all-sessions` to clean up historical unconsolidated data

## Expected

The Hermes plugin should either:
1. Default `auto_sleep` to `true` to match Mnemosyne core defaults, OR
2. Prompt during `hermes setup` or first-run to let the user opt in

## Environment

- Hermes version: current main branch (git-based install)
- Mnemosyne version: 3.10.0
- Platform: macOS
- Config: `memory.provider: mnemosyne`, no `auto_sleep` in default config

## Relevant Code

`integrations/hermes/src/mnemosyne_hermes/__init__.py:662`:
```python
{"key": "auto_sleep", "description": "Auto-run sleep() when working memory exceeds threshold. Set true to enable. Backward-compatible with MNEMOSYNE_AUTO_SLEEP_ENABLED env var.", "default": False},
```

`integrations/hermes/src/mnemosyne_hermes/__init__.py:352-353`:
```python
self._auto_sleep_threshold = 50
self._auto_sleep_enabled = os.environ.get("MNEMOSYNE_AUTO_SLEEP_ENABLED", "").strip().lower() in ("1", "true", "yes", "on")
```

webtecnica:
## Summary

The Hermes Mnemosyne plugin (`integrations/hermes/src/mnemosyne_hermes/__init__.py`) defines `auto_sleep` with `default: False` in `get_config_schema()`, overriding Mnemosyne core's default of `True`. This prevents automatic cross-session memory consolidation on fresh Hermes+Mnemosyne installations.

## Root Cause

Line 812 of `integrations/hermes/src/mnemosyne_hermes/__init__.py`:
```python
{"key": "auto_sleep", "description": "...", "default": False},
```

Mnemosyne core default is `True` — consolidation runs on session start and end. The Hermes plugin imposes `False` with no notice.

## Fix

Changed `"default": False` to `"default": True` on line 812.

## Impact

- **Before:** Fresh install: memory appears to work but never consolidates. Users discover it when the agent keeps forgetting cross-session context.
- **After:** auto_sleep defaults to `True`, matching Mnemosyne core behavior. Memory consolidates automatically.

Fixes NousResearch/hermes-agent#59836

## Summary by CodeRabbit

* **Bug Fixes**
  * Changed the default setting for automatic sleep behavior to be enabled, so new setups will now pause more reliably without requiring manual configuration.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/__init__.py
+++ b/integrations/hermes/src/mnemosyne_hermes/__init__.py
@@ -809,7 +809,7 @@ class MnemosyneMemoryProvider(MemoryProvider):
 
     def get_config_schema(self) -> List[Dict[str, Any]]:
         return [
-            {"key": "auto_sleep", "description": "Auto-run sleep() when working memory exceeds threshold. Set true to enable. Backward-compatible with MNEMOSYNE_AUTO_SLEEP_ENABLED env var.", "default": False},
+            {"key": "auto_sleep", "description": "Auto-run sleep() when working memory exceeds threshold. Set true to enable. Backward-compatible with MNEMOSYNE_AUTO_SLEEP_ENABLED env var.", "default": True},
             {"key": "sleep_threshold", "description": "Working memory count before auto-sleep triggers", "default": 50},
             {"key": "reflect", "description": "Reflection/sleep guardrails. Supports disabled_for_cron (default true) and max_calls_per_session (default 3; negative disables cap). Env: MNEMOSYNE_REFLECT_DISABLED_FOR_CRON, MNEMOSYNE_REFLECT_MAX_CALLS_PER_SESSION.", "default": {"disabled_for_cron": True, "max_calls_per_session": 3}},
             {"key": "vector_type", "description": "Vector storage type (note: not yet wired to BeamMemory at runtime; reserved for future use)", "choices": ["float32", "int8", "bit"], "default": "int8"},


dplush:
Nice catch. I think this needs the same default change in the root provider too.

Right now the PR updates only `integrations/hermes/src/mnemosyne_hermes/__init__.py`, while `hermes_memory_provider/__init__.py` still defaults `auto_sleep` to `False`.

That creates provider drift and is why the parity test fails:

`tests/test_hermes_provider_parity.py::test_provider_config_defaults_match`

Suggested minimal fix:
- set `auto_sleep` default to `True` in both provider surfaces
- keep the existing parity test as coverage for this drift class

CodeRabbit also raised a related point about the effective runtime/config default, so it may be worth checking both the schema default and the actual config-loading path together.

dplush:
Thanks — no worries, I'll pick this up from here.

I'll open a small follow-up PR that covers the missing provider parity piece and the effective runtime/config default, so `auto_sleep` is consistently enabled across both Hermes provider surfaces and fresh installs don't silently fall back to `False`.

dplush:
Follow-up PR is up here: #429

It covers the missing root-provider parity, the effective runtime/config default, env/config opt-outs, and regression tests across both provider surfaces.

AxDSan:
CodeRabbit flagged a runtime fallback mismatch: the auto_sleep runtime default (env var fallback empty string → falsy) doesn't match the schema default. The intent is auto_sleep ON by default, but at line 470 `os.environ.get("MNEMOSYNE_AUTO_SLEEP_ENABLED", "")` returns empty string which is falsy, so fresh installs boot with it off.

The fix: update the env fallback default to "true" so the runtime matches the schema — `os.environ.get("MNEMOSYNE_AUTO_SLEEP_ENABLED", "true")`. Could you push that? This is a real bug — the default was supposed to flip with this PR.

dplush:
#429 has been merged, so this PR is superseded now. Thanks for starting the fix!

AxDSan:
@webtecnica the config parity test is failing across all Python versions. The issue is the auto_sleep default mismatch — the test expects one value but the config provides another. Can you take a look at the failing test output and update? Happy to help debug if needed.

AxDSan:
Superseded by #429

AxDSan:
@webtecnica thanks for catching this. The auto_sleep default mismatch was real and your PR started the fix.

@dplush picked it up and shipped the full fix in #429 (merged), covering both provider surfaces and regression tests. Closing this as superseded by #429.

Appreciate you flagging it.

PR #425, opened 2026-07-07
fix: make host LLM timeout configurable

TurgutKural:
## Summary
- make the host LLM adapter timeout configurable via `MNEMOSYNE_HOST_LLM_TIMEOUT`
- keep the existing 15s default when the env var is unset
- add regression coverage for env-based host timeout configuration

## Why
Long host-adapter consolidation prompts can take longer than the previous hardcoded 15s cap. This lets deployments raise the per-attempt host timeout without changing code while preserving current default behavior.

## Tests
- `python3 -m pytest tests/test_local_llm.py -q` → 39 passed
- `python3 -m pytest tests/test_local_llm.py::TestHostLLMBackend::test_host_llm_timeout_can_be_configured_from_env -q` → 1 passed
- `python3 -m pytest -q` → collection failed in this local checkout because `mnemosyne_hermes` package metadata was not installed
- `PYTHONPATH=integrations/hermes/src python3 -m pytest -q` → 1806 passed, 21 skipped, 1 failed (`mnemosyne-hermes` distribution metadata not installed locally)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the host LLM timeout via an environment variable, allowing the timeout to be adjusted without code changes.

* **Tests**
  * Added coverage to verify the timeout value is correctly read from the environment.

### DIFF
--- a/mnemosyne/core/local_llm.py
+++ b/mnemosyne/core/local_llm.py
@@ -54,7 +54,7 @@ LLM_FALLBACK_API_KEY = os.environ.get("MNEMOSYNE_LLM_FALLBACK_API_KEY", "") or L
 HOST_LLM_ENABLED = os.environ.get("MNEMOSYNE_HOST_LLM_ENABLED", "false").lower() in ("1", "true", "yes")
 HOST_LLM_PROVIDER = os.environ.get("MNEMOSYNE_HOST_LLM_PROVIDER", "").strip() or None
 HOST_LLM_MODEL = os.environ.get("MNEMOSYNE_HOST_LLM_MODEL", "").strip() or None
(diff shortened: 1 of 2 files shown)


AxDSan:
CodeRabbit flagged one issue: the env-derived HOST_LLM_TIMEOUT at `mnemosyne/core/local_llm.py:57` needs safe parsing with a fallback so imports don't crash on bad config. Something like:

```python
try:
    HOST_LLM_TIMEOUT = float(os.environ.get("MNEMOSYNE_HOST_LLM_TIMEOUT", "15.0"))
except (ValueError, TypeError):
    HOST_LLM_TIMEOUT = 15.0
```

Could you push that fix? Otherwise the PR shape is good — configurable timeout is overdue.

PR #439, opened 2026-07-09
fix(beam): roll back the recall-touch transaction on failure

Milgauss:
## Problem

Follow-up to #432, same lock-hygiene family, different call site. After a transient write collision, every Mnemosyne write on the affected thread can start failing "database is locked" instantly (milliseconds, not after busy_timeout), continuing for minutes after the original writer has finished, then self-healing. Observed live on 2026-07-09: a 95-second storm of instant lock failures on `recall`/`remember` roughly half an hour after a consolidation pass completed.

## Mechanism

`get_context()`'s recall touch bumps `recall_count` and `last_recalled` inside an explicit `BEGIN TRANSACTION ... commit` with no error handling (`mnemosyne/core/beam.py:3783`). If any statement in that window fails — for example "database is locked" while a consolidation pass is writing — the exception propagates with the transaction still open on the long-lived thread-local connection.

From then on that connection is poisoned: it holds a read snapshot that predates any subsequent writer, so each later write attempt takes the stale-snapshot upgrade path and returns SQLITE_BUSY immediately, without consulting the busy handler. No busy_timeout setting helps. The condition persists until something resets the transaction, which is why the storms self-heal after minutes.

## Fix

Wrap the touch transaction with rollback-on-error so a transient collision costs one failed call instead of poisoning the connection. The rollback itself is guarded so a dead connection cannot mask the original exception.

## Verification

New regression test in `tests/test_sleep_lock_hygiene.py` (`TestRecallTouchRollback`): a failing commit inside the touch path must leave no open transaction, and the connection must accept writes immediately afterward. Red on unpatched main (the connection is left in a transaction and the follow-up write fails), green with the fix. The existing lock-hygiene suite still passes (4 tests).

This PR improves Mnemosyne’s core BEAM lock hygiene by introducing a shared `_guarded_transaction(conn)` helper that explicitly rolls back on any exception. It applies this to (1) the recall “touch” transaction inside `BeamMemory.get_context()` and (2) the cascading working-memory delete path inside `BeamMemory.forget_working()`, preventing transient lock/commit failures from leaving thread-local SQLite connections in a poisoned “open transaction / immediate SQLITE_BUSY” state.

Architecturally, this is the right call for local-first reliability: it contains concurrency/locking failures to the current operation without changing tiered-memory semantics (working vs episodic vs BEAM), retrieval strategy, consolidation pipeline, veracity logic, or any sync/agent integration surface (Hermes/MCP/CLI). It also strengthens long-term maintainability by centralizing the guarded-transaction pattern and adding rollback-on-error behavior where it matters most for long-lived connections.

The added regression test (`TestRecallTouchRollback`) verifies both that the failed recall touch leaves no open transaction and that the seeded row’s `recall_count` / `last_recalled` are unchanged, confirming rollback correctness, followed by successful subsequent writes and context retrieval.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -3780,14 +3780,28 @@ class BeamMemory:
             ts = new_last.isoformat()
             updates.setdefault(ts, []).append(row["id"])
 
-        cursor.execute("BEGIN TRANSACTION")
-        for ts, ids in updates.items():
-            placeholders = ",".join("?" for _ in ids)
-            cursor.execute(
-                f"UPDATE working_memory SET recall_count = recall_count + 1, last_recalled = ? WHERE id IN ({placeholders})",
-                (ts, *ids)
-            )
-        self.conn.commit()
+        # Roll back on ANY failure inside this explicit transaction. Left
+        # unprotected, a "database is locked" here (e.g. while a consolidation
+        # pass is writing) abandons the connection inside an open, stale
+        # transaction -- and because the connection is a long-lived thread
+        # local, EVERY later write on this thread then fails "database is
+        # locked" instantly (stale-snapshot upgrade; the busy handler is not
+        # consulted) until something resets it.
+        try:
+            cursor.execute("BEGIN TRANSACTION")
+            for ts, ids in updates.items():
+                placeholders = ",".join("?" for _ in ids)
+                cursor.execute(
+                    f"UPDATE working_memory SET recall_count = recall_count + 1, last_recalled = ? WHERE id IN ({placeholders})",
+                    (ts, *ids)
+                )
+            self.conn.commit()
+        except Exception:
+            try:
+                self.conn.rollback()
+            except Exception:
+                pass  # rollback of a dead connection must not mask the cause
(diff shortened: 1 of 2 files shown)


Milgauss:
Both review comments addressed in the latest commit:

- Added a `_guarded_transaction` context manager and ported both plain guarded-transaction call sites onto it (`get_context()`'s recall touch and `forget_working()`'s cascade delete). `_deferred_commits()` was deliberately left as is: it implements commit-deferral semantics behind a per-connection flag (suppressing nested commits, final commit on exit), which is a different mechanism from a plain guarded transaction — folding it into the shared helper would change its behavior. Happy to revisit if you'd prefer otherwise.
- The regression test now also asserts the failed touch was rolled back (the seeded row's `recall_count`/`last_recalled` are unchanged), not just that no transaction remains open.

Suites: `test_sleep_lock_hygiene.py` (4), `test_beam_e3_additive_sleep.py` + `test_consolidate_fact_concurrency.py` (34), and the forget/annotation selection (64) all pass, identical to unpatched main where applicable.

Issue #214, opened 2026-06-01
Undocumented config keys for multi-profile Hermes + typo/recall findings


bitr8:
**Use case:** Two-user household assistant on one Hermes instance (0.15.x, Mnemosyne 3.3.0). Both users interact via Telegram and web UI. One types fast on mobile with heaps of typos. Both need isolated memory — what one user tells the agent shouldn't bleed into the other's recall.

Turns out a few config keys that make this work aren't documented anywhere: `sync_roles`, `profile_isolation`, `bank_id`, and `skip_contexts`. Found them by reading the provider source. Figured I'd share what I hit and what worked in case it's useful.

## Undocumented config keys (config.yaml under memory.mnemosyne)

- **`sync_roles`** — controls which conversation roles get autosaved via `sync_turn`. We set `[user]` to stop assistant boilerplate from filling working memory. Default appears to be `["user", "assistant"]`.
- **`profile_isolation`** — memory namespace isolation per Hermes profile. Without it, both users' memories are visible to each other's recalls.
- **`bank_id`** — explicit bank selection. Without it, `_resolve_profile_bank()` derives bank names from `agent_identity` and `hermes_home`, which resolved to "default" for both our profiles. Setting it explicitly fixed the cross-contamination.
- **`skip_contexts`** — set of session contexts that skip memory init (defaults include `cron`, `flush`, `subagent`, `background`, `skill_loop`). We needed to know this existed to add `webhook` for our n8n-triggered sessions.

## Default mismatches between docs and code

- `sleep_threshold`: `docs/configuration.md` says 20, provider source (`__init__.py` line 637) says 50
- `auto_sleep`: `docs/configuration.md` says default true, provider code defaults to false

## `hermes mnemosyne doctor` doesn't check bank DBs

If you use `bank_id`, doctor reports against the root `mnemosyne.db` and shows 0 memories even when the bank DB has hundreds. Took a while to figure out the agent was working fine — doctor was just looking at the wrong database.

## Typo problem with embeddings and FTS

User types "checlk your memeory health" on mobile. That goes into working memory as-is, so `mnemosyne_recall(query="check memory health")` doesn't find it — embeddings diverge, FTS keywords don't match. We ended up patching `sync_turn` to run a SymSpell corrector (edit distance 1, protected-term whitelist) before storage. Not suggesting this should be upstream, but wanted to flag that typo-heavy input degrades recall quality.

## Config guide

[...]

PR #448, opened 2026-07-12
fix(hermes): make doctor respect resolved bank

TurgutKural:
## Summary

Hermes resolves the active profile bank before dispatching Mnemosyne CLI
commands, but the `doctor` branch did not honor that resolved bank, so
`doctor` inspected the profile-root metadata database instead of the active
production bank.

## Fix

- Add bank selection support to the `doctor` command.
- Pass the resolved bank into `run_diagnostics()`; `run_diagnostics()` now
  returns `resolved_bank` and `resolved_db` in its top-level summary dict
  (was only logged into `entries`), so the CLI prints the real DB path for
  both default and named-bank runs.
- Unknown-bank validation runs **before** Mnemosyne initializes, using a
  side-effect-free existence check (`bank_exists_read_only`) that does NOT
  create the parent `banks/` directory or the named-bank directory/DB.
- Preserve existing default-bank behavior.

## Regression coverage

- Default profile resolves the default database
- Named profile (`profile_isolation`) resolves its named bank
- Explicit `--bank` selection resolves the requested bank
- Bank name is sanitized via the same resolver used elsewhere
- Empty profile-root metadata DB does not override the named bank
- Unknown bank returns a clear error and creates no filesystem artifacts
  (no `banks/` dir, no named dir, no `mnemosyne.db`)
- Malformed bank name returns a clear error and creates no artifacts
- `resolved_bank` and `resolved_db` appear in terminal output
- Existing stats/sleep routing is unaffected
- No schema/data mutation (working-memory row counts unchanged)

## Tests

- `15/15` focused doctor-routing regression tests passed (including a
  strengthened sleep dry-run test that asserts the resolved named-bank beam
  is bound and `beam.sleep(dry_run=True)` is invoked on it)
- `40/40` core bank unit tests passed (including the
  `bank_exists_read_only` env-override test, which now relies solely on
  `monkeypatch.setenv`, no manual module-global mutation)
- Tests are located under `tests/` and are collected by the root CI command
- Existing stats/sleep bank routing remains covered

Command:

```bash
MNEMOSYNE_NO_EMBEDDINGS=1 \
pytest tests/test_hermes_doctor_bank_routing.py -v
```

## Known limitation: shell-exit-code propagation

The handler correctly returns a non-zero integer for missing or malformed
banks. The installed `hermes` launcher currently discards integer handler
return values, so the shell may still observe status 0.

This is a Hermes CLI dispatcher issue tracked upstream in
NousResearch/hermes-agent#62810, with the general fix under
NousResearch/hermes-agent#43645. A Mnemosyne-specific `SystemExit`
workaround is intentionally out of scope for this PR.

## Compatibility

The routing change does not mutate memory, schema, or configuration.
Unknown named banks are rejected before Mnemosyne/diagnostic database
initialization. Existing doctor dependency auto-fix behavior is unchanged.

No database schema, migration, storage, production package, or production
configuration changes.

References #214


TurgutKural:
CI note: the Python 3.13 matrix job currently fails on three tests that also fail unchanged on clean upstream `main` at `493c099`:

- `tests/test_beam.py::TestProviderContextSafety::test_subagent_context_does_not_initialize_or_write`
- `tests/test_hermes_memory_provider.py::test_initialize_skips_for_non_primary_context`
- `tests/test_sync_roles.py::TestSyncRolesConfig::test_not_set_preserves_default`

The doctor-routing regression suite is collected by the root CI command and passes `13/13`:

```
MNEMOSYNE_NO_EMBEDINGS=1 pytest tests/test_hermes_doctor_bank_routing.py -v
```

Python 3.10–3.12 were cancelled by fail-fast after the 3.13 failure. No unrelated baseline or version changes are included in this PR.

TurgutKural:
Re: CodeRabbit doctor-bank review (commit 4a048a8, inline on cli.py 168-188)

> Update the doctor-only guard so bank_exists_read_only rejects missing banks only when the bank was explicitly supplied via --bank, while allowing an unmaterialized profile-derived fallback to proceed through normal lazy creation.

This is intentional for `doctor`. A profile-derived bank is still the resolved diagnostic target. Allowing a missing implicit bank through would construct `Mnemosyne(bank=...)` and silently materialize an empty bank, making a health check mutate the state it is supposed to inspect.

The guard therefore applies to both explicit and profile-derived named banks. I added a regression test (`test_profile_isolation_implicit_bank_missing_fails_cleanly` in tests/test_hermes_doctor_bank_routing.py) confirming that a missing implicit profile bank fails cleanly (exit 1, "Bank not found") without creating filesystem artifacts.

Re: non-ValueError escaping as tracebacks — applied. The preflight guard now catches `except Exception as e` and prints `Bank validation failed: <err>` with exit 1, so import/permission/probe runtime failures use the established clean error path instead of bubbling a raw traceback. Malformed/missing names still map to `Bank not found: <bank>` per the ValueError branch.

Re: docstring + focused unit tests for bank_exists_read_only / _validate_bank_name — applied in tests/test_memory_banks.py (9 new tests: existing/missing bank, no-dir-creation, explicit data_dir, MNEMOSYNE_DATA_DIR honor, malformed/over-length ValueError, default short-circuit, no FS creation).

TurgutKural:
The handler correctly returns a non-zero integer for missing or malformed banks. The installed `hermes` launcher currently discards integer handler return values, so the shell may still observe status 0.

This is a Hermes CLI dispatcher issue tracked upstream in NousResearch/hermes-agent#62810, with the general fix under NousResearch/hermes-agent#43645. A Mnemosyne-specific `SystemExit` workaround is intentionally out of scope for this PR.

PR #457, opened 2026-07-13
feat: write_approval gate for mnemosyne_remember and mnemosyne_batch

### LINKED ISSUE #456: [FEATURE]  write_approval support for mnemosyne_remember
## Problem / Motivation

`memory.write_approval: true` currently only gates the **legacy** memory tool (`tools.memory_tool`). When `memory.provider: mnemosyne` is active — which is the recommended provider for production use — the `mnemosyne_remember` tool writes directly to the Mnemosyne SQLite database without any approval check.

This means:
- Turning on `memory.write_approval` gives **no** protection when Mnemosyne is the active provider — the setting is silently ineffective
- Operators who want human-in-the-loop review for memory writes have no way to achieve it without switching off Mnemosyne
- Silent/autonomous Mnemosyne writes in background jobs (cron, delegation) can accumulate stale or incorrect facts with no review surface

## Proposed Solution

Add a `write_approval` check to the Mnemosyne plugin's write tools (at minimum `mnemosyne_remember`, and optionally `mnemosyne_update` / `mnemosyne_forget` / `mnemosyne_triple_add`).

### Behaviour when `memory.write_approval: true`:

1. **Intercept writes:** instead of writing directly to the Mnemosyne DB, stage the record to `<home>/pending/memory/<id>.json` using the same schema as the legacy memory tool:

   ```json
   {
     "id": "a1b2c3d4",
     "subsystem": "memory",
     "action": "add",
     "summary": "User prefers dark mode in all UI tools and terminals",
     "origin": "assistant_tool",
     "created_at": 1783970800.0,
     "payload": {
       "target": "user",
       "content": "User prefers dark mode in all UI tools and terminals"
     }
   }
   ```

2. **On approval** (called by `hermes memory approve <id>`), apply the staged payload to the Mnemosyne DB via the normal write path.

3. **On rejection**, delete the pending JSON file (same as the legacy flow).

4. **No config change needed** — reuse the existing `memory.write_approval` boolean since it's already in the `memory:` config block that Mnemosyne reads from (vs. introducing a new `mnemosyne.write_approval` key).

### What this makes possible:
- Users get a single `write_approval` toggle that works regardless of which memory provider is active
- Background/cron jobs that write memory are also subject to the gate when enabled

## Alternatives Considered

1. **Introduce a separate `mnemosyne.write_approval` config key** — More explicit but adds config surface area. Since Mnemosyne already sits under the `memory:` config block and reads `memory.provider`, reusing `memory.write_approval` is simpler and avoids user confusion.

2. **Use a Hermes hook (post_tool_call / pre_tool_call)** — The Mnemosyne plugin already provides `pre_llm_call`, `on_session_start`, and `post_tool_call` hooks. A pre-write hook could intercept and stage, but hooks don't have a standardised "stage then apply" return mechanism today. A direct check inside the tool handler is more reliable and easier to reason about.

3. **Do nothing (status quo)** — Leaves a gap where users who enable write_approval get a false sense of security. Memory writes from Mnemosyne bypass the gate silently with no warning or indication.

## Impact

- **Who benefits:** Anyone using Mnemosyne with `write_approval: true` — operators, multi-user setups, compliance-conscious users, and anyone running background/cron agents that write memory autonomously
- **Breaking change?** No — when `memory.write_approval: false` (the default), behaviour is identical to today. When `true`, writes are staged instead of immediate, which is the expected contract of write_approval. Users who explicitly rely on `write_approval: true` with Mnemosyne are already not getting the gating they expect, so this is a bugfix, not a breaking change `pending/memory/*.json`

## Willing to Contribute

- [ ] I am willing to submit a PR for this feature
- [x] I need guidance on where to start
- [ ] Just sharing the idea

AxDSan:
Closes #456

## What this does

When `memory.write_approval: true` is set in Hermes config.yaml, Mnemosyne write tools now stage records to `pending/memory/<id>.json` instead of writing directly to the SQLite database.

Previously, `write_approval` only protected the legacy `MEMORY.md` / `USER.md` file writes while Mnemosyne SQLite writes bypassed the gate entirely. Users who enabled `write_approval` expecting protection got none with Mnemosyne as the active provider.

## Changes

- **`_write_approval_enabled()`** helper reads `memory.write_approval` from Hermes config.yaml (not `memory.mnemosyne.write_approval`)
- **`_handle_remember()`** stages to `pending/memory/<id>.json` when gated
- **`_handle_batch()`** stages each operation individually when gated
- Read-only tools (`recall`, `get`, `stats`, etc.) never gated
- `sync_turn()` autosave never gated (infrastructural, not agent-initiated)
- Both provider copies updated (standalone + bundled)

## Behavior

| `write_approval` | Mnemosyne behavior |
|---|---|
| `false` (default) | Writes directly to SQLite (unchanged) |
| `true` | Writes staged to `pending/memory/<id>.json` |

The pending store uses the same schema as the legacy `write_approval.stage_pending()` so the existing `/memory pending` slash command can review and approve/reject staged records.

## Verification

- [x] Both provider copies compile clean
- [x] No change to read-only tools
- [x] No change to sync_turn autosave
- [x] Default behavior unchanged (write_approval defaults to false)

## Summary

Adds Hermes-wide `memory.write_approval` support to Mnemosyne write tools. When enabled, `mnemosyne_remember` and each operation in `mnemosyne_batch` are staged as pending memory records for approval; disabled or invalid configuration preserves direct SQLite writes.

## Architectural impact

- Protects Mnemosyne writes consistently with legacy memory-file writes using the existing pending-record schema and `/memory pending` workflow.
- Applies approval at the agent integration boundary while leaving read-only tools and `sync_turn()` autosave ungated.
- Updates both standalone and bundled Hermes provider implementations.
- Does not alter working, episodic, or BEAM tiering; retrieval, consolidation, veracity, synchronization, and benchmark behavior remain unchanged.
- Batch staging is per operation, improving reviewability and allowing individual approval decisions.

This is the right call for privacy and local-first guarantees: autonomous or background write requests no longer become durable local memory without explicit approval when the shared toggle is enabled. Pending records remain in the local Hermes directory, and no Mnemosyne-specific remote or service dependency is introduced.

The default remains local and immediate when approval is disabled. Because `sync_turn()` autosave is intentionally unchanged, some memory writes can still bypass the approval gate; this preserves existing behavior but leaves a narrower privacy boundary for automatic turn synchronization.

## Integration and maintainability

Hermes gains a single configuration surface without introducing provider-specific settings. MCP/CLI approval remains mediated through the existing pending-memory workflow rather than requiring new interfaces. Duplicating the implementation across standalone and bundled providers maintains distribution parity, though it increases future maintenance risk if the two copies diverge.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1602,6 +1602,37 @@ class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
 
         return None
 
+    def _write_approval_enabled(self) -> bool:
+        """Check whether memory.write_approval is enabled in Hermes config.
+
+        Reads the top-level ``memory.write_approval`` boolean from Hermes
+        config.yaml (NOT ``memory.mnemosyne.write_approval``). Defaults to
+        ``False`` when the key is missing or unparsable.
+
+        This gates ALL memory writes (both legacy MEMORY.md and Mnemosyne
+        SQLite writes) through a single boolean, so users get consistent
+        protection regardless of which provider is active.
+        """
+        hermes_home = getattr(self, "_hermes_home", None)
+        if not hermes_home:
+            return False
+        config_path = os.path.join(hermes_home, "config.yaml")
+        if not os.path.exists(config_path):
+            return False
+        try:
+            import yaml
+        except ImportError:
+            return False
+        try:
+            with open(config_path) as f:
+                config = yaml.safe_load(f) or {}
+        except Exception:
+            return False
+        memory = config.get("memory")
+        if not isinstance(memory, dict):
+            return False
+        return bool(memory.get("write_approval", False))
+
     def _configured_tool_schemas(self) -> List[Dict[str, Any]]:
         """Return schemas filtered by memory.mnemosyne.tools, if configured.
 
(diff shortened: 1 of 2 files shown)


AxDSan:
Closing as incomplete. #456 step 1 only. Step 2 (apply_pending) needed. Will ship complete fix tomorrow.

AxDSan:
@ExtYoSky you're right. #457 only implements step 1 (staging). Without step 2 (apply-on-approval), the staged records can never be committed to Mnemosyne. 

Closing this PR as incomplete. Your proposed `apply_pending()` design is the right approach: replay the staged payload through `apply_beam_batch()` on approval, not through `_handle_tool_call` (which would re-hit the write_approval gate).

Will ship the complete fix (staging + apply_pending) in a follow-up PR tomorrow. Thanks for the thorough review.

PR #470, opened 2026-07-15
fix(memory): restore beam initialization lifecycle

dplush:
## Summary

Restores the pre-#382 Phase 8 lazy-state and `BeamMemory` initialization to `Mnemosyne.__init__`. The WAL cleanup methods remain lifecycle-only.

The #382 insertion accidentally left this initialization under `__del__`, so normal instances had no `beam` attribute and the main CI failed.

## Verification

- Reproduced the current-main failure: `TestMnemosyneIntegration::test_legacy_and_beam_dual_write` raised `AttributeError: Mnemosyne has no attribute beam`.
- Added constructor + idempotent-close regression coverage.
- Focused affected matrix: `44 passed`.
- `py_compile` and `git diff --check` pass.

Note: the unrelated current-main polyphonic recall failure (`PolyphonicResult` has no `content`) remains outside this narrow regression fix.

## Summary

This PR corrects `Mnemosyne`’s core memory lifecycle by restoring Phase 8 lazy-subsystem placeholder setup and `BeamMemory` construction back into `Mnemosyne.__init__` (where it belongs), undoing an accidental relocation under `__del__` introduced previously. That ensures `BeamMemory` is created during construction with the intended event wiring (`event_emitter=self._stream_emit`), while `close()` remains focused on WAL-cleanup and idempotent shutdown. It also adds constructor-and-close regression coverage, including safe behavior across multiple wrappers and exception paths.

- **Core architecture (tiered memory / BEAM):** Re-establishes the intended BEAM tier initialization order by performing Phase 8 streaming/patterns/plugins/delta-sync wiring scaffolding during construction and then instantiating `BeamMemory` with the correct emitter. Removes duplicated/late Phase 8 initialization logic, preventing partially-initialized BEAM state from reaching runtime.
- **Connection lifecycle / long-term correctness:** Introduces instance-scoped ownership/refcounting around thread-local SQLite and BEAM resources so multiple `Mnemosyne` wrappers in the same thread (and/or across different banks/db paths) don’t double-close shared connections. `_close_connection`/`_close_beam_connection` now support targeted `db_path`-aware cleanup to avoid invalidating a newer live instance that switched banks.
- **Retrieval / consolidation / veracity:** No changes to retrieval strategies, consolidation pipeline, or the veracity system behavior are indicated; changes are architectural/lifecycle (initialization timing, correct event plumbing, and safe shutdown).
[...]

Issue #382, opened 2026-06-29
WAL checkpoint blocked after session ends — thread-local SQLite connections not cleaned up


kirocop:
## Description

When a cron job (or any non-primary session) writes to Mnemosyne and then terminates, its thread-local SQLite connection is not explicitly closed. Under WAL journal mode, this can leave the WAL in a state where `PRAGMA wal_checkpoint(TRUNCATE)` returns `busy=1` (cannot checkpoint), blocking ALL subsequent reads and writes to the memory database.

The only recovery is restarting the Hermes gateway (exit 75), which releases all 8+ file handles held by the gateway process on the DB file.

## Symptoms

- `mnemosyne_stats` returns `OperationalError: database is locked`
- Any Mnemosyne tool call (remember/recall) fails with "database is locked"
- `mnemosyne.db-wal` file does not shrink (stuck at several MB)
- `mnemosyne.db-shm` remains stale (same timestamp for hours)
- Gateway process holds multiple file descriptors on the DB (observed: 8 handles from PID 14574)
- `PRAGMA wal_checkpoint(TRUNCATE)` returns `1|<N>|<N>` (busy, pages not checkpointed)

## Root cause

1. **Thread-local connections** — `beam.py:_get_connection()` stores connections in `_thread_local.conn`. Each thread (main session, cron job, subagent) gets its own SQLite connection.

2. **WAL mode** — `PRAGMA journal_mode=WAL` + `PRAGMA busy_timeout=5000` is set on every connection (`beam.py:433-434`). Under WAL, a checkpoint is blocked while ANY connection has an open reader statement.

3. **No explicit `conn.close()` on thread exit** — When a session ends, the thread dies but the SQLite connection is garbage collected, not explicitly closed. SQLite may keep the WAL alive if prepared statements or reader handles remain.

4. **VeracityConsolidator creates its own connection** — `veracity_consolidation.py:286-302` opens a separate connection with its own WAL/busy_timeout PRAGMAs. This is another writer path not explicitly closed.

5. **No periodic checkpoint** — There is no `wal_checkpoint(TRUNCATE)` call anywhere in the maintenance cycle or `on_session_end`.

## Observed sequence

| Time | Event |
|---|---|
| 07:05 | Cron job starts (writes to Mnemosyne) |
| 07:07 | Last Mnemosyne write recorded |
| 07:12 | WAL checkpoint blocked — `mnemosyne.db-wal` stuck at 4.6 MB |
[...]

PR #472, opened 2026-07-15
fix(sqlite): isolate thread-local connections by database

dplush:
## Summary

Replacement for the unsafe #382 follow-up. Stores core and BEAM SQLite connections per canonical database path in thread-local maps, with per-path wrapper ownership so one live instance cannot close another database or same-database owner. Also restores the constructor initialization accidentally moved under `__del__`.

## Verification

- Same-path live-owner and different-path connection smoke passed.
- Focused integration suite: `4 passed`.
- `py_compile` and `git diff --check` passed.

Supersedes #470.

## Summary

- Reworked SQLite connection management into canonical per-thread, per-database caches.
- Added explicit wrapper ownership so closing one `Mnemosyne` or `BeamMemory` instance cannot close another live owner, including same-path instances.
- Restored guarded constructor initialization and added reliable cleanup via `close()` and best-effort `__del__()`.
- Preserved WAL, busy-timeout, foreign-key, and `sqlite_vec` setup while improving checkpoint and test isolation.
- Added integration coverage for same-path and different-path connection isolation.

## Architectural impact

This is the right call for Mnemosyne’s local-first core: it strengthens local SQLite reliability without introducing remote dependencies or weakening privacy guarantees. The change is limited to connection lifecycle management and does not alter working, episodic, or BEAM memory semantics, retrieval strategies, consolidation, veracity, synchronization, or benchmark methodology.

Hermes, MCP, and CLI integration surfaces remain behaviorally unchanged; they benefit indirectly from safer concurrent wrapper usage and deterministic cleanup. The per-path ownership model improves long-term maintainability by making resource ownership explicit and preventing cross-instance shutdown bugs.

Validation covered same-path/different-path smoke tests, a focused four-test integration suite, compilation, and whitespace checks.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -437,6 +437,27 @@ if VEC_TYPE not in ("float32", "int8", "bit"):
     VEC_TYPE = "float32"
 
 
+def _canonical_db_path(db_path: Path = None) -> tuple[str, Path]:
+    """Return a stable cache key and path for a database location."""
+    path = Path(db_path) if db_path else _default_db_path()
+    path = path.expanduser().resolve()
+    return str(path), path
+
+
+def _connections() -> dict[str, sqlite3.Connection]:
+    """Return this thread's BEAM connections, keyed by canonical path."""
+    if not hasattr(_thread_local, "connections"):
+        _thread_local.connections = {}
+    return _thread_local.connections
+
+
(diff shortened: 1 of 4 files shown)


dplush:
Superseded by #477: the per-DB cache direction exposed a larger lease/generation contract that should be designed separately. #477 restores the previously stable behavior first.

Issue #389, opened 2026-07-01
[BUG] Polyphonic recall collapses to a single result when one voice dominates (_estimate_similarity uses voice-name Jaccard)


NodeGuy:
## Summary

`PolyphonicRecallEngine._diversity_rerank` discards **all but one** result whenever a single voice dominates the candidate set. The cause is `_estimate_similarity`, which estimates the similarity between two candidates as the Jaccard overlap of their **voice names** (`voice_scores.keys()`) rather than their content. When only one voice is active — the normal case for a working-memory-only store, where the graph / fact / temporal voices have nothing to return — every candidate's voice-set is identical (e.g. `{"vector"}`), so every pair scores `1.0`, exceeds the `0.8` "too similar" threshold, and gets dropped as a duplicate.

Net effect: with `MNEMOSYNE_POLYPHONIC_RECALL=1`, `recall()` returns a single memory no matter how many relevant candidates the vector voice found.

## Environment

- `mnemosyne-memory` **3.10.1** (pip), with `[embeddings]` (fastembed `BAAI/bge-small-en-v1.5` + sqlite-vec).
- Confirmed still present on `main` (`HEAD`): `_estimate_similarity` is byte-identical there.

## Root cause

`mnemosyne/core/polyphonic_recall.py`:

```python
# _diversity_rerank (line ~737)
for sel in selected:
    similarity = self._estimate_similarity(result, sel)
    if similarity > 0.8:            # Too similar  (line ~760)
        is_diverse = False
        break

# _estimate_similarity (line ~769)
def _estimate_similarity(self, a, b) -> float:
    """Estimate similarity between two results."""
    # Simple Jaccard-like similarity on voice scores
    voices_a = set(a.voice_scores.keys())
    voices_b = set(b.voice_scores.keys())
    if not voices_a or not voices_b:
        return 0.0
    intersection = voices_a & voices_b
    union = voices_a | voices_b
    return len(intersection) / len(union)
```

Two genuinely different memories that were both surfaced by (only) the vector voice each have `voice_scores == {"vector": ...}`, so `_estimate_similarity` returns `1/1 = 1.0`. The diversity filter then treats the second, third, … as duplicates of the first.

## Minimal reproduction

Self-contained; no LLM required (`remember()` embeds at write time when `[embeddings]` is installed):

```python
import os, tempfile, sqlite3
os.environ["MNEMOSYNE_DB_PATH"] = os.path.join(tempfile.mkdtemp(), "repro.db")

[...]

PR #477, opened 2026-07-16
revert: unsafe #382 connection lifecycle change

dplush:
## Summary

Reverts `c527c40` (#382). The change accidentally moved Mnemosyne constructor initialization under `__del__`, breaking every normal instance with a missing `beam` attribute. Follow-up attempts exposed that safe multi-DB lifecycle support needs a defined lease contract, not incremental patching.

This restores the previously stable connection behavior. The WAL cleanup design will be revisited separately with an explicit lifecycle contract and test matrix.

## Verification

- Reproduced the #382 regression before rollback.
- Focused formerly failing Beam/CLI/allowlist matrix: `43 passed`.
- `py_compile` and `git diff --check` pass.

Supersedes #470 and #472.

## Summary

Reverts the lifecycle changes that caused normal `Mnemosyne` instances to lose their `beam` attribute, restoring stable core initialization and connection behavior.

## Architectural impact

- Restores the existing working, episodic, and BEAM memory architecture without changing tiering, retrieval, consolidation, veracity, or synchronization behavior.
- Removes the experimental thread-local connection shutdown and WAL checkpointing paths from BEAM and `Mnemosyne`.
- Preserves local-first operation by avoiding new remote dependencies or changes to local persistence.
- Keeps Hermes, MCP, and CLI integration surfaces behaviorally unchanged while restoring their access to properly initialized Mnemosyne instances.
- Simplifies the lifecycle implementation and avoids an implicit destructor-based contract; explicit multi-database lifecycle handling and WAL cleanup should be addressed separately with defined semantics and test coverage.

## Validation

- 43 Beam, CLI, and allowlist tests pass.
- `py_compile` succeeds.
- `git diff --check` succeeds.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -487,27 +487,6 @@ def _get_connection(db_path: Path = None) -> sqlite3.Connection:
     return _thread_local.conn
 
 
-def _close_beam_connection() -> None:
-    """Close the thread-local BEAM connection and checkpoint the WAL.
-
-    Thread-local connections under WAL mode can block checkpoints
-    after the owning thread exits.  Explicitly closing the connection
-    and running ``PRAGMA wal_checkpoint(TRUNCATE)`` prevents the
-    ``database is locked`` cascade documented in #382.
-    """
-    if hasattr(_thread_local, 'conn') and _thread_local.conn is not None:
-        try:
-            _thread_local.conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")
-        except Exception:
-            pass
-        try:
-            _thread_local.conn.close()
-        except Exception:
-            pass
-        _thread_local.conn = None
-        _thread_local.db_path = None
-
-
 def _detect_vec_type(conn: sqlite3.Connection) -> str:
     """
     Detect whether sqlite-vec supports int8/bit.
--- a/mnemosyne/core/memory.py
+++ b/mnemosyne/core/memory.py
@@ -77,45 +77,6 @@ def _get_connection(db_path = None) -> sqlite3.Connection:
     return _thread_local.conn
 
 
-def _close_connection() -> None:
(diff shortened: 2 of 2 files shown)


dplush:
Verification note: this revert removes the `Mnemosyne has no attribute beam` failure (0 occurrences in the PR CI logs). The remaining Python failures are only the separate #389 `PolyphonicResult.content` regression. Local focused Beam/CLI/allowlist coverage: 43 passed; CodeRabbit found no actionable issues.

dplush:
@AxDSan 
Context on the PR sequence:

- #382 introduced a regression where `Mnemosyne.__init__()` no longer initialized `beam`, causing the CI-wide `AttributeError`.
- #470 was opened as a narrow restoration fix. Review then exposed that the new cleanup logic could invalidate shared thread-local connections when multiple instances/databases were live.
- #472 explored per-database connection caching, but review correctly showed that this needs a properly designed lease/generation contract rather than incremental patches.
- To avoid merging an incomplete lifecycle redesign, #470 and #472 were closed.
- #477 is the safe rollback of #382: it restores the previously stable behavior and removes the `beam` regression. The remaining CI failures are the independent #389 polyphonic issue, addressed separately in #471.

The intent is to stabilize `main` first, then revisit WAL/connection cleanup as a separately designed change with clear ownership semantics and a full lifecycle test matrix.

—

Here is the request for the following procedure:

1. First, please merge #477
   - CodeRabbit is green; explanation/verification posted.
   - The red CI has been verified to affect only #389.

2. Immediately afterward, I’ll rebase #471 onto the new “main”
   - then the #382 base regression will disappear,
   - run the full matrix again,
   - if it’s green, #471 is merge-ready.

3. After that, I won’t touch Lifecycle/WAL again right away
   - only as a standalone, design-first topic with a lease contract and tests.

Issue #473, opened 2026-07-16
mnemosyne_diagnose ignores $HERMES_HOME and silently writes to ~/.hermes/mnemosyne/logs/


sky770825:
## Environment

- mnemosyne-memory: latest from PyPI (verified in `site-packages/mnemosyne/diagnose.py:22`)
- Python: 3.11 (macOS, also reproducible on Linux)
- Reproduced inside an isolated `HERMES_HOME=/tmp/fake-home` sandbox

## Reproduction

```bash
export HERMES_HOME=/tmp/fake-home
export MNEMOSYNE_DATA_DIR=$HERMES_HOME/mnemosyne/data
rm -rf $HERMES_HOME/mnemosyne
python -m mnemosyne.diagnose
ls $HERMES_HOME/mnemosyne/logs/      # empty — diagnose wrote elsewhere
ls ~/.hermes/mnemosyne/logs/         # new diagnose_*.jsonl appears ❌
```

Concrete output from a real reproduction (Hermes Agent audit, 2026-07-16):

```
[before] real_home diagnose logs: 2
[after]  fake_home logs: 0
[after]  real_home logs: 3 (before 2)
❌ BLOCKER #1 REPRODUCED — wrote to real ~/.hermes even though HERMES_HOME was set:
   /Users/caijunchang/.hermes/mnemosyne/logs/diagnose_2026-07-16_082729.jsonl
```

## Expected behavior

When `$HERMES_HOME` is set, `mnemosyne_diagnose` should write its JSONL log under `<HERMES_HOME>/mnemosyne/logs/` to match the convention used by the rest of the package:

| File | Line | Implementation |
|------|------|----------------|
| `core/banks.py` | 31 | `Path(os.environ.get("HERMES_HOME", Path.home() / ".hermes"))` ✅ |
| `core/embeddings.py` | 54 | `os.environ.get("HERMES_HOME", os.path.expanduser("~/.hermes"))` ✅ |
| `dr/recovery.py` | 39 | `Path(os.environ.get("HERMES_HOME", Path.home() / ".hermes"))` ✅ |
| `mcp_tools.py` | 116 | `os.environ.get("HERMES_HOME", str(Path.home() / ".hermes"))` ✅ |
| `diagnose.py` | 22 | **`Path.home() / ".hermes" / "mnemosyne" / "logs"`** ❌ hardcoded |

## Actual behavior

[...]

Issue #475, opened 2026-07-16
Transient embedding API failures are not retried consistently


mia-fourier:
## Description

Mnemosyne's OpenAI-compatible embedding path currently retries string-detected rate-limit errors, but transient network failures and HTTP 5xx responses can return `None` immediately. This leaves callers with silent embedding gaps when the embedding endpoint is temporarily unavailable.

## Environment

- **OS:** macOS
- **Python version:** 3.11
- **Mnemosyne version/commit:** upstream `main` at `0a64d2e`
- **Installation method:** editable source checkout

## Reproduction Steps

1. Configure `MNEMOSYNE_EMBEDDING_API_URL` for an OpenAI-compatible endpoint.
2. Have the endpoint return HTTP 503 once, followed by a valid embedding response.
3. Call `_embed_api`.

## Expected Behavior

Transient network errors, HTTP 429, and HTTP 5xx responses receive bounded retries with exponential backoff and jitter. Non-transient client errors such as HTTP 400 are not retried.

## Actual Behavior

The first HTTP 503 returns `None` without retrying.

## Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I can reproduce this bug consistently
- [x] I have included environment details

AxDSan:
Fixed in #478 (raitoxlol, merged Jul 16). Bounded retry with jitter for HTTP 429/5xx and transient network failures.

PR #478, opened 2026-07-16
Harden runtime dependencies, diagnostics paths, and embedding retries

raitoxlol:
## What changed

- Declare `PyYAML>=6.0` as a runtime dependency and update `uv.lock`.
- Resolve diagnostic logs under `HERMES_HOME` when it is set.
- Retry embedding API HTTP 429/5xx responses and transient network failures with bounded exponential backoff; permanent 4xx errors still fail fast.
- Add regression coverage for the dependency/path/retry behavior and document the fixes in the unreleased changelog.

## Why

Clean installs failed during first-run config initialization because the package imported `yaml` without declaring PyYAML. Diagnostics ignored `HERMES_HOME` and could write into a different Hermes installation. Transient embedding endpoint failures returned immediately, creating silent embedding gaps.

## Validation

- Focused regression tests: `13 passed`
- `python3 -m compileall`: passed
- `scripts/verify-docs.py`: passed (website docs skipped because the sibling checkout is unavailable)
- `python -m build`: passed
- Subprocess diagnostic check confirmed logs were written under the configured `HERMES_HOME`.

The full test suite was not run here; one existing diagnostics test fails because the current implementation closes a database connection that the test later reuses.


AxDSan:
Merged in 21ff279. @raitoxlol solid first contribution. Three fixes in one PR, each with focused tests.

- PyYAML dependency declared explicitly
- Diagnostics respect HERMES_HOME
- Embedding API retries with bounded backoff

Fixes #473 and #475. Welcome to the project.

Issue #474, opened 2026-07-16
embeddings.available() reports True while vec_episodes is never created — silent sqlite-vec fallback hides hybrid retrieval regression


sky770825:
## Environment

- mnemosyne-memory: latest from PyPI
- Python: 3.11 (macOS arm64, also reproducible on Linux)
- Verified in `site-packages/mnemosyne/core/embeddings.py:42` and `site-packages/mnemosyne/core/beam.py:480-485`
- Reproduced inside an isolated `HERMES_HOME=/tmp/fake-home` sandbox

## Reproduction

```bash
# fresh venv with both fastembed==0.8.0 and sqlite-vec installed
export HERMES_HOME=/tmp/fake-home  # any clean dir
export MNEMOSYNE_DATA_DIR=$HERMES_HOME/mnemosyne/data
python -c "
from mnemosyne.core import embeddings
from mnemosyne.core.memory import Mnemosyne
print('embeddings.available():', embeddings.available())   # True
m = Mnemosyne()
m.remember('fox jumps over lazy dog')
m.sleep(force=True)
import sqlite3; c = sqlite3.connect(m.db_path)
print(c.execute('SELECT COUNT(*) FROM vec_episodes').fetchone())  # OperationalError: no such module: vec0
"
```

Concrete output from a real reproduction (Hermes Agent audit, 2026-07-16):

```
=== Step 1: static availability ===
available()=True | frozen=True
=== Step 2: warm up ===
_get_model() -> TextEmbedding            # fastembed 0.8.0 really can load
embed(['hello']) shape = (1, 384)        # embed function also OK
=== Step 3: Mnemosyne round-trip ===
db_path: .../fake-home/mnemosyne/data/mnemosyne.db
=== DB schema check ===
  memories: 0 rows
  memory_embeddings: 0 rows
  working_memory: 0 rows
  episodic_memory: 0 rows
  vec_episodes: NO TABLE (no such module: vec0)   ← virtual table never created
  vec_working:  NO TABLE (no such module: vec0)
remember -> id=e615c34ec988ec1e            # remember() still returns OK
memories row: ('e615c34ec988ec1e', 'fox jumps over lazy dog')
memory_embeddings after 2 inserts: 2     # memory_embeddings has row (but FK errored)
```

## Expected behavior

[...]

PR #480, opened 2026-07-16
fix: restore main CI contracts after merge

dplush:
Restores main CI after the #456/#476/#478 merge sequence.

- add the missing apply-pending schema to the Hermes integration copy
- implement bounded exponential embedding-retry backoff with jitter
- align the provider count and replace tests for the reverted lifecycle API with stable constructor coverage

Validation: focused regression suite (82 passed), py_compile, git diff --check.

## Summary

- Restores Hermes agent integration compatibility by registering the missing pending-memory commit tool (`mnemosyne_apply_pending`, `APPLY_PENDING_SCHEMA`). This re-enables the approved “staged/approved pending writes” flow to be committed through the BEAM write path (i.e., it fixes the integration surface contract rather than changing memory semantics).
- Improves embedding-request resilience by switching embedding retry delays to **bounded exponential backoff with randomized jitter** for HTTP 429/5xx, network exceptions, and the message-text-based rate-limit fallback path—reducing retry stampedes without changing what data is sent.
- Adjusts BEAM/polyphonic diversity behavior expectations by computing diversity similarity from `PolyphonicResult.content` when present (falling back to `metadata["content"]`), updating the related regression assertion accordingly.
- Hardens long-term maintainability of the local-first/ownership lifecycle by replacing reverted lifecycle/disposal assertions with a stable “construction contract” test, and aligning tool-wiring tests to the updated exported tool count.

### Impact across architecture & guarantees
- **Core memory architecture / tiered memory (working/episodic/BEAM):** No tiered-memory logic changes in these diffs; the BEAM-related impact is limited to formalizing the expected pending-write commit tool wiring.
- **Retrieval strategies & consolidation/veracity/sync layers:** Not modified.
- **Privacy posture & local-first guarantees:** Low risk—`APPLY_PENDING_SCHEMA` commits existing staged/approved local pending records, and the embedding retry change alters timing only (not request payloads).
- **Agent integration surfaces (Hermes/MCP/CLI):** Hermes is directly fixed; MCP/CLI surfaces are not indicated as changed.
- **Validation:** 82 passing focused regression tests, `py_compile`, and `git diff --check`.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/tools.py
+++ b/integrations/hermes/src/mnemosyne_hermes/tools.py
@@ -772,12 +772,34 @@ SYNC_STATUS_SCHEMA = {
     },
 }
 
+APPLY_PENDING_SCHEMA = {
+    "name": "mnemosyne_apply_pending",
+    "description": (
+        "Commit staged pending memory writes to Mnemosyne. "
+        "When memory.write_approval is enabled, calls to mnemosyne_remember "
+        "and mnemosyne_batch are staged to pending/memory/ instead of "
+        "written directly. This tool replays approved pending records "
+        "through the BEAM write path, committing them to the database."
+    ),
+    "parameters": {
+        "type": "object",
+        "properties": {
+            "pending_ids": {
+                "type": "array",
+                "items": {"type": "string"},
+                "description": "List of pending record IDs to commit (e.g., ['a1b2c3d4']). From the staged response.",
+            },
+        },
+        "required": ["pending_ids"],
+    },
+}
+
 ALL_TOOL_SCHEMAS = [
     REMEMBER_SCHEMA, RECALL_SCHEMA, SHARED_REMEMBER_SCHEMA, SHARED_RECALL_SCHEMA,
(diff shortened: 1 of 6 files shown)


dplush:
@AxDSan #480 is ready to merge.

Full CI is green across Python 3.10–3.13, plus build, docs, CLA, and CodeRabbit.

It restores the post-merge contracts introduced across #456/#476/#478:
- syncs `mnemosyne_apply_pending` into the Hermes integration tool registry,
- makes embedding retry backoff match the intended bounded exponential + jitter behavior,
- keeps the #477 lifecycle rollback intact while aligning its regression coverage,
- completes the #389 diversity path so reranking uses hydrated `result.content`.

No `uv.lock` changes are included.

dplush:
After #480 lands, I suggest one clean `main` CI verification, then closing #473 and #475 if the shipped changes fully cover their reports.

The next substantive item is #474, which I’d treat as a separate design-first effort: make the FastEmbed/sqlite-vec capability state observable end-to-end, avoid silent FTS-only degradation, and reproduce the embedding-storage/FK path before changing it.

I’d avoid folding that into another quick stabilization PR.

PR #492, opened 2026-07-18
fix: guard working embedding parent race

### LINKED ISSUE #491: Trim-before-embedding can trigger a foreign-key failure
## Summary

`BeamMemory.remember()` commits a new `working_memory` row, runs `_trim_working_memory()`, and only then generates/stores the embedding. If trimming or a concurrent delete removes that parent row before `_store_working_embedding()` inserts into `memory_embeddings`, SQLite raises `IntegrityError: FOREIGN KEY constraint failed`.

The database remains consistent because the child insert fails, but the embedding is lost and the normal trim path emits a warning.

## Reproduction

A deterministic regression can monkeypatch `_trim_working_memory()` to delete the just-created session row before embedding storage, with embeddings enabled. On current `main`, `remember()` logs `embedding storage failed ... FOREIGN KEY constraint failed`.

## Proposed fix

Make the compatibility/fallback insert parent-guarded in the same SQLite statement:

```sql
INSERT OR REPLACE INTO memory_embeddings (memory_id, embedding_json, model)
SELECT ?, ?, ?
WHERE EXISTS (SELECT 1 FROM working_memory WHERE id = ?)
```

Then skip the `vec_working` upsert when the statement inserts no row. This keeps foreign keys enabled, creates no orphan, and treats legitimate parent removal as a clean no-op.

I have a focused regression covering both the missing-parent and present-parent paths, plus the full `remember()` trim-before-embedding ordering.

pvspencer22:
## Summary

- make the fallback `memory_embeddings` insert atomic with its `working_memory` parent-existence check
- if trimming or concurrent deletion removes the parent before that insert executes, skip both the fallback row and `vec_working` upsert
- model the production no-FK `memory_embeddings` schema in regression tests
- preserve normal insert/update behavior when the parent exists

This PR intentionally does not define cleanup semantics for a parent deleted after a successful embedding insert.

Closes #491.

## Verification

```text
pytest -q tests/test_working_embedding_parent_race.py tests/test_beam.py
87 passed, 1 skipped

ruff check tests/test_working_embedding_parent_race.py
All checks passed!

git diff --check
passed
```

Independent read-only review of the final correction: `PASS`.

## AI Assistance

The implementation, regression tests, and initial analysis were prepared with AI agents.

## Summary

- Fixes a race in Mnemosyne’s **working-memory (WM) embedding persistence** path by making `_store_working_embedding` **parent-guarded**: the write now only occurs if the referenced `working_memory.id` still exists, so trimming or concurrent deletion becomes a **clean no-op** instead of a **foreign-key enforcement failure**.
- Prevents **orphan `memory_embeddings`** rows and ensures the WM vector side (`vec_working` / `_wm_vec_upsert`) is **skipped when the parent is gone**, while keeping **SQLite foreign keys enabled** and preserving expected **ON DELETE CASCADE** behavior.
- Preserves Mnemosyne’s **local-first integrity** by containing the correctness fix at the **SQLite write boundary** (no new services, constraints aren’t relaxed, and no cross-layer consistency assumptions are introduced).
- Adds regression coverage for:
  - **missing-parent** insertion (no embedding row written; no FK violations),
  - **present-parent** upsert + **cascade-delete** of embeddings,
  - an end-to-end `remember()` flow where **trim happens before embedding storage** (quiet/no warning; no vector upsert call).
- Updates **changelog** and bumps version to **3.15.0** documenting the **trim-before-embedding foreign-key failure**.

## Architectural impact (core memory architecture)

This change is narrowly scoped to the **WM → embedding** bridge in the tiered memory architecture (WM embeddings backing **working retrieval**), without altering behavior of **episodic/BEAM tiers**, retrieval strategies, consolidation, veracity, or sync.

## Privacy posture / agent integration surfaces

No changes to the privacy posture or agent integration surfaces (**Hermes, MCP, CLI**).

## Long-term maintainability

The fix improves maintainability by:
- hardening an otherwise timing-sensitive DB write into an **invariant-preserving** operation, and
- adding focused tests that lock in expected behavior for the **trim-before-embedding** ordering case and concurrent deletion scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [3.15.0] - 2026-07-18
+
+### Fixed
+
+- **Trim-before-embedding foreign-key failure (#491).** Working-memory embedding storage now atomically checks that its parent row still exists. A row legitimately removed by trimming or concurrent deletion becomes a clean no-op instead of logging a foreign-key failure; foreign-key enforcement remains enabled and no orphan embedding is created.
+
 ## [3.14.0] - 2026-07-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
-- **Trim-before-embedding foreign-key failure (#491).** Working-memory embedding storage now atomically checks that its parent row still exists. A row legitimately removed by trimming or concurrent deletion becomes a clean no-op instead of logging a foreign-key failure; foreign-key enforcement remains enabled and no orphan embedding is created.
+- **Trim-before-embedding race (#491).** Working-memory embedding storage now atomically checks that its parent row still exists. If trimming or concurrent deletion removes the parent before the fallback insert executes, both fallback and `vec_working` writes become a clean no-op instead of logging an embedding-storage failure.
 
 ## [3.14.0] - 2026-07-17
 
--- a/mnemosyne/__init__.py
+++ b/mnemosyne/__init__.py
@@ -10,7 +10,7 @@ Example:
     >>> results = recall("user preferences")
 """
 
-__version__ = "3.14.0"
+__version__ = "3.15.0"
 __author__ = "Abdias J"
 __license__ = "MIT"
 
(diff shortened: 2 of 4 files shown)


dplush:
Thanks for the focused fix. I verified the guarded insert handles the parent-missing-before-insert case.

One schema mismatch remains: production `memory_embeddings` has no FK to `working_memory` (and the migration intentionally removes the legacy FK), while the new test models `ON DELETE CASCADE`. After a successful guarded insert, deleting the working-memory parent still leaves the fallback embedding row, so the changelog claim that no orphan embedding is created is stronger than the production invariant.

Could we either scope the claim and regression to the pre-insert no-op using the production DDL, or define explicit cleanup behavior on parent deletion? The first option seems the smallest fix for this PR.

pvspencer22:
Thanks—agreed. The cascade fixture encoded a stronger invariant than the production schema.

I pushed `63a50c4` with the smaller correction you suggested:

- the fixture now matches the production no-FK `memory_embeddings` DDL;
- the cascade-delete assertion and claim are removed;
- the regression is scoped to the parent already being absent when the guarded insert executes;
- the present-parent insert/update path remains covered; and
- the changelog and PR description now state that narrower guarantee.

No production implementation change was needed. Explicit cleanup after a later parent deletion remains outside this PR. Focused verification is still `87 passed, 1 skipped`; Ruff and diff checks pass.

PR #500, opened 2026-07-21
ci: pin remaining GitHub Action refs

dplush:
## Summary
- pin `actions/cache` in CI to the immutable v4.3.0 commit
- pin both PyPI publish action uses in the release workflow to the immutable v1.14.1 commit

## Verification
- `git diff --check`
- parsed both changed workflows with PyYAML
- verified every workflow `uses:` reference is a full 40-character SHA
- verified the pinned SHAs against the upstream action tags
- independent narrow Claude review: pass-with-followups; tag-to-SHA mapping verified afterwards

Keeps the diff workflow-only; file-permission hardening and `uv.lock` policy remain separate follow-ups.

## Summary

- Pins GitHub Actions to immutable commit SHAs in CI and release workflows.
- No changes to Mnemosyne’s tiered memory, retrieval, consolidation, veracity, sync, benchmarking, or agent integration surfaces (Hermes, MCP, CLI).
- Preserves local-first behavior and privacy posture; no new external runtime dependencies or data flows are introduced.
- Improves long-term maintainability and supply-chain reproducibility by preventing mutable action tags from changing unexpectedly. This is the right call for reliable, auditable automation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Cache embedding model
-        uses: actions/cache@v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ${{ env.MNEMOSYNE_FASTEMBED_CACHE_DIR }}
           key: fastembed-bge-small-en-v1.5-${{ matrix.python-version }}-${{ hashFiles('pyproject.toml') }}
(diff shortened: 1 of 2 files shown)


AxDSan:
Merged. Clean diff, correct SHAs, version comments in place. Exactly the narrow scope we needed.

On the memory errors concern — don't carry that weight. You're not responsible for what other PRs broke. If there's a real pattern in the reports, we'll trace it to the root cause and fix it. That's on me to triage, not on you to worry about.

Thanks for the responsible disclosure on GHSA, the fix verification, and now this. You're the most thorough contributor we've got.

Next up whenever you're ready: file permissions and uv.lock as separate follow-ups.

PR #508, opened 2026-07-22
fix: require instruction keyword boundaries

### LINKED ISSUE #507: MEMORIA instruction extraction inverts "whenever X" into "never X" (plus general fragment-noise findings from a 61-row production audit)
## Summary
The regex-based MEMORIA instruction extractor can store the opposite of what the user said. On our production bank, "Good - whenever needed we can use it. I've disabled it." was stored as the instruction "never needed we can use it" (active=1). Two more rows show the same inversion pattern from "whenever ..." inputs. Because memoria_instructions content is surfaced at recall as "user constraints," an inverted instruction is worse than a missed one.

## Environment
- Version: v0.5.0 tag (mnemosyne-hermes integration versioning; core reports 3.14.0; commit fbe0d13), running as the Hermes Agent memory provider (plugin symlink install), SQLite bank on local disk, Linux/WSL2, Python 3.11.
- Extraction path: the non-LLM regex pass in `BeamMemory` (beam.py, instruction loop around the `memoria_instructions` INSERT, ~L4779-4796 in our checkout): `re.finditer(pat['instruction'], content, re.IGNORECASE)`.

## Reproduction
Store content containing "whenever" followed by more words, e.g.:
- "Good - whenever needed we can use it. I've disabled it."
- "Next useful step, whenever you're ready ..."
Then inspect memoria_instructions. Observed rows (production, verbatim):
- id 59: instruction "never needed we can use it", topic "needed we can use it" (source text said "whenever needed")
- id 38: instruction "never you're ready" (source: "whenever you're ready")
- id 43: instruction "never the agent creates a temporary folder/files for a smoke test and the smoke test completes successfully" (source: "whenever the agent creates ... completes successfully", a positive preference stored as a prohibition)

## Root cause (suspected)
The instruction pattern's "never" alternative matches without a leading word boundary (or the boundary is defeated by how the alternation is assembled after the `IMPVERBS` substitution), so the substring "never" inside "whenever" matches and the captured instruction begins mid-word. A leading `\b` (or `(?<![a-zA-Z])`) on the modal/negation alternatives should fix it. "whenever" may also deserve a spot in `instruction_false_positives`.

## Secondary findings from auditing all 61 rows in one production bank
1. Error-message harvesting: "must be a string" was extracted three times from Python tracebacks in tool output ("int() argument must be a string..."). Suggest excluding content that matches traceback/error shapes, or gating extraction to user-authored turns.
2. Mid-sentence fragments stored as instructions: "never came to be" (from "discussions that never came to be"), "never ends) https://github" (from "(it never ends)"), "must be preserved" without its subject. The captured group starts at the modal, so the constraint's subject is routinely lost, making rows unusable at recall time.
3. Positive observation: the session-scoped recall filter (`active = 1 AND session_id = ?`) plus the dedup deactivation pass meant none of these bad rows were actually reaching prompts in new sessions; the blast radius of the inversion bug is therefore limited in the default setup. Worth stating in docs that memoria_instructions is per-session by design.

## Suggested fixes, in order of value
1. Word-boundary the negation/modal alternatives in `pat['instruction']` ("never" must not match inside "whenever"; check "always" vs "hallways" etc. for the same class of bug).
2. Skip extraction on content that looks like tracebacks/tool errors, or restrict instruction extraction to user-role content.
3. Require the capture to include a subject or start at a sentence boundary, so "must be preserved"-style fragments are dropped instead of stored.
4. Optional: a doctor/diagnose check that flags instructions whose source snippet contains "whenever"/"wherever" against a stored "never"/"ever" instruction, to spot historical inversions in existing banks.

We have deactivated the affected active rows locally; happy to provide the full 61-row dump (redacted) if useful.

Sanjays2402:
## Description

Prevents the `never` instruction keyword from matching inside `whenever`, which inverted positive statements into prohibitions. The focused regression fails before the word-boundary fix and passes after it.

## Related Issue

Closes #507

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [x] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [x] Manual verification (focused regression and standalone `never` instruction)

## Checklist

## Summary

- Fixed BEAM instruction keyword extraction so negation/modals like `never` only match at word boundaries, preventing embedded occurrences (e.g., `whenever`) from being inverted into prohibitions.
- Added regression tests for both the embedded-keyword false positive case and a valid standalone `Never …` instruction.
- Bumped the package version to `3.16.0` and recorded the change in `CHANGELOG.md` (including the issue reference).

## Architectural impact

- **Memory architecture (tiered/BEAM pipeline):** Scope is limited to the **instruction parsing step inside BEAM** (how instruction facts are recognized and stored). It does not change the working/episodic tiering model, nor the structure of the consolidation pipeline.
- **Retrieval strategies & consolidation:** No changes to retrieval ranking logic or consolidation behavior implied by the fix; the impact is upstream at fact-extraction time.
- **Veracity system:** No behavioral changes to veracity scoring/propagation; the fix prevents incorrect “instruction” fact creation in the first place.
- **Sync layer / local-first posture:** No new sync, networking, or persistence flows were introduced by the keyword-boundary change; this is a local parsing/regex correctness fix.
- **Agent integration surfaces (Hermes, MCP, CLI):** Hermes/MCP/CLI continue to rely on the same BEAM fact-extraction path; they benefit from improved correctness when user text contains embedded instruction substrings.

## Privacy & local-first guarantees

- Preserves local-first behavior: the change is purely in-text pattern matching at extraction time, with no added external calls, data export, or altered storage boundaries.

## Maintainability

- The word-boundary requirement plus targeted regression coverage makes this class of parsing bug much less likely to recur, and it explicitly documents the intended keyword-matching contract for future changes.

**This is the right call:** it fixes a narrow extraction boundary issue without expanding scope into retrieval/consolidation/sync, and it strengthens correctness while maintaining local-first design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [3.16.0] - 2026-07-21
+
+### Fixed
+
+- **Instruction extraction inverted “whenever” into “never” (#507).** English instruction keywords now require a word boundary, preventing positive “whenever” statements from being stored as prohibitions.
+
 ## [3.15.0] - 2026-07-20
 
 ### Fixed
--- a/mnemosyne/__init__.py
+++ b/mnemosyne/__init__.py
@@ -10,7 +10,7 @@ Example:
     >>> results = recall("user preferences")
 """
 
-__version__ = "3.15.0"
+__version__ = "3.16.0"
 __author__ = "Abdias J"
 __license__ = "MIT"
 
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -4505,7 +4505,7 @@ class BeamMemory:
             'sequence': r'((?:first|second|third|fourth|fifth|finally|next|then|after that)[^.,;!?\n]{15,120})',
             'instruction_false_positives': ['i think you should leave', 'should behave', 'their work style'],
             'instruction_imperative': 'always|never|remember|use|keep|avoid|ensure|check|verify|run|test|build|deploy|push|pull|merge|commit|close|open|update|install|configure|set|enable|disable|add|remove|create|delete|start|stop|restart|reload|reset|try|implement|write|read|switch|move|copy|rename|send|reply|respond',
-            'instruction': r'(?:always|never|must|must not|should(?: not)?(?=\s+(?:you|we|i|one)\s+(?:IMPVERBS))|need(?:s)? to(?: not)?|required to|prefer(?: not)? to|want to(?: avoid| ensure| use| keep))\s+([^.,;!?\n]{10,200})',
+            'instruction': r'\b(?:always|never|must|must not|should(?: not)?(?=\s+(?:you|we|i|one)\s+(?:IMPVERBS))|need(?:s)? to(?: not)?|required to|prefer(?: not)? to|want to(?: avoid| ensure| use| keep))\s+([^.,;!?\n]{10,200})',
             'preference': r'(?:'
                 # First person (original) and second person
                 r'(?:I|You|you|YOU)(?: |\')?(?:like|love|prefer|hate|dislike|enjoy|use|stick with|switched to|moved to|changed to|want|need|tend to|usually|would rather|don\'t like|don\'t want|not a fan of|am okay with|am comfortable with|am used to|am happy with|am tired of|am sick of|prefer not to|try to avoid|find it easier to|find it better to|find it useful to)'
--- a/tests/test_instruction_word_boundary.py
+++ b/tests/test_instruction_word_boundary.py
@@ -1,0 +1,29 @@
+from mnemosyne.core.beam import BeamMemory
+
+
+def test_whenever_is_not_extracted_as_never(tmp_path):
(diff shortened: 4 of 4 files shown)


AxDSan:
@Sanjays2402 Clean fix. The word-boundary regex prevents "whenever" from being parsed as "never", and the test covers both the false positive and the true positive case. One blocker: the CLA check needs your signature. Sign at https://cla-assistant.io/mnemosyne-oss/mnemosyne?pullRequest=508 (30 seconds, GitHub login). Once signed, merging.

AxDSan:
The fix is correct and CI is green. Only thing left is the CLA, which still shows as not signed for your account. Sign it at the link in the `license/cla` check and I'll get this merged. Thanks for the fix.

AxDSan:
Superseded, closing.

The word-boundary fix landed in `334ca47` covering all five locale patterns (en/de/ru/it/es). Your diagnosis and fix approach are credited by name in the CHANGELOG entry for #507. This branch is also CONFLICTING now that the change is on main.

Thanks for the original catch. The `\b` anchor was exactly right, it just grew to cover the other four locales on the way in.

Related follow-ups if you want them, both confirmed still live on main: #559 (the negation patterns have the same missing anchor, and the `en` one pairs it with `\s*` so it can match zero-width) and #560 (the `ru` instruction pattern uses `\\s+` inside a raw string, so Russian instruction extraction never fires at all).

Issue #482, opened 2026-07-16
[BUG] 50 of 106 config.yaml keys are silently ignored — core modules read env vars into module-level constants at import


Pawls:
## Summary

Roughly half the `config.yaml` keys have no effect on behaviour. `mnemosyne config set <key>` writes them, `mnemosyne config get <key>` reads them back, and `MnemosyneConfig` resolves them correctly — but the code that actually *uses* them reads `os.environ` into a **module-level constant at import time** and never consults the config.

The failure is silent: no error, no warning, and `config get` cheerfully reports the value you set.

## Version

- `mnemosyne-memory` 3.12.2
- `mnemosyne-hermes` 0.4.0

## Minimal reproduction

`config.yaml` contains `cross_session: 'true'`; `MNEMOSYNE_CROSS_SESSION` is **not** set in the environment.

```python
>>> from mnemosyne.core.config import get_config
>>> from mnemosyne.core.beam import _cross_session_enabled
>>> get_config().get_bool('cross_session')
True                      # the config system reads it
>>> _cross_session_enabled()
False                     # the gate that actually controls recall ignores it
```

Same key, two answers, in one process.

## Root cause

`mnemosyne/core/beam.py:406`:

```python
_CROSS_SESSION = os.environ.get("MNEMOSYNE_CROSS_SESSION", "0") == "1"

def _cross_session_enabled() -> bool:
    return _CROSS_SESSION or os.environ.get("MNEMOSYNE_CROSS_SESSION", "0") == "1"
```

Nothing bridges config → env: the Hermes provider never assigns `os.environ[...]`, and `core/config.py` only *reads* env (line 380) while seeding.

This contradicts the stated design. `core/config.py`, immediately below `REQUIRES_RESTART`, declares:

> Mapping from config key (snake_case, no MNEMOSYNE_ prefix) to env var name.
> **This is the canonical bridge between config.yaml keys and env vars.**

That bridge only runs one way (env → seeded `config.yaml`, via `config migrate`). The reverse — `config.yaml` → the constants that gate behaviour — doesn't exist.

## Scope: 50 of 106 mapped keys

Cross-referencing the config key→env map in `core/config.py` against module-scope `os.environ.get("MNEMOSYNE_*")` assignments in `mnemosyne/core/*.py`:

| module | affected keys |
|---|---|
| `beam.py` | 17 |
| `local_llm.py` | 15 |
| `shmr.py` | 7 |
| `llm_conflict_detector.py` | 4 |
[...]

Issue #446, opened 2026-07-11
Proposal: versioned, provenance-aware migrations for auto-seeded config.yaml


dplush:
## Motivation

#445 fixes the immediate Hermes provider-default drift conservatively: new `config.yaml` files receive safe defaults, while existing legacy files are detected and users receive explicit remediation guidance.

That is the right patch-level behaviour. More generally, once an auto-seeded config has been materialized, changing `DEFAULTS` alone cannot safely evolve it. The same values may originate from a prior seed, environment-derived values, a profile, or an intentional manual override.

## Goal

Define a small, versioned and provenance-aware migration path for **future** auto-seeded `config.yaml` changes.

## Non-goals

- Do not block or broaden #445.
- Do not infer provenance from a matching value pair.
- Do not silently rewrite manual, ambiguous, environment-derived, profile-specific, or kwargs-driven configuration.
- Do not turn every default change into an automatic migration.

## Proposed contract

- New auto-seeded configurations record a durable config format/version and reliable seed provenance.
- Migrations are explicit, narrowly preconditioned, and registered per version transition.
- A CLI can report applicability before mutation, e.g. `mnemosyne config migrate --dry-run`, and requires explicit apply for changes.
- Applicable migrations are idempotent and atomic; failures preserve the original file and permissions.
- Manual or ambiguous files remain unchanged and receive actionable guidance instead.
- Profile-aware behavior is defined rather than assumed.

## Acceptance criteria

- [ ] Version/provenance format is documented and backward-compatible.
- [ ] Migration discovery has a read-only preview with clear reasons for skip/apply.
- [ ] Apply path is explicit, atomic, idempotent, and has a rollback/backup story.
- [ ] Tests cover fresh seed, known legacy seed, manually edited seed, explicit environment values, profiles, malformed YAML, and repeated invocation.
- [ ] The provider-facing compatibility matrix covers `kwargs > config.yaml > env > runtime default` across both Hermes provider surfaces.

## Context

Follow-up to #445. This is a design/maintenance issue, not a request to retrofit broad automatic migration into that PR.

AxDSan:
@dplush agreed on the direction. Versioned, provenance-aware migrations for auto-seeded config.yaml is the right call.

The current auto-seed path (`config.py:_seed()`) has no migration logic. It either creates a fresh config or leaves existing files alone. There's no way to safely add new defaults without breaking existing installs.

The implementation would:
1. Add a `config_version` field to the auto-seeded YAML
[...]

Issue #421, opened 2026-07-07
[BUG] cross-session/scope memory visibility — MNEMOSYNE_CROSS_SESSION crashes recall + CLI store ignores MNEMOSYNE_DEFAULT_SCOPE


jluzny:
Two related bugs make cross-session memory visibility unusable. Both concern
   how memories are scoped and recalled across sessions. Confirmed present in
   `main` tip `d7387b4` (v3.11.1).

   ---

   ## Bug 1 — `MNEMOSYNE_CROSS_SESSION=1` crashes recall (`sqlite3.ProgrammingError`)

   Setting `MNEMOSYNE_CROSS_SESSION=1` (the intended way to make memories visible
   across sessions) crashes every recall path with:

 ```

 sqlite3.ProgrammingError: Incorrect number of bindings supplied.
 The current statement uses 0, and there are 1 supplied.

 ```

   ### Root cause

   `_session_scope_filter()` in `mnemosyne/core/beam.py` (line 409, added in #371)
   returns `"(1=1)"` — **zero placeholders** — when cross-session is on:

   ```python
   def _session_scope_filter(extra_col: str = "") -> str:
       if _CROSS_SESSION:
           return "(1=1)"                                            # 0 '?'
       if extra_col:
           return f"(session_id = ? OR scope = 'global' OR {extra_col} = ?)"  # 2 '?'
       return "(session_id = ? OR scope = 'global')"                # 1 '?'
 ```

 But every caller still appends params unconditionally, e.g. beam.py:5404:

 ```python
   wm_where_clauses.append(_session_scope_filter())
   wm_params.append(self.session_id)   # appended even when (1=1) has 0 placeholders
 ```

 Affected call sites (all desync): working memory (5399/5404), episodic
 (5657/5663, 5902/5907), fact+episodic (5779/5785), enhanced recall
 (6345/6349, 6949/6953).

 ### Reproduce

 ```python
   import os, tempfile
   os.environ["MNEMOSYNE_CROSS_SESSION"] = "1"
   from mnemosyne.core.beam import BeamMemory
   db = tempfile.mktemp(suffix=".db")
   a = BeamMemory(db_path=db, session_id="A"); a.remember("test", scope="global")
   b = BeamMemory(db_path=db, session_id="B"); b.recall("test")  # → ProgrammingError
 ```

 ### Fix (minimal, no caller changes)

 Keep the placeholder count identical to the non-cross-session branch but make
[...]

PR #510, opened 2026-07-22
fix(config): honor config for cross-session recall

dplush:
## Summary

Part 1 of #482: migrate the real Beam `cross_session` consumer from import-time environment reads to the existing typed config resolver.

- preserves the documented `config.yaml > env > default` contract
- removes the stale import-time `_CROSS_SESSION` cache; the existing config reader observes reloads
- keeps existing env-only behavior when `config.yaml` does not set the key

## Coverage

- fresh-process direct-core matrix with conflicting YAML/env values
- fresh-process Hermes-provider matrix with the same conflicts
- isolates existing #421 env-fallback tests from an ambient developer config

## Non-goals

- no YAML-to-`os.environ` bridge or process-global environment mutation
- no config migration/provenance work (#446)
- no conversion of the remaining config consumers; those follow after this pattern lands

## Tests

```text
PYTHONPATH=.:integrations/hermes/src uv run --no-sync --with pytest --with pyyaml pytest tests/test_config_cross_session_runtime.py tests/test_issue_421_scope.py tests/test_config_profiles.py integrations/hermes/tests/test_recall_diagnostics_tool.py -q
# 87 passed
```

## Summary

- **Core architecture:** Cross-session recall is now driven by typed, hot-reloadable runtime configuration (`resolve_beam_runtime().cross_session`) instead of an import-time environment cache. `BeamMemory.recall()` resolves `cross_session` once per call and threads it through all session-scoped SQL construction (working memory and episodic memory, including channel vs. non-channel branching and the author-only isolation path). The polyphonic recall/update path propagates the same value end-to-end (`_recall_polyphonic(..., cross_session=...)` → `_polyphonic_row_passes_filters(...)`) so both selection guards and the `recall_count/last_recalled` update scoping are consistent.
- **Config precedence & local-first determinism:** The existing precedence model is preserved (**`config.yaml` > env > default**), including explicit YAML-versus-env conflicts. By removing module-level caching, BEAM behavior tracks the existing config reader/reloadable configuration during runtime, avoiding stale ambient settings.
- **Privacy/local-first posture:** No new sync, external data flows, provenance/migration, or network-facing behavior is introduced. The change is strictly a local storage visibility boundary—what rows are eligible for recall and which rows get recall bookkeeping—based on the resolved `cross_session` flag.
- **Agent integration surfaces:** Hermes-provider coverage is added to ensure the same YAML-versus-env override semantics apply when Hermes writes/loads `mnemosyne/config.yaml`, matching direct-core recall behavior.
- **Maintainability:** Centralizing runtime resolution makes precedence and reload behavior explicit, reduces duplicated implicit env handling, and removes drift between import-time state and runtime configuration. `recall_enhanced()`’s query cache key is updated to include the resolved `cross_session` value to prevent cross-session cache collisions. Subprocess-based regression tests also isolate environment-fallback behavior from ambient developer configuration.
- **Validation:** 87 tests pass; benchmark methodology is unchanged.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -23,6 +23,8 @@ import hashlib
 import threading
 import math
 
+from mnemosyne.core.config import resolve_beam_runtime
+
 logger = logging.getLogger(__name__)
 from datetime import datetime, timedelta, timezone
 from typing import List, Dict, Optional, Any, Set, Union, Tuple, Callable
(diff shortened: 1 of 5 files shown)


AxDSan (approved):
This is the #482 design direction implemented correctly. resolve_beam_runtime() replaces the module-level _CROSS_SESSION env read, config.yaml > env > default precedence is preserved, and the cross_session snapshot is threaded consistently through all recall paths. Query cache scoping by runtime snapshot prevents stale hits after hot reload. Tests cover hot reload and scope isolation. Merging.

PR #512, opened 2026-07-22
fix: default enhanced recall weights when omitted

### LINKED ISSUE #511: [BUG] Enhanced Recall crashes when public recall forwards omitted hybrid weights as None
## Summary

`MNEMOSYNE_ENHANCED_RECALL=1` can crash the public `Mnemosyne.recall()` API whenever query-intent classification is non-general and callers omit the optional hybrid weights.

This appears to be a follow-up regression after #436 wired the public API to `BeamMemory.recall_enhanced()`.

## Reproduction

Tested with `mnemosyne-memory 3.14.0`, Python 3.13.5:

```python
import os
from pathlib import Path
from mnemosyne.core.memory import Mnemosyne

m = Mnemosyne(session_id="repro", db_path=Path("/tmp/repro.db"))
m.remember("The sprint ends on March 29.", scope="global")
os.environ["MNEMOSYNE_ENHANCED_RECALL"] = "1"

m.recall("When does the sprint end?", top_k=3)
```

Observed:

```text
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
...
mnemosyne/core/query_intent.py", line 158, in adjust_weights
    vw = base_vec * intent.vec_bias
```

## Root cause

`Mnemosyne.recall()` has optional defaults of `None` for `vec_weight`, `fts_weight`, and `importance_weight`, and forwards those explicit `None` values to `BeamMemory.recall_enhanced()` when the feature flag is enabled.

The enhanced intent branch then does:

```python
vw, fw, iw = adjust_weights(
    base_vec=kwargs.pop("vec_weight", 0.5),
    base_fts=kwargs.pop("fts_weight", 0.3),
    base_importance=kwargs.pop("importance_weight", 0.2),
    intent=intent,
)
```

`dict.pop(..., default)` only applies when the key is absent; here it is present with value `None`. `adjust_weights()` multiplies that `None` by an intent bias.

Current `main` has the same shape in `mnemosyne/core/memory.py` (public routing) and `mnemosyne/core/beam.py` (intent adjustment), so this is not only a packaged-3.14.0 concern.

## Expected behavior

Enabling Enhanced Recall without explicitly supplying weights should use the documented defaults (`0.5`, `0.3`, `0.2`) and never crash.

## Minimal fix direction

Resolve each optional value before calling `adjust_weights()`, preserving explicit zero values:

```python
vec_weight = kwargs.pop("vec_weight", None)
fts_weight = kwargs.pop("fts_weight", None)
importance_weight = kwargs.pop("importance_weight", None)

vw, fw, iw = adjust_weights(
    base_vec=0.5 if vec_weight is None else vec_weight,
    base_fts=0.3 if fts_weight is None else fts_weight,
    base_importance=0.2 if importance_weight is None else importance_weight,
    intent=intent,
)
```

A focused regression should exercise the public `Mnemosyne.recall()` path with `MNEMOSYNE_ENHANCED_RECALL=1`, a query that is classified as non-general, and no explicit weights.

## Scope

No production data or secrets are involved. This is a deterministic local reproduction using an isolated SQLite database.

dplush:
## Summary

Fixes #511.

When `MNEMOSYNE_ENHANCED_RECALL=1`, the public `Mnemosyne.recall()` path forwards optional hybrid weights as explicit `None` values. The enhanced intent path treated those as numeric values and crashed during weight adjustment.

This normalizes only `None` to the documented defaults (`0.5`, `0.3`, `0.2`) before calling `adjust_weights()`. Explicit zero values remain valid overrides.

## Tests

```text
uv run --no-sync --with pytest --with pyyaml pytest \
  tests/test_query_intent_recall.py tests/test_weibull_mmr_intent.py -q
# 26 passed
```

The regression covers the public `Mnemosyne.recall()` API with Enhanced Recall enabled, a non-general query, and omitted weights; it also exercises explicit zero-valued weights.

## Summary

- Fixes Enhanced Recall crashes when hybrid weights are omitted through `Mnemosyne.recall()`.
- Normalizes omitted (`None`) hybrid weights to the documented defaults—vector `0.5`, FTS `0.3`, importance `0.2`—before intent-based adjustment, while preserving explicit zero values as valid overrides.
- Adds regression coverage for omitted-weight behavior, explicit zero overrides, and environment-variable configured weights via the public recall path.

## Architectural impact

- BEAM retrieval reliability: makes the Enhanced Recall weight-adjustment boundary more robust, preventing `None` from propagating into weight adjustment logic.
- No tiered-memory changes: working/episodic/BEAM structure, consolidation, and veracity systems are unaffected.
- No retrieval-strategy redesign: only the parameter normalization step for Enhanced Recall is modified; the underlying recall flow remains the same.
- Local-first & privacy posture preserved: no new services, network calls, storage, or external dependencies are introduced.
- Agent integration surfaces unchanged: Hermes, MCP, and CLI integration points are not modified; they benefit indirectly by using the safer public recall path.
- Maintainability win: aligns runtime behavior with documented defaults at the API boundary, reducing edge-case complexity and preventing regressions around optional arguments.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -6697,10 +6697,15 @@ class BeamMemory:
         if use_intent and classify_intent is not None:
             intent = classify_intent(query)
             if intent.category != "general" and adjust_weights is not None:
+                vec_weight, fts_weight, importance_weight = _normalize_weights(
+                    kwargs.pop("vec_weight", None),
+                    kwargs.pop("fts_weight", None),
+                    kwargs.pop("importance_weight", None),
+                )
                 vw, fw, iw = adjust_weights(
-                    base_vec=kwargs.pop("vec_weight", 0.5),
-                    base_fts=kwargs.pop("fts_weight", 0.3),
-                    base_importance=kwargs.pop("importance_weight", 0.2),
+                    base_vec=vec_weight,
+                    base_fts=fts_weight,
+                    base_importance=importance_weight,
                     intent=intent,
                 )
                 kwargs["vec_weight"] = vw
--- a/tests/test_query_intent_recall.py
+++ b/tests/test_query_intent_recall.py
@@ -64,3 +64,56 @@ def test_explicit_recall_weights_override_query_intent(monkeypatch):
             importance_weight=0.1,
         )
         assert results
+
+
+def test_public_enhanced_recall_resolves_weight_defaults_and_overrides(monkeypatch):
+    from mnemosyne.core import beam as beam_module
+    from mnemosyne.core.memory import Mnemosyne
+
+    observed_base_weights = []
+    original_adjust_weights = beam_module.adjust_weights
+
+    def capture_adjust_weights(*args, **kwargs):
+        observed_base_weights.append(
+            (kwargs["base_vec"], kwargs["base_fts"], kwargs["base_importance"])
+        )
+        return original_adjust_weights(*args, **kwargs)
+
+    monkeypatch.setattr(beam_module, "adjust_weights", capture_adjust_weights)
+    monkeypatch.setenv("MNEMOSYNE_ENHANCED_RECALL", "1")
+    monkeypatch.delenv("MNEMOSYNE_VEC_WEIGHT", raising=False)
+    monkeypatch.delenv("MNEMOSYNE_FTS_WEIGHT", raising=False)
+    monkeypatch.delenv("MNEMOSYNE_IMPORTANCE_WEIGHT", raising=False)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        db_path = Path(tmpdir) / "mnemosyne.db"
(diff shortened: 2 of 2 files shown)


dplush:
Addressed the two #511-scoped findings in dc0f9d4:

- Enhanced recall now resolves omitted weights through the existing environment-aware `_normalize_weights()` path, preserving explicit zero values.
- The public-path regression now uses distinct queries, asserts returned content, and captures the inputs passed to `adjust_weights()` for default, explicit-zero, and environment-configured weights.

The cache-key finding is valid but broader than #511: the cache currently omits more than weights (filters and other ranking options). I filed #513 to define and fix that cache contract separately, keeping this crash fix atomic.

AxDSan (approved):
Correct fix. Routing through _normalize_weights ensures config.yaml > env > default precedence is respected for recall weights, consistent with the #482 design direction. Tests verify all three paths (default, explicit, configured). Merging.

PR #515, opened 2026-07-23
fix(embeddings): sync config.yaml and env vars in embeddings module

HaoNgo232:
## Description
Fixes an issue where `mnemosyne/core/embeddings.py` was reading direct `os.environ` variables at import time instead of dynamically checking `get_config()` / `config.yaml`.

## Changes
- Updated `_get_api_key()`, `_get_base_url()`, `_get_default_model()`, `_is_disabled()`, `_is_api_model()`, and `_get_embedding_dim()` in `embeddings.py` to resolve configuration values dynamically via `get_config()`.
- Added unit test suite `tests/test_embedding_config_sync.py` to verify configuration resolution.

## Testing
- Ran all embedding tests with 100% pass rate (`23 passed in 3.10s`).

## Summary

- Makes dense-embedding configuration in `mnemosyne/core/embeddings.py` resolve dynamically at runtime: API key, base URL, default model, disabled state, API-vs-local routing, and embedding dimension are derived from `get_config()` / `config.yaml` with env-var overrides, eliminating import-time “stale” configuration.
- Preserves local-first behavior by gating availability and embedding generation on the resolved decision (local fastembed vs API) and explicit disable flags; embed calls now early-return `None` when embeddings are disabled.
- Strengthens robustness of vector dimensionality: `_get_embedding_dim()` supports an `MNEMOSYNE_EMBEDDING_DIM` override (safe int parsing), otherwise uses a known-model dimension map, then falls back to positive `embedding_dim` from config; BEAM’s sqlite-vec dimension-mismatch guard continues to prevent opening incompatible vector DBs and reindex rebuilds with the resolved dimension.
- Adds `tests/test_embedding_config_sync.py` to verify that embedding resolution helpers stay synchronized with the pinned active configuration source (`config.yaml` under `HERMES_HOME` plus env overrides), and extends coverage around embedding prefix behavior in `tests/test_embedding_prefixes.py`.

## Architectural impact

- **Local-first / privacy posture:** Dense retrieval/provider choice is now synchronized with the active configuration rather than what happened to be loaded at import time. Privacy risk remains provider-dependent: for OpenRouter-style endpoints the code requires an API key, while for custom endpoints it may call without a key and only attaches `Authorization` when a key is present. Misconfiguring `embedding_api_url` (or enabling API routing) can therefore still cause external data egress even in setups intended to be local-first—now driven by config correctness at runtime.
- **Memory architecture (tiers, retrieval, consolidation, veracity, sync):** No direct changes to working/episodic/BEAM tier logic, retrieval strategy implementations, consolidation pipeline, veracity, or the sync layer. Indirectly, any recall/embedding-dependent pathways that rely on `embeddings.available()` and `embed_query()`/`embed()` now consistently inherit the synchronized “API vs local” decision and the resolved embedding dimension.
- **Agent integration surfaces (Hermes, MCP, CLI):** Any integration that indirectly imports/uses embeddings benefits from reduced drift between “startup config” and “current config,” since the embedding module resolves credentials/routing/dimensions via the runtime configuration helpers.
- **Maintainability / drift prevention:** The helper-based resolution plus the new synchronization tests are the “right call” for long-term maintainability: they prevent configuration drift across `config.yaml`, env vars, and module import timing, and they reduce dimension mismatch surprises by ensuring dimension selection follows the same resolved model/config source.

## Validation

- New pytest coverage asserts synchronized helper behavior across config/env precedence for default model, embedding dimension, API base URL, API key, API-vs-local routing (`_is_api_model()`), and disabled-state enforcement (`_is_disabled()` / `available_api()`), with additional prefix-path coverage to ensure correct document/query prefix application on both local and API paths.

### DIFF
--- a/mnemosyne/core/embeddings.py
+++ b/mnemosyne/core/embeddings.py
@@ -6,10 +6,10 @@ Falls back to keyword-only if neither is available.
 from __future__ import annotations
 
 import json
+import logging
 import os
 import random
 import ssl
-import sys
 import time
 import urllib.error
 import urllib.request
@@ -72,8 +72,10 @@ def _get_config_safe():
     try:
         if _default_config_path().exists():
             return get_config()
-    except Exception:
-        pass
+    except Exception as exc:
+        logging.getLogger(__name__).warning(
+            "_get_config_safe: failed to load config: %s", exc
+        )
     return None
 
 def _get_api_key() -> str:
@@ -92,12 +94,6 @@ def _get_base_url() -> str:
 def _get_default_model() -> str:
     if "MNEMOSYNE_EMBEDDING_MODEL" in os.environ:
         return os.environ["MNEMOSYNE_EMBEDDING_MODEL"]
-    # In test context (pytest running), do NOT read from config.yaml:
-    # the dev machine's configured model would break tests that hardcode
-    # "BAAI/bge-small-en-v1.5" as the expected default. Tests that want
-    # a specific model must set MNEMOSYNE_EMBEDDING_MODEL via monkeypatch.
-    if "pytest" in sys.modules:
-        return "BAAI/bge-small-en-v1.5"
     cfg = _get_config_safe()
     cfg_model = cfg.get_str("embedding_model") if cfg else ""
     return cfg_model or "BAAI/bge-small-en-v1.5"
@@ -124,7 +120,6 @@ def _get_prefix(kind: str) -> str:
     prefix = os.environ.get(var, "")
     global _PREFIXES_LOGGED
     if prefix and not _PREFIXES_LOGGED:
-        import logging
         logging.getLogger(__name__).info(
             "embedding prefixes active: query=%r doc=%r",
             os.environ.get("MNEMOSYNE_EMBEDDING_QUERY_PREFIX", ""),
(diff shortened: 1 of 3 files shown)


HaoNgo232:
@AxDSan PR has been updated with fixes for all previously reported issues:

- Removed `"pytest" in sys.modules` guards from production code (3 locations)
- Restored `_is_disabled()` to match original env-var-presence semantics
- Removed redundant `"text-embedding" in model_name` check in `_is_api_model()`
- Added config.yaml reading for `embedding_api_url` in `_is_api_model()`
- Added `_is_disabled()` check to `available_api()`
- Added warning logging to `_get_config_safe()` instead of silent `except: pass`
- Rewrote tests to use real config.yaml files and cover config-only code paths
- Fixed `test_embedding_prefixes.py` to be environment-independent

All embedding tests passing: 6/6 ✅. Ready for review. CC @coderabbitai

zodiacg:
I hit the exact import-order failure that this PR fixes in a live Hermes deployment, and I think this PR may have been closed accidentally before it received maintainer review.

**Observed on current releases**

- `mnemosyne-memory==3.15.1`, `mnemosyne-hermes==0.5.0`
- current Hermes Agent `main` at `892ec7af57`
- Mnemosyne's own persisted config explicitly contains:
  ```yaml
  embedding_model: BAAI/bge-small-zh-v1.5
  embedding_dim: 512
  ```
- Hermes `.env` also explicitly contains the same `MNEMOSYNE_EMBEDDING_MODEL` and `MNEMOSYNE_EMBEDDING_DIM=512` values.
- Existing sqlite-vec tables contain 512-dimensional vectors.

Current Hermes imports `hermes_cli.config` before it calls `load_hermes_dotenv()`. `config` eagerly discovers pip entry-point providers; that imports `mnemosyne_hermes`, then `mnemosyne.core.embeddings`. At that point the Hermes dotenv has not been applied, so current `embeddings.py` freezes:

```python
_DEFAULT_MODEL = ... or "BAAI/bge-small-en-v1.5"
EMBEDDING_DIM = _get_embedding_dim(_DEFAULT_MODEL)  # 384
```

The gateway then logs the expected failure on every startup:

```
database stores 512-dim vectors but this process is configured for 384-dim
```

This is not solved by setting either config file correctly: both are already correct. It is precisely the stale import-time configuration problem that this PR's runtime `get_config()` resolution addresses.

GitHub's timeline shows this PR was closed by its author, not merged or closed by a maintainer; 37 seconds after the final close the author commented that the review fixes were complete and it was ready for review. None of the PR's final commits are in current `main`.

Could a maintainer please consider reopening/rebasing this PR, or confirm the intended follow-up? The later #521 fail-loud change is valuable but deliberately does not make embedding model/dimension resolve through `MnemosyneConfig`, so this Hermes integration failure remains reproducible.

HaoNgo232:
Thanks @zodiacg for the detailed breakdown and for bringing this back up!

I've reopened the PR. To provide some background: I originally closed it because the changes grew a bit too broad, touched multiple core parts, and ran into CI failures that I didn't have the time to resolve back then.

The core intent remains fixing the stale import-time config via runtime `get_config()` resolution.

Maintainers and contributors are more than welcome to push fixes/rebase directly to this branch, or help streamline/simplify the changes so we can get CI green and get this merged!

dplush:
Thanks for reopening this. The reported import-order failure is real: model and vector dimension can be frozen before the Hermes dotenv/config bootstrap, while BEAM later uses that frozen dimension for the sqlite-vec mismatch boundary.

I do not think this branch is safe to revive by rebasing. It is currently conflicting with main, its last test matrix is red, and the patch mixes a restart-scoped embedding identity contract with partial runtime re-resolution. In particular, model, dimension, route, and API URL can diverge from cached local-model/query-cache state, while the vector initialization boundary still retains import-time globals. The current helpers also make environment values win where the documented configuration contract is YAML > environment > default.

Recommended disposition: supersede this branch with a new, narrow, contract-first PR rather than continuing to patch it.

The successor should resolve one typed embedding startup snapshot after Hermes dotenv/config bootstrap and before any embeddings or BEAM/vector initialization. That same snapshot should drive model, dimension, route, API URL, and the vec mismatch guard. These settings should remain restart-scoped, not promise arbitrary hot reload. Preserve YAML > environment > default, avoid retaining API keys in compatibility globals, and fail before vec writes on model/dimension incoherence.

The decisive regression is a fresh-process Hermes-order test: import/discover the provider before dotenv, then load a configured 512-dimension model and initialize Beam against an existing 512-vector database successfully. Add the matching negative mismatch case and ensure query-cache identity cannot cross model/route boundaries.

Once such a successor exists, closing #515 as superseded would keep the report and author credit while avoiding a risky conflict-resolution loop on this historical branch.

dplush:
Tracked by #818. This branch should be closed as superseded once the replacement PR is open.

PR #517, opened 2026-07-23
fix(doctor): load sqlite-vec in read-only reports

dplush:
## Summary
- load the installed optional `sqlite-vec` extension only inside Doctor’s already `mode=ro` / `query_only` connection
- retain `present_but_unloadable` when the optional load fails
- fail closed if extension loading cannot be disabled again

## Why this path
The Doctor already detects actual vec0-table capability, but it previously did so in a connection where the installed extension was never loaded. This produced a false `present_but_unloadable` warning for healthy sqlite-vec deployments.

## Non-goals / rollback
No provider, repair, schema, or data changes. If sqlite-vec is absent or fails to load, behavior remains the existing conservative warning path.

## Why not
Do not infer capability from package import/runtime diagnostics alone: the Doctor continues to query the live vec0 tables after loading.

## Verification
- `PYTHONDONTWRITEBYTECODE=1 .venv/bin/python -m pytest -p no:cacheprovider tests/test_doctor.py tests/test_doctor_cli.py -q` → 66 passed
- product-DB read-only smoke: vec0 `available`, no `sqlite.vec0_capability` warning
- regression coverage for extension disablement, load failure, real vec0 availability, and fail-closed disable failure.

## Summary

- Updates Doctor’s read-only reporting to load the installed optional `sqlite-vec` extension within its existing `mode=ro` / `query_only` connection.
- Prevents false diagnostics by keeping the existing `present_but_unloadable` fallback when `sqlite-vec` can’t be imported/loaded or when `vec0` is present but not usable in the current runtime.
- Adds “fail closed” safety: if extension loading cannot be disabled after enabling it, Doctor marks multiple report sections (`sqlite_health`, `reference_contracts`, `vector_coverage`, `hygiene_summary`) as unavailable and exits early, preserving the read-only contract.
- Adds targeted regression tests covering optional load behavior, fallback correctness, and failure-closed handling when extension enable/disable operations error.

## Architectural impact

- No changes to Mnemosyne’s core memory architecture (working/episodic/BEAM tiers), retrieval strategies, consolidation pipeline, veracity system, sync layer, schema/data model, or benchmark methodology.
- No changes to agent integration surfaces (Hermes, MCP, CLI command behavior beyond using the updated Doctor report builder).
- The right call for local-first/privacy: vector capability verification is done entirely against the local product database via Doctor’s bounded, content-safe read-only inspection (no repairs, no migrations, no data writes), and the connection is kept constrained (`mode=ro`, `query_only`) even when optional extension loading is attempted.
- Maintains long-term safety/maintainability by making extension lifecycle handling explicit and test-covered, specifically avoiding ambiguous “healthy runtime” vs “false warning” states for `vec0`.

### DIFF
--- a/mnemosyne/doctor.py
+++ b/mnemosyne/doctor.py
@@ -66,6 +66,11 @@ _RUNTIME_ABSOLUTE_PATH = re.compile(
     r"(?<![A-Za-z0-9_.-])(?:~[\\/]|(?:[A-Za-z]:)?[\\/])[^\s`<>\"']+"
 )
 
+
+class _SQLiteVecExtensionDisableError(RuntimeError):
+    """The connection cannot safely continue after extension loading was enabled."""
+
+
 # Adapter SQL is deliberately limited to these known contracts. Catalog
 # metadata only proves that one of these names exists; it never turns an
 # arbitrary application table into a doctor query target.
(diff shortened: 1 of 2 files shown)


AxDSan (approved):
Good catch on the read-only doctor path. sqlite-vec should be loadable for diagnostics even when the main DB is in a degraded state. Tests look thorough. Merging.

PR #519, opened 2026-07-23
fix(beam): make embedding dim-mismatch message explicitly self-healing

### LINKED ISSUE #518: Embedding 'dimension mismatch' is misread as 'database corrupt' — make the guard message self-healing
## Problem

When a mnemosyne database stores `vec0` vectors at one dimension but the process is configured for another — most commonly because an unlisted embedding model silently resolved to the 384-dim fallback in `_get_embedding_dim`, then a later process is configured for the real dimension — the `init_beam` dimension-consistency guard logs:

> Embedding dimension mismatch: this database stores 384-dimensional vectors, but the process is configured for 1024 ... Set MNEMOSYNE_EMBEDDING_DIM / MNEMOSYNE_EMBEDDING_MODEL to match the stored data, or run `mnemosyne reindex` to rebuild all vectors at the configured dimension.

In practice this is **routinely misread as "the database is corrupt."** I observed an agent framework (and users) interpret "dimension mismatch" as corruption, fall back to legacy storage, and abandon the store — never running the `reindex` that fixes it in seconds. The guard is correct (it leaves all memories and existing `vec0` tables untouched, so this is **never** corruption — only the vector index is affected), but the message doesn't say so, and it conflates two recovery options without giving the exact command.

(The recurring per-model additions to the dim table — e.g. the Jina v2 fix in v3.15.0, bge-m3 — are symptoms of the same root cause: an unlisted model silently assuming 384. That deeper fix is separate.)

## Proposal

Make the guard message **explicit and self-healing** so it can't be misread as corruption:
- state plainly it is **NOT database corruption** and memories are intact (recall falls back to keyword search meanwhile);
- give **both** recovery paths with the **exact** commands:
  - keep the existing vectors: relaunch with `MNEMOSYNE_EMBEDDING_DIM=<existing>`,
  - re-embed: `MNEMOSYNE_EMBEDDING_DIM=<configured> mnemosyne reindex` (backs up first);
- point at `mnemosyne doctor` to re-check.

Implemented in #PRPLACEHOLDER via a pure `_dim_mismatch_message(existing, configured)` helper (so the wording is unit-testable without sqlite-vec). Behavior unchanged: same ERROR level, same trigger, tables still untouched.

## Out of scope (separate follow-ups, happy to open if useful)
- `_get_embedding_dim` silently returning 384 for unlisted models — the upstream root cause; a fail-loud or auto-detect (probe the API) fix is behavior-changing and deserves its own PR.
- `mnemosyne doctor` doesn't detect the dim mismatch today (only integrity) — a read-only dim check would surface it proactively.
- Consolidate the other hardcoded 384 fallbacks: `polyphonic_recall.py` (`_EMB_DIM = 384`), `binary_vectors.py` (`EMBEDDING_DIM = 384`).

paoloantinori:
Closes #518.

## What
Rewrites the `init_beam` embedding-dimension-mismatch log message to be explicit and self-healing, so it is not misread as "database corrupt" (see #518 for the failure mode and observed impact).

## Why
The guard is correct — it leaves all memories and existing `vec0` tables untouched — but the old message ("dimension mismatch ... run `mnemosyne reindex`") gets misread as corruption, and users/agents abandon the store instead of reindexing. The new message states plainly that it is **not** corruption and gives both recovery paths with the exact commands.

## How
- Extract the message into a pure `_dim_mismatch_message(existing_dim, configured_dim)` helper (unit-testable without sqlite-vec).
- `init_beam` calls `logger.error(_dim_mismatch_message(existing_dim, EMBEDDING_DIM))`.
- Add `test_dim_mismatch_message_is_self_healing`.

## Behavior
Unchanged: same ERROR level, same trigger condition (`existing_dim != EMBEDDING_DIM`), `vec0` tables still left untouched. Pure refactor of the message text plus a new test — no change to `_get_embedding_dim` or the dim-resolution logic.

## Verification
`pytest tests/test_vec_dim_guard.py tests/test_embeddings_multilingual.py` → 15 passed, 3 skipped (sqlite-vec-gated). Deeper root-cause fixes (fail-loud/auto-detect dim, `doctor` dim check) are intentionally out of scope — tracked in #518.

CLA: happy to sign on review.

## Summary

- Refactors the BEAM initialization embedding-dimension mismatch error into a pure helper (`_dim_mismatch_message(existing_dim, configured_dim)`) and updates `init_beam()` to use it.
- The logged message now explicitly states the database/memories are **not corrupt** and provides exact, local recovery commands (two self-heal paths via `MNEMOSYNE_EMBEDDING_DIM=... mnemosyne reindex` and a follow-up `mnemosyne doctor` hint), while keeping the same guard behavior: it still refuses to create mismatched `sqlite-vec` virtual tables and leaves existing `vec0`/vector structures untouched.
- Adds a unit test that locks down the self-healing message content (including both dimension values and `mnemosyne reindex` guidance).

## Architectural impact

- **Core memory architecture (tiered BEAM):** No behavioral changes to working memory, episodic storage, or the BEAM vector/FTS hybrid retrieval pipeline. The only change is operational clarity when BEAM’s `sqlite-vec` `vec0` schema dimension doesn’t match the configured embedding dimension.
- **Retrieval strategies & consolidation pipeline:** Unchanged. This PR does not modify ranking, recall logic, FTS5 behavior, vector reads/writes, or any consolidation/veracity computations.
- **Veracity system:** Unchanged; no consolidation/veracity parameters or logic were touched.
- **Local-first & privacy posture:** Preserved. Recovery is explicitly local via Mnemosyne CLI commands; no added networking, sync, or data movement is introduced.
- **Sync layer:** Unchanged (no sync code paths modified).
- **Agent integration surfaces (Hermes, MCP, CLI):** No API surface changes. The improvement is limited to clearer operator feedback during BEAM schema initialization, which benefits CLI users and any agent frameworks that surface the same startup/log error.
- **Long-term maintainability:** Improves diagnostics maintainability by isolating the error text into a testable pure function, reducing risk of future wording regressions and making the “self-heal” guidance easier to evolve safely.

## Verification

- 15 tests passed; 3 skipped due to `sqlite-vec` gating.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -562,6 +562,32 @@ def _existing_vec_dim(conn: sqlite3.Connection) -> Optional[int]:
     return None
 
 
+def _dim_mismatch_message(existing_dim: int, configured_dim: int) -> str:
+    """Build the embedding-dimension-mismatch message.
+
+    Extracted as a pure function so the wording -- explicitly NOT corruption plus
+    the exact self-heal commands -- is unit-testable without a sqlite-vec
+    database. The phrasing matters: 'dimension mismatch' is otherwise misread as
+    'database corrupt', and users (and agent frameworks) abandon the store
+    instead of reindexing.
+    """
+    return (
+        f"Embedding dimension mismatch — NOT database corruption: your memories "
+        f"are intact, only the vector index is affected (recall falls back to "
+        f"keyword search until this is fixed). This database stores "
+        f"{existing_dim}-dim vectors but this process is configured for "
+        f"{configured_dim}-dim (MNEMOSYNE_EMBEDDING_DIM / "
+        f"MNEMOSYNE_EMBEDDING_MODEL); sqlite-vec tables were left untouched. To "
+        f"self-heal, choose ONE:\n"
+        f"  * Keep the existing {existing_dim}-dim vectors: relaunch with "
+        f"MNEMOSYNE_EMBEDDING_DIM={existing_dim} (and the matching model).\n"
+        f"  * Re-embed all memories at {configured_dim}-dim: run "
+        f"`MNEMOSYNE_EMBEDDING_DIM={configured_dim} mnemosyne reindex` (it backs "
+        f"up first).\n"
+        f"Run `mnemosyne doctor` to re-check."
+    )
+
+
 def init_beam(db_path: Path = None):
     """Initialize BEAM schema."""
     conn = _get_connection(db_path)
@@ -772,17 +798,7 @@ def init_beam(db_path: Path = None):
         existing_dim = _existing_vec_dim(conn)
         if existing_dim is not None and existing_dim != EMBEDDING_DIM:
             vec_dim_mismatch = True
-            logger.error(
-                "Embedding dimension mismatch: this database stores %d-dimensional "
-                "vectors, but the process is configured for %d "
-                "(MNEMOSYNE_EMBEDDING_DIM / embedding model). Not creating "
-                "sqlite-vec tables at the wrong dimension. Set "
-                "MNEMOSYNE_EMBEDDING_DIM / MNEMOSYNE_EMBEDDING_MODEL to match the "
-                "stored data, or run `mnemosyne reindex` to rebuild all vectors at "
-                "the configured dimension.",
-                existing_dim,
-                EMBEDDING_DIM,
-            )
+            logger.error(_dim_mismatch_message(existing_dim, EMBEDDING_DIM))
         else:
             try:
                 cursor.execute(f"""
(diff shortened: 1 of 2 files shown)


AxDSan:
@paoloantinori Welcome, and thank you for this fix! The dim-mismatch message extraction is exactly the right approach, and the test pins the critical "NOT corruption" phrasing. One thing before I can merge: the CLA check needs your signature. You can sign it at https://cla-assistant.io/mnemosyne-oss/mnemosyne?pullRequest=519 (takes 30 seconds, just click the link and sign with your GitHub account). Once signed, I will merge immediately.

paoloantinori:
done,  thank you!

PR #521, opened 2026-07-23
fix(embeddings): fail loud on unknown embedding model

paoloantinori:
Fail loud on unknown embedding models instead of silently assuming 384, per the direction @dplush outlined in #518. Accepted as a breaking change by @AxDSan (a silent 384 fallback that corrupts vector search is worse than a startup error).

A silent 384-dim fallback corrupts vector search when the real model dimension differs (the recurring per-model additions to the table, and the #519 Jina incident). This makes the resolver refuse to guess: an unknown model with no explicit `MNEMOSYNE_EMBEDDING_DIM` raises at import, before any vec0 DDL or vector write.

## Contract (per #518)
- [x] explicit valid `MNEMOSYNE_EMBEDDING_DIM` wins; known-model dims unchanged
- [x] unknown model + no explicit dim -> raise before any vec0 DDL / embedding write (no silent 384)
- [x] non-integer / non-positive `MNEMOSYNE_EMBEDDING_DIM` -> raise (config error); blank/empty treated as unset
- [x] embeddings-disabled -> keep the 384 fallback (unused path; CI/opt-out clean)
- [x] one authoritative resolver: Beam delegates to `embeddings._get_embedding_dim`; the duplicate fallback path is removed
- [x] direct-core vs Hermes-provider surfaces covered in fresh processes (see note below; not parity)
- YAML precedence/migration explicitly out of scope (#482)

## Surfaces: fail-fast vs graceful-degrade (NOT parity)
Direct core and `hermes_memory_provider` fail fast (non-zero exit at import). `mnemosyne_hermes` does NOT: its broad `except Exception` guard (`integrations/hermes/src/mnemosyne_hermes/__init__.py`) catches the configuration error, logs the actionable message, and disables the affected batch tools rather than crashing the provider. That is intentional (a provider should not crash on a mnemosyne config error) and is a known limitation, not parity with direct core. `test_unknown_model_parity_across_surfaces` asserts the actionable error is exposed on all three surfaces, non-zero exit for the two fail-fast ones, and exit 0 for `mnemosyne_hermes` (so a regression that crashes the provider is still caught).

## Blank / empty env vars
`MNEMOSYNE_EMBEDDING_DIM=` and `MNEMOSYNE_EMBEDDING_MODEL=` (set but empty/whitespace, common in Docker Compose `- VAR=${X}` with X unset, `.env` files, CI matrices) normalize to unset/default rather than raising even for a known model. Both use `.strip()` for consistent blank semantics.

## Tests
`pytest tests/test_embeddings_multilingual.py tests/test_embedding_dim_fresh_process.py` -> 25 passed. Covers known models, explicit override, non-integer/non-positive/blank DIM (raises), blank/whitespace MODEL (default), unknown model (raises), embeddings-disabled (384 fallback), and fresh-process fail-fast before any DB write across direct core, `hermes_memory_provider`, and `mnemosyne_hermes`.

## Relationship to #519
Complementary, not a replacement: this prevents new wrong-dimension stores; #519 makes the existing-store-mismatch guard message self-healing.

CLA: happy to sign.

## Summary

- Makes embedding-dimension resolution **fail fast** for unknown embedding models (instead of silently defaulting to 384), treating blank/whitespace-only `MNEMOSYNE_EMBEDDING_MODEL` / `MNEMOSYNE_EMBEDDING_DIM` as **unset**.
- Enforces that a set `MNEMOSYNE_EMBEDDING_DIM` must be a **positive integer**; non-integers and non-positive values raise immediately.
- Centralizes dimension resolution in `mnemosyne.core.embeddings._get_embedding_dim`; `mnemosyne.core.beam` (and related components like derived constants) now re-export the single source of truth.
- Preserves the historical **384 fallback only when dense embeddings are disabled** (so opt-out behavior doesn’t regress).
- Adds fresh-process tests that assert **import/startup fails before any local DB/vector initialization** for unsupported/unknown models, with **surface-specific** error handling across core vs Hermes.

## Architectural impact (core memory + local-first)

- **Core correctness win:** embedding dimensionality is validated and resolved up front before vector storage/table creation, preventing the wrong-dimension “fresh database” scenario that could silently corrupt vector search.
- **Local-first posture strengthened:** invalid embedding configuration now exits before any on-disk mutation; the new fresh-process tests explicitly check that no `*.db` artifacts are written prior to failure.
- **No changes to tiered memory / retrieval / consolidation / veracity / sync / benchmarks:** the behavior change is narrowly scoped to the **initial embedding dimension used for vector initialization** (and any downstream logic that depends on `EMBEDDING_DIM`); the pipelines themselves are not altered.

## Agent integration surfaces (Hermes, MCP/provider, CLI)

- **Direct core + provider-style processes:** import-time eager initialization enforces fail-loud behavior with actionable errors when a remote/custom embedding API is configured with a model not present in the built-in dimension table and `MNEMOSYNE_EMBEDDING_DIM` is missing.
- **`mnemosyne-hermes` wrapper:** maintains graceful degradation by design—init failures are captured, the affected provider/tooling is disabled/unavailable, and the overall Hermes agent process can stay alive.
- **CLI/docs alignment:** README/docs + changelog now clearly state the remote/custom embedding requirement that unknown models must specify `MNEMOSYNE_EMBEDDING_DIM` (and that blank env values behave as unset).

## Privacy posture

- No privacy model changes—this is configuration validation and early termination to avoid local state corruption (i.e., it **protects local-first integrity** rather than changing data handling semantics).

## Maintainability

- Reduces future drift by removing/avoiding duplicate embedding-dimension resolvers and ensuring Beam and other modules derive from `mnemosyne.core.embeddings.EMBEDDING_DIM`.
- Fresh-process regression tests lock in both **failure timing** (before DB/vector init) and **error semantics** consistency across core vs Hermes surfaces.

## Breaking change callout

- Using `MNEMOSYNE_EMBEDDING_API_URL` with an embedding model not in the built-in table now **requires** `MNEMOSYNE_EMBEDDING_DIM=<positive int>`; otherwise startup/import fails (with Hermes handling it via graceful provider unavailability).


dplush:
ciao Paolo 👋 — thank you for putting this draft together so quickly. It is very much aligned with the fail-loud direction, and having one authoritative resolver makes the behavior much easier to reason about.

Just one small edge case before it is ready: could we also require MNEMOSYNE_EMBEDDING_DIM to be a positive integer? At the moment, 0 or a negative integer would pass parsing but still is not a usable vector dimension.

tanti grazie 🙏

paoloantinori:
@dplush grazie! 🙏 Done in 84752f1 — `MNEMOSYNE_EMBEDDING_DIM` now must be a positive integer: `value <= 0` raises `ValueError("MNEMOSYNE_EMBEDDING_DIM=<n> must be a positive integer; vector dimensions are >= 1")`. Covered by `test_get_embedding_dim_non_positive_explicit_raises` (0, -1, -384). 17 passed / 3 skipped locally; CI is re-running on this push. Let me know if anything else before this is ready.

dplush:
Thanks Paolo — the positive-integer validation closes that edge, and the production resolver direction looks aligned with the #518 contract. I took a closer pass and think two small but important integration gates remain before this is ready:

1. Please add fresh-process coverage for unknown local, custom-endpoint, and API models with no explicit dimension. It should assert the actionable error happens before any `vec0` DDL or vector write. The current helper-level tests run after `embeddings` is already imported, so they do not cover that import/startup boundary.
2. Please add fresh-process parity coverage showing that direct core and both Hermes provider surfaces expose the same actionable configuration-failure outcome. The current parity fixture uses one interpreter and does not initialize an unknown model.

No YAML behavior needs to move into this PR. Once those are covered, a rebase onto current `main`, fresh CI, and then marking it ready would make it straightforward to review again. Thanks for carrying the draft forward.

paoloantinori:
Both covered in the latest push (rebased onto current main):

1. **Startup boundary** — `test_unknown_model_fails_at_import_before_any_write` spawns a fresh interpreter for an unknown local, custom-endpoint, and API model (no explicit dim) and asserts the actionable error fires at import (non-zero exit) *before* any `vec0` DDL: `MNEMOSYNE_DATA_DIR` is isolated under `tmp_path` and asserted to contain no `.db`.

2. **Parity** — `test_unknown_model_parity_across_surfaces` runs fresh-process for direct core, `hermes_memory_provider`, and `mnemosyne_hermes`. One intentional difference surfaced: direct core and `hermes_memory_provider` fail fast (non-zero exit at import), while `mnemosyne_hermes` catches the configuration error and graceful-degrades (logs the actionable message, disables batch tools, exit 0) — by design, so a provider doesn't crash on a mnemosyne config error. The test asserts the actionable error is exposed on all three surfaces and asserts non-zero exit for the two fail-fast ones, so a regression that swallows it is caught.

I also stripped the embedding-disable env flags (`MNEMOSYNE_NO_EMBEDDINGS`/`SKIP`/`OFF`) from the subprocess env, so a CI config with those set can't silently flip the resolver to the 384 fallback and false-pass the test.

Marking ready; fresh CI is running on the rebased branch.

AxDSan (changes requested):
The implementation is good. `embeddings.py:186-215` implements the #518 contract correctly — explicit valid dim wins, known-model table unchanged, unknown model + no explicit dim raises, `_is_disabled()` keeps the 384 fallback — and collapsing Beam's duplicate resolver to `_embeddings.EMBEDDING_DIM` at `beam.py:257-265` means the two can no longer drift. That second part is a real win independent of the fail-loud question.

### Blocking: blank env var becomes a hard startup failure

`MNEMOSYNE_EMBEDDING_DIM=` (set but empty) is treated as *explicit*, so it reaches the parse and raises `ValueError` at import — even for a known model, even with embeddings disabled. Empty-string env vars are extremely common in Docker Compose (`- MNEMOSYNE_EMBEDDING_DIM=${DIM}` with `DIM` unset), `.env` files, and CI matrices.

```python
if env_dim is not None and env_dim.strip():   # embeddings.py:186
```

Same treatment for blank `MNEMOSYNE_EMBEDDING_MODEL` — normalize to unset rather than "a model named empty string".

### One correction to the diff's framing

The crash doesn't actually originate at the new Beam line. `beam.py:44-54` imports `EMBEDDING_DIM` from `binary_vectors`, which at `binary_vectors.py:33-37` imports it from `embeddings` under `except ImportError` only, so the `ValueError` propagates there first. Same net outcome (fail before any vec0 DDL), but the comment about Beam being the delegation point overstates where the failure fires.

### This needs a maintainer decision, not just code review

#518 is **closed** (by merged #519). The fail-loud contract lives only in its comment thread and was endorsed by @dplush, not signed off by me. This is a breaking behavior change, and right now `README.md:359` and `docs/hermes-integration.md:11` advertise arbitrary model names via `MNEMOSYNE_EMBEDDING_API_URL` — which this turns into a hard startup failure.

My call: **accept the change**, but the same PR must carry the blank-env fix above, a CHANGELOG entry + version bump (`CONTRIBUTING.md:32`), and the docs update. I don't want users discovering this from a traceback.

### Tests

Genuinely non-vacuous: `test_get_embedding_dim_unknown_model_raises` fails on main (main returns 384 — literally the assertion this PR deletes), and `..._invalid_explicit_raises` / `..._non_positive_explicit_raises` fail on main because the old code `pass`ed on parse failure and returned `int(env)` unvalidated. `_disabled_falls_back` correctly pins the CI opt-out. The fresh-process assertions (`returncode != 0` + stderr marker) also fail on main — I verified `MNEMOSYNE_EMBEDDING_MODEL=some/unknown-model python -c 'from mnemosyne.core import beam'` exits 0 with `EMBEDDING_DIM=384`.

Two that don't do what they claim:

- **`_assert_fail_fast`'s `assert not list((tmp_path / "data").rglob("*.db"))` is vacuous.** I verified on main that `MNEMOSYNE_DATA_DIR=<tmp> python -c 'from mnemosyne.core import beam'` writes no `.db` at all, so it holds on unpatched code and proves nothing about "before any vec0 DDL or vector write" — which was @dplush's explicit review request #1. Only `returncode != 0` plus the stderr marker are load-bearing. A non-vacuous version has to reach `init_beam`.
- **`test_unknown_model_parity_across_surfaces[mnemosyne_hermes]`** passes `fail_fast=False` and asserts only that a marker appears in stderr — so it passes whether `mnemosyne_hermes` degrades gracefully (`integrations/hermes/src/mnemosyne_hermes/__init__.py:84-88` catches `except Exception` and logs a warning whose text reaches stderr via logging's lastResort handler) *or* crashes outright. The docstring claims it pins the catch; it doesn't distinguish the two.

Related and worth saying plainly in the body: those broad `except Exception` guards mean **the Hermes surface doesn't fail loud at all** — a misconfigured dimension becomes "memory tools silently missing, one WARNING line". That's a known limitation, not parity.

Nothing covers the blank-env case above, and nothing covers the diagnostics surface.

Nits: `raise ValueError(...)` inside `except (ValueError, TypeError):` has no `from None`, so the operator sees a "During handling of the above exception…" preamble above the message that's meant to be the readable part (Ruff B904). `TypeError` in that tuple is dead — `os.environ.get` returns `str | None` and `None` is already excluded by the `is not None` guard. Six subprocesses each doing a full mnemosyne import makes this a slow module; consider collapsing the parametrizations or gating behind a marker.

AxDSan:
@paoloantinori — maintainer decision on the open question, since #518 is closed and the fail-loud contract was never actually signed off by me: **I'm accepting it as a breaking change.** A silent 384-dim fallback that corrupts vector search is worse than a startup error, and the #519 Jina incident already showed what silent dimension drift costs us.

Conditions, all in this PR:

1. **The blank-env fix.** `MNEMOSYNE_EMBEDDING_DIM=` (set but empty) is currently treated as *explicit*, so it reaches the parse and raises even for a known model with embeddings disabled. Empty-string env vars are routine in Docker Compose (`- MNEMOSYNE_EMBEDDING_DIM=${DIM}` with `DIM` unset), `.env` files and CI matrices — this would break people who did nothing wrong.

   ```python
   if env_dim is not None and env_dim.strip():   # embeddings.py:186
   ```

   Same for blank `MNEMOSYNE_EMBEDDING_MODEL` — normalize to unset rather than treating it as a model named empty-string.

2. **Docs in the same commit.** `README.md:359` and `docs/hermes-integration.md:11` currently advertise arbitrary model names via `MNEMOSYNE_EMBEDDING_API_URL`. After this, that's a hard startup failure. Nobody should learn about this from a traceback.

3. **CHANGELOG entry + version bump** (`CONTRIBUTING.md:32`). Breaking changes especially.

Worth noting we just shipped exactly this class of bug in the other direction: #555 was a tool that hard-failed on *its own schema default*. Fail-loud is right, but only when the loud failure is reserved for genuinely invalid configuration.

Two corrections to the PR's framing, neither blocking:

- **The crash doesn't originate at the new Beam line.** `beam.py:44-54` imports `EMBEDDING_DIM` from `binary_vectors`, which at `binary_vectors.py:33-37` imports it from `embeddings` under `except ImportError` only — so the `ValueError` propagates there first. Same net outcome (failure precedes any vec0 DDL), but the comment about Beam being the delegation point overstates where it fires.
- **`_assert_fail_fast`'s `assert not list((tmp_path / "data").rglob("*.db"))` is vacuous.** I verified on `main` that `MNEMOSYNE_DATA_DIR=<tmp> python -c 'from mnemosyne.core import beam'` writes no `.db` at all, so it holds on unpatched code and proves nothing about ordering relative to vec0 DDL — which was @dplush's explicit review request. Only `returncode != 0` plus the stderr marker are load-bearing. A real version has to reach `init_beam`.

And one thing to say plainly in the PR body rather than describe as parity: the broad `except Exception` guards at `integrations/hermes/src/mnemosyne_hermes/__init__.py:84-88` mean **the Hermes surface doesn't fail loud at all** — a misconfigured dimension becomes "memory tools silently missing, one WARNING line". `test_unknown_model_parity_across_surfaces[mnemosyne_hermes]` passes whether that path degrades gracefully or crashes outright, so it doesn't pin the contract its docstring claims.

Collapsing Beam's duplicate resolver to `_embeddings.EMBEDDING_DIM` (`beam.py:257-265`) is a genuine win regardless of how the fail-loud question landed — that's the part that stops the two drifting again.

Nits: `raise ValueError(...)` inside `except (ValueError, TypeError):` has no `from None`, so operators get a "During handling of the above exception…" preamble above the message meant to be readable (Ruff B904). `TypeError` in that tuple is dead — `os.environ.get` returns `str | None` and `None` is already excluded by the guard.

Take it off draft when those are in and I'll merge it.

paoloantinori:
Heads-up @AxDSan: I opened #567 to cover the "diagnostics surface" gap you noted in this review. `collect_runtime_diagnostics` (the `doctor` surface) reported `embeddings_model` but not the resolved dimension; #567 adds `embeddings_dim` alongside it (plus a test), based on `main`.

I kept it separate so it doesn't enlarge the fail-loud diff you're currently reviewing. **Happy to fold #567 into this PR as a single change if you'd prefer** (it would ride the 3.16.0 bump already here); either way works for me.

dplush:
This PR now conflicts with current `main`. Once the outstanding integration coverage is in place, please rebase, resolve the conflicts, and push a refreshed head with fresh CI for review.

paoloantinori:
@AxDSan rebased onto current `main` (it had fallen a conflict behind) and confirmed all three of your merge conditions are in:

1. Blank-env fix — `MNEMOSYNE_EMBEDDING_DIM` / `MNEMOSYNE_EMBEDDING_MODEL` set-but-blank (common in Docker Compose / `.env` / CI) normalize to unset via `.strip()`; only an explicit invalid value raises.
2. Docs — README env-var table + `docs/hermes-integration.md` advertise the fail-loud behavior, including the direct-core vs `mnemosyne-hermes` (graceful-degrade) distinction.
3. CHANGELOG + version bump — `[3.16.0]` entry with the `**Breaking:**` note; `__version__` bumped to `3.16.0` (MINOR, per CONTRIBUTING.md Simple Versioning, for a user-facing behavior change).

CI is re-running on the rebased commits. Ready for re-review when green.

dplush (changes requested):
The resolver and fresh-process coverage now look good: blank env handling, the direct versus wrapper behavior, and pre-DB fail-fast coverage are all in place.

One documentation correction remains. `MNEMOSYNE_EMBEDDING_DIM` is described as an override only for models outside the built-in table in `README.md` and `docs/hermes-integration.md`. The implementation intentionally checks a valid explicit override first, so it takes precedence for every model, including a built-in-table model. Please update those descriptions to say that it is an optional positive-integer override which takes precedence over the built-in table, while retaining the blank-as-unset behavior. Also change “fails loud” to “fails loudly” in the affected wording.

No production-code change or new behavior test is needed for this documentation-only correction.

paoloantinori:
@dplush done — updated in f1ce28a:

- `MNEMOSYNE_EMBEDDING_DIM` now described as "Optional embedding dimension override (positive integer); takes precedence over the built-in model table" in the README env-var table.
- "fails loud" → "fails loudly" in the profile rows (README + hermes-integration).
- Also rebased onto current main (resolved CHANGELOG + hermes-integration conflicts) and bumped the plugin manifest to match 3.16.0.

CI re-running. Ready for re-review when green.

dplush:
Thanks Paolo — the rebase cleanup, version-surface alignment, and current CI all look good. I also ran the focused embedding/fresh-process tests and docs verification against the exact head successfully.

Two small release-documentation gates remain before I can give the final approval:

1. `README.md` points to `docs/configuration.md` as the full configuration reference, but that page still describes embedding dimensions as broadly auto-detected and does not document the new positive-integer override, blank-as-unset, and fail-loud behavior for unknown models. Please bring that reference in line with the README and Hermes guide. If the generated API configuration page needs its source metadata updated as well, please regenerate it rather than editing generated output by hand.

2. The `[3.16.0] - 2026-08-03` section is not released or tagged yet. Please either keep this release note under `Unreleased` until the release date is known, or set the date when 3.16.0 is actually released. It would also help to add one short upgrade note for stores created under the old silent-384 fallback: setting the model’s true dimension can surface the existing-store mismatch guard, so operators may need the documented reindex/recovery path rather than treating the override as a one-step fix.

These are release accuracy and upgrade-path items, not new production-code scope. Once they are updated on a fresh head with CI, I can complete the final gate.

dplush:
@paoloantinori please rebase this one also for a final teview.

dplush (changes requested):
The implementation and focused checks look good. One release-documentation blocker remains before approval:

`CHANGELOG.md` has an orphaned malformed entry under `Unreleased > Fixed` (`- - \`raise ValueError(...)\` ...`). It appears to be a rebase artifact: the double bullet is invalid Markdown structure, and the sentence is separated from the #521 change without a PR reference or release context.

Please remove it or fold it into a correctly scoped #521 changelog entry, then push a fresh head with CI. After that I expect this to be ready for final approval.

paoloantinori:
@dplush rebased onto current main and addressed both documentation gates:

1. docs/configuration.md: documents the embedding dim resolution order (explicit positive-integer override wins, then built-in table, then fail-loud for unknown). Documents blank-as-unset. Added upgrade note for stores created under the old silent-384 fallback (mismatch guard may trigger; use reindex/recovery path).
2. CHANGELOG: moved fail-loud entry from [3.16.0] to [Unreleased] with the upgrade note (no tag yet).

CI green. Ready for final review.

dplush:
@paoloantinori Thanks, the original documentation gates are addressed. One tiny release-doc cleanup remains from the final review: please remove the orphaned double-bullet line in `CHANGELOG.md` under `Unreleased > Fixed` (the line beginning `- - raise ValueError(...)`), or fold it into a properly scoped #521 entry. It appears to be a rebase artifact. Please push that minimal correction with fresh CI; then I expect this to be ready for approval.

paoloantinori:
@dplush removed the orphaned B904 double-bullet entry — it was a rebase artifact from the [3.16.0] → [Unreleased] merge. The B904 cleanup is a trivial lint detail already covered by the fail-loud entry under ### Changed. CI green. Ready for final approval.

dplush (approved):
Final review on c6fe8de is clean. The release-note artifact is removed; the Unreleased entry and upgrade guidance are accurate. Focused fresh-process coverage, docs, full CI, and an independent review are all green. Approving for merge.

AxDSan (commented):
@paoloantinori good work on this. The code is solid after all the iterations. Two things still blocking:\n\n1. Merge conflict in docs/hermes-integration.md needs a rebase\n2. dplush's doc wording feedback (MNEMOSYNE_EMBEDDING_DIM description) still needs addressing\n\nCan you rebase and address those? Happy to merge once that's done.

paoloantinori:
@AxDSan thanks! I believe this crossed with the merge (merged 11:36, your comment landed 12:29): both items were already in the head that went in. The hermes-integration conflict was cleared by the final rebase (`c6fe8de` merged clean, and the surface-specific fail-loud note is on main), and dplush's wording landed earlier in `f1ce28a` (README now reads "Optional embedding dimension override (positive integer); takes precedence over the built-in model table"). Nothing was left unaddressed from my side. I've also opened #748 to pick up the remaining CodeRabbit minors (e2e success-path coverage for the explicit-dimension contract and the privacy note for `MNEMOSYNE_EMBEDDING_API_URL`) if you want them.

PR #522, opened 2026-07-23
fix: isolate enhanced recall cache requests

### LINKED ISSUE #513: [BUG] Enhanced Recall query cache ignores effective ranking options and filters
## Summary

Enhanced Recall's persistent `QueryCache` identifies an entry only by the normalized query text, while the result set also depends on ranking configuration and recall filters.

## Source audit

`BeamMemory.recall_enhanced()` currently does:

```python
cached = self._query_cache.get(original_query)
# ... enhanced recall using kwargs and feature options ...
self._query_cache.put(original_query, results)
```

The cache key therefore excludes, among other inputs:

- resolved vector / FTS / importance weights;
- `top_k`;
- temporal settings;
- source, topic, date, author, and channel filters;
- Enhanced Recall ranking options such as MMR, Weibull, synonym expansion, and associative retrieval.

A later call with the same query but different inputs can return an incompatible cached result. Because the cache database is stored next to the main database rather than per call, this is also an isolation concern for filtered recalls.

## Expected behavior

A cache hit must be valid for the full effective recall request, including effective ranking settings and filters. Equally safe alternatives are to bypass the cache for non-default/filtered calls, or to use a versioned structured cache key that captures every result-affecting input.

## Suggested scope

Keep this separate from #511, which is only about resolving omitted hybrid weights without crashing the public API. A follow-up should define the cache-key contract and add direct cache-hit/miss regression coverage for distinct weight/filter configurations.

dplush:
## Summary

Fixes #513 by making Enhanced Recall's persistent cache identity depend on the full effective request rather than only normalized query text.

- use a canonical, opaque `v2:` SHA-256 request key for Enhanced Recall
- isolate database namespace, scope, filters, ranking/temporal inputs, and enabled pipeline options
- route valid opaque keys through exact-only cache access, bypassing all semantic/fuzzy tiers
- preserve the legacy generic `QueryCache` behavior for ordinary strings beginning with `v2:`
- bypass cache reads and writes for `explain=True` and `use_cache=False`

## Verification

```text
uv run --extra test pytest -q tests/test_enhanced_recall_cache.py tests/test_query_cache_synonyms.py tests/test_query_intent_recall.py tests/test_config_cross_session_runtime.py
25 passed in 18.23s
```

Also checked with `git diff --check` and an independent read-only review.

## Summary

- Prevents Enhanced Recall cache collisions by deriving a canonical opaque `v2:` SHA-256 key from the full effective request, including scope, filters, ranking, temporal inputs, and pipeline options.
- Adds exact-only opaque cache access that bypasses semantic/fuzzy tiers, while preserving legacy behavior for ordinary `v2:`-prefixed strings.
- Bypasses cache reads and writes for `explain=True` and `use_cache=False`.
- Adds regression coverage for request variants, pipeline configuration, session/database isolation, persistence, legacy compatibility, and bypass behavior.

## Architectural impact

This strengthens BEAM/Enhanced Recall retrieval without changing working or episodic memory tiers, consolidation, veracity, synchronization, or agent integration surfaces such as Hermes, MCP, and the CLI. It preserves local SQLite-backed caching and therefore maintains Mnemosyne’s local-first and privacy posture: no new external service, telemetry path, or plaintext query data is introduced into the opaque key.

The exact request identity is the right call for correctness and maintainability. Dedicated opaque storage, TTL handling, eviction, persistence, and statistics keep enhanced retrieval isolated from fuzzy query semantics and make future result-affecting options easier to include safely. The expanded targeted tests provide focused regression coverage, though they do not represent a broader benchmark methodology change.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -6662,6 +6662,126 @@ class BeamMemory:
 
         return final_results
 
+    def _enhanced_recall_cache_key(
+        self,
+        *,
+        original_query: str,
+        expanded_query: str,
+        top_k: int,
+        runtime: Any,
+        use_weibull: bool,
+        use_mmr: bool,
+        use_intent: bool,
+        use_synonyms: bool,
+        use_associative: bool,
+        associative_depth: int,
+        mmr_lambda: float,
+        recall_kwargs: Dict[str, Any],
+    ) -> str:
+        """Build a versioned opaque key for one effective enhanced request."""
+        def canonicalize(value: Any) -> Any:
+            if isinstance(value, datetime):
+                return _normalize_datetime_utc(value).isoformat()
+            if isinstance(value, Path):
+                return str(value.resolve())
+            if isinstance(value, dict):
+                return {str(key): canonicalize(val) for key, val in value.items()}
+            if isinstance(value, (list, tuple)):
+                return [canonicalize(item) for item in value]
+            if isinstance(value, set):
+                return sorted(canonicalize(item) for item in value)
+            if value is None or isinstance(value, (str, int, float, bool)):
+                return value
+            return str(value)
+
+        raw_weights = (
+            recall_kwargs.get("vec_weight"),
+            recall_kwargs.get("fts_weight"),
+            recall_kwargs.get("importance_weight"),
+        )
+        resolved_weights = _normalize_weights(*raw_weights)
+        if (all(weight is None for weight in raw_weights)
+                and os.environ.get("MNEMOSYNE_QUERY_INTENT", "0") == "1"
+                and classify_intent is not None and adjust_weights is not None):
+            try:
+                resolved_weights = adjust_weights(
+                    base_vec=resolved_weights[0],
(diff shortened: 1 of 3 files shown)


dplush:
I also ran a small isolated, BEAM-grounded cache-correctness A/B against the parent commit.

Using official BEAM 100K/chat 1 (188 messages), explicit Enhanced Recall weights, and 21 deliberately different effective-request cases:
- parent baseline: 18/21 warm-cache results differed from an independently computed cold reference;
- this PR: 0/21 incorrect reuses;
- identical-request controls retained 21/21 observable cache hits.

The mismatches reproduced for `top_k`, source/date filters, and sibling database namespace. This is intentionally a cache-correctness result only — not a Hit@k, retrieval-quality, or overall latency claim.

dplush:
@AxDSan I opened #523 for the persisted QueryCache TTL/eviction lifecycle follow-up raised during review. Since it touches the same `query_cache.py` area, I’ll keep it intentionally separate and prepare the implementation only after this PR merges, so #522 stays focused and does not become a stack.

PR #531, opened 2026-07-24
feat(hermes): add polyphonic_recall config key so working memory can be vector-searched

okalentiev:
## What

Adds a `memory.mnemosyne.polyphonic_recall` config key to both Hermes providers, routing recall through the polyphonic engine so working memory is searched by vector similarity rather than FTS5 keywords alone.

**Default `false` — current behavior is unchanged.** Per review, the default is left as a maintainer decision pending a quality/latency check.

## Why

`BeamMemory.recall`'s linear scorer only vector-searches *episodic* memory:

> Uses sqlite-vec + FTS5 for episodic, FTS5 for working.

Every memory starts in working memory, so until `sleep()` promotes it, recall matches literal keyword overlap only. A paraphrased question misses the memory even when it is a strong embedding match.

### Reproduction (synthetic, fresh home)

```python
p.handle_tool_call("mnemosyne_remember", {
    "content": "Casey is vegan; shared meals need genuinely vegan options.",
    "source": "fact", "importance": 0.9, "scope": "global"})

recall("plant based diet, avoid animal products")   # -> 0 results
recall("vegan")                                     # -> 1 result (literal keyword)
```

With the key enabled, the paraphrase returns the memory.

Storage is healthy — the gap is only in retrieval:

| check | result |
|---|---|
| `cos(fresh_embed(text), stored_vector)` | 1.0 — stored vector is genuine fastembed output |
| `cos(embed("plant based diet…"), stored_vector)` | 0.614 — strong semantic match present |
| raw `vec_working` int8 KNN for that query | ranks the memory #1 |
| `recall(...)` on the linear path | 0 results |

`recall(..., vec_weight=1.0, fts_weight=0.0, explain=True)` shows the vector stage running and discarding everything:

```json
"embedding": {"available": true, "computed": true},
"stages": [{"name": "wm_primary", "raw_count": 9, "after_filter_count": 9, "kept_count": 0}]
```

## Change

- New `polyphonic_recall` key (default `false`), advertised via `get_config_schema()`, following the existing `kwargs > config.yaml > default` precedence used by `profile_isolation`.
- The provider records the value it writes to `MNEMOSYNE_POLYPHONIC_RECALL` and refreshes it on re-initialization, so a second profile / session switch / edited config applies. A value the provider did **not** write stays authoritative. Comparing against the written value (rather than snapshotting at import) keeps this correct when `.env` loads after import.
- Applied to both shipped provider copies (`hermes_memory_provider/`, `integrations/hermes/src/mnemosyne_hermes/`).

## Tests

`tests/test_polyphonic_recall_default.py` — 34 tests, parametrized across **both** shipped providers (the pip package is skipped *visibly* with a reason when the integration isn't installed, rather than silently shrinking coverage; CI installs it via `pip install -e ./integrations/hermes`).

- config/env matrix: default, config on/off, kwargs precedence, string coercion, explicit-env-var-wins both directions, schema advertisement
- same-process default-then-override regression (the issue raised in review)
- external-override-survives-reinitialization
- **retrieval behavior**: stores an unconsolidated fact and asserts a paraphrased query finds it when enabled, and does not on the linear path — following the `test_ab_toggles.py` convention that a toggle must surface in real recall results

```
pytest tests/ -k "hermes or recall or polyphonic or config"   ->  635 passed
```

Stable across randomized orderings. No existing tests modified.

## Not included

I have not benchmarked polyphonic vs linear for recall quality or latency at scale — the evidence here is correctness on the case above. Happy to run an A/B if that would help settle the default.

## Summary

Adds the Hermes provider configuration key `memory.mnemosyne.polyphonic_recall`, with a default of **false**. When enabled, working-memory recall uses Mnemosyne’s polyphonic recall engine and can use vector similarity to retrieve unconsolidated entries from paraphrased queries. The default preserves existing recall behavior.

The setting follows **kwargs > config.yaml > default** precedence. The providers manage `MNEMOSYNE_POLYPHONIC_RECALL` while preserving externally supplied environment values and refreshing provider-managed values during reinitialization.

## Architecture and integrations

- **Tiered memory:** Changes working-memory recall only. The PR does not change working-to-episodic consolidation, `sleep()`, BEAM, or storage semantics.
- **Retrieval:** Adds an optional vector-based recall path for unconsolidated working-memory entries.
- **Consolidation, veracity, and sync:** No changes.
- **Hermes:** Both providers expose the setting through `get_config_schema()` and apply consistent configuration and environment precedence.
- **MCP and CLI:** No direct changes.

## Privacy and local-first guarantees

The change remains in-process. It adds no network calls, telemetry, external data flow, or cross-agent sharing. It preserves Mnemosyne’s local-first design.

## Maintainability and risks

- Adds schema documentation and changelog coverage.
- Adds tests for defaults, precedence, string coercion, environment handling, reinitialization, schema advertisement, and paraphrased retrieval.
- Runs retrieval tests in a subprocess to isolate embedding resources, SQLite state, and file descriptors.
- Uses atomic updates for provider-managed environment values.
- A broader configuration-scope issue remains: both providers mutate the process-global `MNEMOSYNE_POLYPHONIC_RECALL` switch while tracking managed state independently. Conflicting provider settings can affect all active `BeamMemory` instances, and the last initialized provider can control the process-wide recall mode.
- The Python 3.13 `test_repair.py` failure remains unresolved. Retrieval assertions and the process-wide configuration model require further maintainer review.
- No benchmark methodology changes measure the possible recall, latency, or resource-use impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [Unreleased]
+
+### Changed
+
+- **Hermes provider opts into polyphonic recall by default.** The linear recall scorer only vector-searches episodic memory (`Uses sqlite-vec + FTS5 for episodic, FTS5 for working` — `BeamMemory.recall`), and every memory starts in working memory, so until `sleep()` promotes it a memory is retrievable by literal keyword overlap only — a paraphrased query misses it even when the stored embedding is a strong match (0.61 cosine in the reproduction, ranked #1 by the underlying `vec_working` KNN). Routing Hermes recall through the polyphonic engine gives working memory a vector voice. New `memory.mnemosyne.polyphonic_recall` config key (default `true`); an explicit `MNEMOSYNE_POLYPHONIC_RECALL` env var still takes precedence, so existing opt-ins and opt-outs are preserved.
+
 ## [3.15.0] - 2026-07-20
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
-### Changed
+### Added
 
-- **Hermes provider opts into polyphonic recall by default.** The linear recall scorer only vector-searches episodic memory (`Uses sqlite-vec + FTS5 for episodic, FTS5 for working` — `BeamMemory.recall`), and every memory starts in working memory, so until `sleep()` promotes it a memory is retrievable by literal keyword overlap only — a paraphrased query misses it even when the stored embedding is a strong match (0.61 cosine in the reproduction, ranked #1 by the underlying `vec_working` KNN). Routing Hermes recall through the polyphonic engine gives working memory a vector voice. New `memory.mnemosyne.polyphonic_recall` config key (default `true`); an explicit `MNEMOSYNE_POLYPHONIC_RECALL` env var still takes precedence, so existing opt-ins and opt-outs are preserved.
+- **`memory.mnemosyne.polyphonic_recall` config key for the Hermes providers.** Routes recall through the polyphonic engine so working memory gets a vector voice. Default `false` — current behavior is unchanged — with an explicit `MNEMOSYNE_POLYPHONIC_RECALL` env var still taking precedence. Enable it when paraphrased queries fail to retrieve recent memories: the linear scorer only vector-searches episodic memory (`Uses sqlite-vec + FTS5 for episodic, FTS5 for working` — `BeamMemory.recall`) and every memory starts in working memory, so until `sleep()` promotes it a memory is reachable by literal keyword overlap only — even when the stored embedding is a strong match (0.61 cosine in the reproduction, ranked #1 by the underlying `vec_working` KNN). Values written by the provider are refreshed on re-initialization rather than mistaken for an operator override.
 
 ## [3.15.0] - 2026-07-20
 
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1325,7 +1325,18 @@ _POLYPHONIC_ENV = "MNEMOSYNE_POLYPHONIC_RECALL"
 # operator-supplied one (must never be clobbered). Comparing against the
 # written value keeps this independent of import order, so an env var loaded
 # from .env after this module is imported still counts as external.
+#
+# _polyphonic_lock serializes the read-check-write below across threads.
+# Hermes' MemoryManager runs background sync/prefetch work on daemon
+# threads, so two provider instances in one profile process can call
+# _apply_provider_config() concurrently. In practice both always resolve
+# the same target (kwargs are never used for this key in the framework;
+# every instance in a process reads the same config.yaml), so the race is
+# a redundant identical write rather than a wrong one -- but making the
+# check-then-write atomic removes the question entirely rather than
+# relying on that being true forever.
 _provider_managed_polyphonic: Optional[str] = None
+_polyphonic_lock = threading.Lock()
 
 
 class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
@@ -1611,19 +1622,20 @@ class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
         # this provider wrote earlier is refreshed, so re-initializing with a
         # different config (session switch, second profile in-process) applies.
         global _provider_managed_polyphonic
-        _current_polyphonic = os.environ.get(_POLYPHONIC_ENV)
-        if _current_polyphonic is None or _current_polyphonic == _provider_managed_polyphonic:
-            polyphonic = kwargs.get("polyphonic_recall")
-            if polyphonic is None:
-                polyphonic = self._read_config_key("polyphonic_recall")
-            if polyphonic is None:
-                polyphonic = False
-            if isinstance(polyphonic, str):
-                polyphonic_enabled = polyphonic.strip().lower() in ("true", "1", "yes", "on")
-            else:
-                polyphonic_enabled = bool(polyphonic)
-            _provider_managed_polyphonic = "1" if polyphonic_enabled else "0"
-            os.environ[_POLYPHONIC_ENV] = _provider_managed_polyphonic
+        with _polyphonic_lock:
+            _current_polyphonic = os.environ.get(_POLYPHONIC_ENV)
+            if _current_polyphonic is None or _current_polyphonic == _provider_managed_polyphonic:
+                polyphonic = kwargs.get("polyphonic_recall")
+                if polyphonic is None:
+                    polyphonic = self._read_config_key("polyphonic_recall")
+                if polyphonic is None:
+                    polyphonic = False
+                if isinstance(polyphonic, str):
+                    polyphonic_enabled = polyphonic.strip().lower() in ("true", "1", "yes", "on")
+                else:
+                    polyphonic_enabled = bool(polyphonic)
+                _provider_managed_polyphonic = "1" if polyphonic_enabled else "0"
+                os.environ[_POLYPHONIC_ENV] = _provider_managed_polyphonic
 
         shared_surface_path = kwargs.get("shared_surface_path")
         if shared_surface_path is None:
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1318,6 +1318,16 @@ def _parse_env_optional_int(key: str, default: Optional[int]) -> Optional[int]:
     return _coerce_optional_int(os.environ.get(key), default)
 
 
+_POLYPHONIC_ENV = "MNEMOSYNE_POLYPHONIC_RECALL"
+
+# Last value this process wrote to _POLYPHONIC_ENV. Used to tell a
+# provider-managed value (safe to refresh on re-initialization) apart from an
+# operator-supplied one (must never be clobbered). Comparing against the
+# written value keeps this independent of import order, so an env var loaded
+# from .env after this module is imported still counts as external.
+_provider_managed_polyphonic: Optional[str] = None
+
+
 class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
     """Mnemosyne native memory — local SQLite with vector + FTS5 hybrid search."""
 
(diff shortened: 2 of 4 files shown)


dplush:
Nice work — the repro makes the working-memory gap very clear, and the config key is useful regardless of where the default lands.

Two small things I would want to settle before changing the Hermes default:

1. The provider currently writes `MNEMOSYNE_POLYPHONIC_RECALL` itself, so a later provider initialization can mistake that managed value for an external operator override. A same-process default-on → config-off regression test would catch this.

2. Could we add one provider-level behavior test for the promised outcome: an unconsolidated fact should be found by a paraphrased query with polyphonic enabled, while the opt-out preserves the current linear path? It would also be good for both shipped provider implementations to be required rather than silently skipped.

Because default-on changes recall semantics and may affect latency/resource use for existing Hermes installs, I would lean toward shipping the config key first with the current behavior preserved, then making the default a maintainer decision after a small quality/latency check.

Happy to help with the focused test or an A/B if useful.

okalentiev:
Thanks — all three points taken, pushed in a4f36b7.

**1. Provider-written value mistaken for an operator override.** Confirmed and fixed. Reproduced it first:

```
after default init           : 1
after re-init w/ config=False: 1   <- should be 0
```

The provider now records the value it writes and refreshes that, while a value it did not write stays authoritative. I compare against the written value rather than snapshotting at import, so an env var loaded from `.env` *after* the module is imported still counts as external. Regression test added (`test_reinitialization_applies_new_config`), plus `test_external_override_survives_reinitialization` for the other direction.

**2. Behavior test + required provider parity.** Added `TestPolyphonicRecallRetrieval`, parametrized over both providers: it stores an unconsolidated fact, asserts a paraphrased query retrieves it before `sleep()` when enabled, and asserts the opt-out preserves the linear path — following the `test_ab_toggles.py` convention that a toggle should surface in real recall results. The pip provider is no longer silently skipped: it is a required param, skipped *visibly with a reason* only when the integration isn't installed (CI installs it).

Worth flagging that writing this caught a genuine order-dependent bug in my own test: resetting the managed value via `sys.modules[...]` broke when another module reloads the provider, so the reset now resolves through the function's own `__globals__`. It failed in the full suite while passing in isolation.

**3. Default.** Agreed — flipped to `false`, so this ships as the config key with current behavior preserved and the default left to you. That matches where I landed too: the engine's gate is documented as *"Zero behavior change for production"*, and I only have correctness evidence on the case above, not a quality or latency comparison.

I would take you up on the A/B offer. What I can say from running it on 8 Hermes profiles is that paraphrased recall starts working and keyword recall is unaffected; what I cannot say is what it costs at scale or how it behaves on larger stores. If you have a benchmark harness you would trust for that, I am happy to run it and post numbers here.

`pytest tests/ -k "hermes or recall or polyphonic or config"` → 635 passed, stable across randomized orderings.

okalentiev:
Two follow-ups from the last push (1e51d2b), one of which is a caution for anyone running the suite locally.

**Test isolation bug in my own retrieval test — fixed.** Passing `hermes_home=tmp_path` to `initialize()` is *not* sufficient isolation: parts of the database path resolve through the `HERMES_HOME` environment variable, so the test read and wrote the developer's real Mnemosyne store. On my machine it silently inserted the fixture (`"Casey is vegan..."`) into my own working memory — three rows — and left connections open against that database. Now pinned via `monkeypatch.setenv("HERMES_HOME"/"MNEMOSYNE_DATA_DIR", tmp_path)`, verified by asserting the real store's row count is unchanged across a full run of the file.

Might be worth a session-scoped autouse guard in `conftest.py` for the whole suite, if other provider-level tests construct providers the same way. Happy to send that separately if you want it.

**`test (3.13)` failure on a4f36b7.** It is `tests/test_repair.py::test_sidecar_created_after_preflight_is_rejected_before_inode_binding`, failing on `assert len(os.listdir("/proc/self/fd")) == fd_before` with `17 == 130` — a process-wide FD count, in a file my change does not touch. 3.10/3.11/3.12 passed the same commit, and `main` is green, so this looks order-dependent rather than deterministic: my retrieval test is likely the first to load the embedding model and open database handles, which perturbs that global assertion under some `pytest-randomly` seeds.

I could not reproduce it locally — those tests skip on macOS (no `/proc`) — and I do not have rerun rights on the repo. The isolation fix above should reduce the footprint, but if 3.13 fails again I would rather not paper over it: tell me which you prefer and I will do it —

1. gate the retrieval test behind a marker/opt-in env flag so it does not run in the default suite, or
2. make it use a stub embedding backend instead of loading the real model, or
3. leave it and treat the FD assertion in `test_repair` as the thing to harden.

I lean towards (2) if the goal is keeping real retrieval coverage in the default suite, but it is your call.

okalentiev:
Update on the `test (3.13)` failure — I dug in, and I no longer think my file descriptors are the cause. Handing this over rather than keep guessing at your CI.

**What the failure actually is.** Only 3.13 genuinely fails: `test_repair.py::test_sidecar_created_after_preflight_is_rejected_before_inode_binding`, on

```python
assert len(os.listdir("/proc/self/fd")) == fd_before   # assert 17 == 130
```

`test (3.10)` and `test (3.12)` in the last run were fail-fast cancellations (`The operation was canceled`), not failures.

**Evidence it is not my test's descriptors.** The assertion values are *identical* — `17 == 130` — across three commits with materially different FD behavior, including 38540c1 where the retrieval coverage runs entirely in a child process and leaks nothing into pytest. If my test contributed to `fd_before`, that number would have moved. It did not move at all.

Since `pytest-randomly` is not in the test extra, collection order is deterministic, which is consistent with the exact same numbers each run.

**My hypothesis** (unverified — those tests skip on macOS, so I cannot reproduce locally, and I do not have rerun rights): the assertion compares a process-wide FD count across a window in which a GC cycle can run. Adding any test module ahead of `test_repair.py` alphabetically changes allocation/GC timing, so descriptors held by earlier tests get collected *during* `run_repair` and the count collapses. That would make the assertion sensitive to unrelated additions rather than to what the test is actually guarding.

If that is right, the durable fix is on the `test_repair` side — e.g. `gc.collect()` immediately before sampling `fd_before`, or counting only the descriptors the repair itself opens rather than the whole process. Happy to send that as a separate PR if you agree with the diagnosis; I did not want to touch a security-ish test in this PR without your say-so.

**What I would like to know:** would you prefer I

1. leave the retrieval test in a subprocess (current state — most isolated, ~40 lines of harness), or
2. revert it to the simpler in-process version (the `HERMES_HOME` pin from 1e51d2b still keeps it off any real store), given the subprocess did not change the CI outcome?

I lean towards (2) now that isolation is demonstrably not what is failing, but it is your suite and your call. Everything else is green, and locally `pytest tests/ -k "hermes or recall or polyphonic or config"` is 635 passed with the real store's row count unchanged.

okalentiev:
Did a strict self-audit before asking for more review time (6b4a3ec). Two findings, both fixed:

**1. The check-then-write on `_provider_managed_polyphonic` was not atomic.** `MemoryManager` runs background sync/prefetch on daemon threads, so two provider instances in one profile process can call `_apply_provider_config()` concurrently. Reproduced with 128 threads split between conflicting configs, fired concurrently — threads observed both `'0'` and `'1'` mid-race.

In the actual integration this can't produce a *wrong* outcome (`polyphonic_recall` is never passed via kwargs by the framework, only read from `config.yaml`, so every concurrent init in a process targets the same value from the same source) — but that safety lives in caller behavior, not in this code, and I'd rather not ship a known non-atomic read-modify-write on shared state and rely on "nothing calls it that way yet." There's already a `threading.Lock()` precedent in this file for the same class of problem, so I added a dedicated one (not `_provider_lock` — that guards an unrelated counter and isn't reentrant; reusing it would risk deadlock if this path is ever reached while it's held). Re-ran the 128-thread adversarial test ×5 against the locked code: always a clean last-writer-wins value, never torn.

**2. Dead `monkeypatch` fixture parameter** in the retrieval test, left over from before the subprocess harness (env pinning moved into `_run_child` in 38540c1; the parameter was never referenced in the body after that). Removed.

`pytest tests/ -k "hermes or recall or polyphonic or config"`: 635 passed, stable across two full-suite runs. Real store row count unchanged before/after.

Still open: the `test (3.13)` / `test_repair.py` FD-count question from my last comment — no update there, still think it's worth your eyes rather than mine guessing further.

dplush:
Thanks — the working-memory retrieval gap and the default-off rollout both make sense.

I found one blocking configuration-scope issue on the current head: both provider modules track their own “provider-managed” value, but mutate the same process-global `MNEMOSYNE_POLYPHONIC_RECALL`. In a process that loads both providers, a value written by one is treated as an external override by the other; I reproduced legacy=`true` followed by pip-provider=`false` still leaving the flag at `1`. More broadly, the last initialized provider controls recall mode for every active `BeamMemory` instance. The lock removes a write race, but not that cross-provider/profile semantic conflict.

Could we settle the intended contract before iterating further?
- If this is a process-wide switch, model and document it as a core/runtime setting rather than a provider/profile key.
- If it is intended to be per provider/profile, pass it as instance-bound recall configuration into `BeamMemory` and keep the environment variable only as an external fallback — providers should not mutate it at runtime.

Two smaller merge gates:
1. strengthen the retrieval test to assert the stored memory ID/content (plus a distractor), not only a non-zero result count;
2. resolve the current Python 3.13 CI failure with fresh green matrix evidence.

I’m happy to review the focused follow-up.

AxDSan (changes requested):
The diagnosis in the body is correct and I verified it: `BeamMemory.recall` reads the gate per call at `beam.py:5490` (`os.environ.get("MNEMOSYNE_POLYPHONIC_RECALL", "0") == "1"`), so a provider writing the env var during `_apply_provider_config` genuinely does change recall behavior — no import-time-constant problem here, #482 doesn't apply. The polyphonic vector voice does cover working memory (`polyphonic_recall.py:_vector_voice` reads `memory_embeddings` for both WM and EM), so enabling it does make the paraphrase retrievable. The diff is faithful to its description: identical block in both provider copies, `get_config_schema()` advertises the key in both, default false, external env wins.

**But it fixes the symptom, not the cause — and I'd rather take the cause.**

The actual defect is in the linear path. At `beam.py:5765` the keep-gate is:

```python
if relevance >= row_min_relevance or (wm_ranks and len(query_words) <= 1 and relevance > 0)
```

A working-memory row that arrived *only* from `_wm_vec_search` — i.e. it's in `wm_vec_sims` but not in `wm_ranks` (`beam.py:5679-5685`) — still has to clear a lexical relevance floor it can't reach, so a true semantic paraphrase match with zero word overlap is discarded. That's a ~2-line fix at `beam.py:5765`, it needs no new config key, and it fixes MCP and CLI users too, not just Hermes.

Turning on the whole polyphonic pipeline as the workaround is a big hammer: it's a process-global env var that leaks into any subprocess Hermes spawns, and it changes ranking for every query, not just the zero-overlap ones.

Two problems with the approach even on its own terms:

- **The env-var handshake can still clobber an operator.** Config `true` → provider writes `"1"`; an operator independently exports `"1"`; config is later edited to `false` → re-init sees `env == managed` and overwrites the operator's explicit `"1"` with `"0"`. Inherent to value-comparison tracking; another argument for instance-bound config over env mutation.
- The var is written during `_apply_provider_config` regardless of whether initialization is later skipped (`skip_contexts`) or fails.

### Tests

The config/env matrix tests (`TestPolyphonicRecallDefault`) are real — they assert `os.environ[ENV_VAR]` after `_apply_provider_config`, which raises `KeyError` on unpatched main since nothing writes that var today. `test_reinitialization_applies_new_config` and `test_external_override_survives_reinitialization` genuinely pin the tracked-value semantics.

**The retrieval test is vacuous with respect to this diff, and worse, it enshrines the bug.** `_RETRIEVAL_CHILD` never references `polyphonic_recall` config or kwargs — it calls `provider.initialize(...)` then sets `os.environ["MNEMOSYNE_POLYPHONIC_RECALL"]` directly. Since `beam.py:5490` reads that env var per call and has done so since before this PR (see `tests/test_beam_e5_polyphonic_recall.py`), the test passes unchanged on unpatched main. It validates E5, not the new key. There's no end-to-end test that `polyphonic_recall: true` in `config.yaml` produces a paraphrase hit through `mnemosyne_recall`, which is the one thing this PR adds.

And the final `assert counts["disabled"] == 0` will start failing the moment anyone fixes the `beam.py:5765` gate — i.e. **the correct fix becomes a test failure.**

Smaller test notes: `counts["keyword"]` is measured with `MNEMOSYNE_POLYPHONIC_RECALL="1"` already set in the child, yet the assertion is commented "it is the FTS5 path" — it doesn't exercise FTS at all. Result *counts* are asserted rather than identities (`recall_count` returns `len(results)`), so an unrelated hit satisfies `>= 1`; CodeRabbit and I both asked for a content/ID assertion plus a distractor and it's still open. Two subprocesses each loading fastembed with `timeout=600` makes this one of the slowest modules in the suite, for coverage that's largely pre-existing E5 behavior.

**CI is not green:** `test (3.13)` fails (`test_repair` FD assertion, `17 == 130`); 3.11/3.12 were cancelled.

Happy to be argued out of this if you think the polyphonic route is independently desirable — but then it should land as "enable polyphonic recall via config" on its own merits, not as the fix for zero-overlap WM retrieval, and #518-style the two shouldn't be coupled.

dplush:
This PR now conflicts with current `main`. Once the configuration-scope decision and the outstanding review points are addressed, please rebase, resolve the conflicts, and push a refreshed head for a new review.

dplush:
Thanks for the work here. I am closing this as superseded rather than asking for a rebase.

The underlying working-memory behavior is now documented in #622, and #639 provides the narrow opt-in `MNEMOSYNE_LEXICAL_GATE_MIN=0` for deployments that want pure vector candidates while preserving the precision-oriented default.

This PR remains conflicted and introduces a process-global `MNEMOSYNE_POLYPHONIC_RECALL` setting managed independently by two provider implementations. That can make active providers or profiles influence each other, and it also extends the legacy provider path we are migrating away from.

If we later want polyphonic recall as a product option, it should start as a fresh standalone-provider-only design with instance-scoped configuration, explicit isolation behavior, and current quality and latency evaluation. Since that changes the recall pipeline, it should be proposed separately rather than revived through this branch.

PR #483, opened 2026-07-17
fix: code audit F-rule cleanup (10 files, 73 lines)

ether-btc:
## Summary

Round 2 of the F-rule audit pass following #460 (merged as `2812e98`). Supersedes the v1 of this PR, which dropped `core/config.py` only after review — the v1 force-pushed version here does the same drop but with a single squashed commit and a clean rebase.

## What's in this branch

Single squashed commit dropping every `core/config.py` change from the v1, plus a few minor cleanups that landed on top of `origin/main` between v1 and now. The remaining 10 files are pure F-rule cleanup (F401 unused imports, F841 dead locals, F541 f-string-without-interpolation, F811 reclassified-to-F401, RUF022 `__all__` defined). Every individual removal was independently verified against current `main` by @AxDSan.

### F841 — dead local variables
- `beam.py`: `content_words_set`, `content_lower` (×2), `query_emb_for_bv`, `dedup_result`
- `episodic_graph.py`: `content_lower`
- `shmr.py`: `timestamp`, `query_blob`
- `cli.py`: `clean_args`

### F401 — unused imports
- `diagnose.py`: `typing.{Any,Dict,List}`, `sys`, `platform`. The `importlib.metadata` line is **kept** under `# noqa: F401` — tests monkeypatch it via `.version`.
- `profiles.py`: `REQUIRES_RESTART`
- `beam.py`: `BinaryVectorStore`, `BYTES_PER_VECTOR` from guarded import block
- `memory_browser.py`: `fastapi.Query`

### F811 reclassified to F401
- `plugins.py`: `from rust_cave_001 import compress` (availability probe). Tests create an alternative `compress` symbol in a different scope; suppression narrowed to `# noqa: F401`.

### F541 — f-string-without-interpolation
- `cli.py`: drop the f-prefix from a `  [entity match]` print in `cmd_recall`.

### RUF022 — `__all__` defined
- `importers/__init__.py`: explicit alphabetical `__all__` with all 20 exported names.

### Live consumer fix
- `importers/mem0.py`: remove the dead `session_id` override branch (6 lines). `Mnemosyne.remember()` has no `session_id` parameter, so the override was a no-op that obscured the import path. The original Mem0 timestamp is preserved as `imported_at_original`.

### CLI ergonomics
- `cli.py`: preserve the dry-run vs applied-fix distinction in `cmd_diagnose`'s auto-fix output (inline "Fixed:" label only on the applied path).

## What is deliberately *not* in this PR

[...]

PR #544, opened 2026-07-26
fix: avoid version regex backtracking

dplush:
## Summary

Fix Pattern A version extraction so capitalized multi-word names require whitespace between words. This removes the nested `\s*` ambiguity that allowed exponential regex backtracking on malformed version text.

## Why this path

The leading name token already consumes contiguous letters greedily, so successful captures stay compatible. Pattern B has no equivalent nested quantifier and is intentionally unchanged.

## Scope

Only the Pattern A regex and its focused regression coverage. No extraction-policy expansion or general regex cleanup.

## Validation

- Focused regression suite: `7 passed`
- `py_compile` passed
- Ruff clean for the new test
- `git diff --check` clean
- Independent Opus review: pass with no blockers

## Regression coverage

- Legacy pathological input times out in an isolated subprocess.
- The fixed pattern completes within the same controlled boundary.
- Public BEAM extraction preserves normal version facts, CamelCase no-separator compatibility, and rejects an incomplete version.

## Summary

- Hardened BEAM “Pattern A” version-string extraction in `mnemosyne/core/beam.py` by switching to a module-level compiled regex (`_VERSION_STRING_RE`) and tightening the name matcher to require whitespace between capitalized multi-word tokens.
- Removes nested `\s*` ambiguity to prevent catastrophic regex backtracking on malformed/pathological inputs, improving robustness of local, offline fact extraction (no new dependencies, no network activity).
- Added focused regression coverage in `tests/test_version_regex_backtracking.py` to ensure:
  - the legacy regex would time out while the new regex safely completes,
  - normal BEAM “public” extraction still produces the expected `memoria_facts` version key/value pairs,
  - incomplete version strings are rejected,
  - CamelCase/compatibility edge cases behave as expected.
- Pattern B and all other Mnemosyne architecture surfaces—tiered memory (working/episodic/BEAM), retrieval, consolidation, veracity, sync, privacy posture, and agent integration surfaces (Hermes/MCP/CLI)—remain unchanged.
- This is the right call for long-term maintainability: safer parsing with clearer intent (compiled, reusable regex) and targeted tests, without eroding local-first guarantees.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -207,6 +207,10 @@ except Exception:
 import os
 import re
 
+_VERSION_STRING_RE = re.compile(
+    r'([A-Z][a-zA-Z]+(?:\s+[A-Z][a-zA-Z]+)*)\s+v?(\d+\.\d+(?:\.\d+)?)'
+)
+
 # On Fly.io and other ephemeral VMs, only ~/.hermes is persisted.
 # Default to the legacy Hermes path so memories survive restarts.
 _DEFAULT_ROOT = Path(
@@ -4748,7 +4752,7 @@ class BeamMemory:
 
         # Version strings — two patterns:
         # Pattern A: "PostgreSQL v14.2", "Docker 27.1.1" (name directly before version)
-        for m in _re.finditer(r'([A-Z][a-zA-Z]+(?:\s*[A-Z][a-zA-Z]+)*)\s+v?(\d+\.\d+(?:\.\d+)?)', content):
+        for m in _VERSION_STRING_RE.finditer(content):
             name = m.group(1).strip()
             ver = m.group(2)
             key = f"{name.lower().replace(' ', '_')}_version"
--- a/tests/test_version_regex_backtracking.py
+++ b/tests/test_version_regex_backtracking.py
@@ -1,0 +1,70 @@
+"""Regression coverage for version-string extraction performance and captures."""
+
+import subprocess
+import sys
+
+import pytest
+
+from mnemosyne.core.beam import BeamMemory
+
+
+_PATHOLOGICAL_INPUT = ("AAAAAA " * 20) + "v1."
+_LEGACY_REPRODUCER = r'''
+import re
+
+pattern = re.compile(
+    r"([A-Z][a-zA-Z]+(?:\s*[A-Z][a-zA-Z]+)*)\s+v?(\d+\.\d+(?:\.\d+)?)"
+)
+pattern.search(("AAAAAA " * 20) + "v1.")
+'''
+_CURRENT_REPRODUCER = (
+    "from mnemosyne.core.beam import _VERSION_STRING_RE\n"
+    f"assert _VERSION_STRING_RE.search({_PATHOLOGICAL_INPUT!r}) is None\n"
+)
+
+
+def test_version_regex_avoids_legacy_backtracking():
+    """The old nested quantifier times out while the replacement returns safely."""
+    with pytest.raises(subprocess.TimeoutExpired):
+        subprocess.run(
+            [sys.executable, "-c", _LEGACY_REPRODUCER],
+            timeout=2,
+        )
+
+    current = subprocess.run(
+        [sys.executable, "-c", _CURRENT_REPRODUCER],
+        capture_output=True,
+        text=True,
+        timeout=2,
+    )
+    assert current.returncode == 0, current.stderr
+
+
+@pytest.mark.parametrize(
+    ("text", "expected_pair"),
+    [
+        ("PostgreSQL v14.2", ("postgresql_version", "14.2")),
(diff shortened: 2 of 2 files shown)


AxDSan (commented):
Correct fix, correct diagnosis, and this one's first in the merge train.

The old pattern at `beam.py:4751` — `([A-Z][a-zA-Z]+(?:\s*[A-Z][a-zA-Z]+)*)\s+v?(\d+\.\d+(?:\.\d+)?)` — is catastrophically ambiguous: because the separator is `\s*`, a single contiguous letter run can be partitioned into the repeated group in exponentially many ways (compositions of a length-n word into parts ≥ 2), so `"AAAAAA " * 20 + "v1."` never terminates. Changing it to `\s+` makes each whitespace-delimited word consumable exactly one way, which kills the exponential.

The path is reachable with user-controlled content: `extract_and_store_facts` (`beam.py:4644`) runs on every `remember()` (`beam.py:3317`, `3394`, `3650`) and on imported content (`importers/hindsight.py:344`), with no length cap.

I verified behavioral equivalence empirically rather than taking the argument on faith: fuzzed **200,000** random strings over letters/whitespace/digits/`v`/`.` comparing old vs new `finditer` `group(1)`/`group(2)`/`span` — **zero differences**. That matches your reasoning (the leading `[a-zA-Z]+` is greedy, so the zero-width-separator branch never adds reachable name strings). Hoisting to `_VERSION_STRING_RE` at `beam.py:210` is not an instance of the #482 import-time-config trap — it's a pattern constant, not a config read.

Three things I'd like folded in, none blocking:

**1. The new pattern is still quadratic, not linear** — and the PR/CodeRabbit summary reads as if the class of problem is gone. Measured `findall` on `('Aaaaaa ' * n) + 'v1.'`:

| n | time |
|---|---|
| 50 | 0.4 ms |
| 100 | 1.5 ms |
| 200 | 6.0 ms |
| 400 | 24.9 ms |
| 800 | 100.5 ms |

An exact 4× per doubling (n start positions × O(n) backtrack each). Extrapolating, ~56 KB of Title-Case prose is ~10 s inside `extract_and_store_facts`, and there's no content length cap at `beam.py:4644`. Exponential → quadratic is a big win; let's just describe it accurately and open a follow-up for a length cap or an anchored variant.

**2. A verbatim copy of the vulnerable regex survives** at `_benchmarks/evaluate_beam_end_to_end.py:368`. Benchmark-only so not production, but it's the one missed call site and it will hang the benchmark on the same input.

**3. The parametrized test is vacuous with respect to the fix.** I ran all six cases of `test_version_regex_preserves_public_extraction` against the unpatched tree (isolated HOME/TMPDIR) and all six pass — `postgresql_version`, `docker_version`, `ubuntu_linux_version`, `product_version`, `postgresqlserver_version`, plus the empty-set case for `'PostgreSQL v1.'`. Useful no-regression coverage, but the body's framing implies it validates the fix; only `test_version_regex_avoids_legacy_backtracking` gates the change. That one does work — it proves the inlined legacy pattern hangs (>2 s, killed) while `_VERSION_STRING_RE` returns None promptly, so re-relaxing `\s+` to `\s*` would fail CI.

Test-mechanics nits:

- `test_version_regex_avoids_legacy_backtracking` asserts that CPython's `re` engine *is* slow. It burns a guaranteed 2 s of wall clock every run and would start failing (`pytest.raises(TimeoutExpired)` unmet) on any engine that adds backtracking memoization, or on a non-CPython runner. Cheap mitigation: mark it slow, or assert on a shape property of the pattern instead.
- The 2 s subprocess budget for `_CURRENT_REPRODUCER` has to cover a full cold `import mnemosyne.core.beam` (numpy + sqlite_vec + config seeding). I measured 1.19 s cold / 0.73-0.77 s warm locally. CI is green on all four legs and pytest runs serially (`ci.yml:109`), so this isn't a real failure — but the margin is thin on a loaded runner, and a miss surfaces as an uncaught `subprocess.TimeoutExpired` from line 40 rather than a readable assertion. Raise the budget or pre-import in the child.
- `_LEGACY_REPRODUCER`/`_CURRENT_REPRODUCER` run via `sys.executable -c`, which puts cwd on `sys.path[0]`; the import only resolves because the package is installed or pytest was invoked from the repo root. Passing `cwd=` explicitly would make that intent explicit. Also the legacy `subprocess.run` at line 37 omits `capture_output=True` — harmless, but inconsistent with the second call.
- No test drives the pathological input through the public `extract_and_store_facts` path under a time bound; the performance assertion is against the module constant in a subprocess rather than the production call site at `beam.py:4752`. Given the constant is now the only consumer that's acceptable coupling, but a future refactor that inlines a fresh regex at the call site would be invisible to these tests.

CI green on 3.10-3.13. Merging this first — it's the most isolated `beam.py` hunk in the queue (`:210` + `:4752`), clear of #483's hunks and of everything else touching that file.

Issue #506, opened 2026-07-21
[BUG] Sleep consolidation output outranks its source memories by default (tier-1 episodic summaries + confidence-as-importance proposals)


israellot:
### Summary

Sleep consolidation output systematically outranks the source memories it summarizes, degrading recall quality. A single large sleep pass on our production store (~9.8k memories) dropped a fixed 8-query recall eval from **8/8 to 2/8 top-5 hits**. Demoting the sleep-generated rows using mnemosyne's own mechanisms (episodic tier + importance) restored 8/8 immediately, with no other changes.

This is the recall-side counterpart of #426 (consolidated *originals* crowding `get_context()`, fixed in #427): here it's the *derived* rows crowding `recall()`.

### Environment

- mnemosyne-memory 3.14.0, mnemosyne-hermes 0.5.0 (Hermes provider)
- bge-base-en-v1.5 local embeddings, sqlite-vec
- Store: ~9,790 working_memory rows (4.5k wiki chunks at importance 0.67–0.75, 2.3k session digests ~0.7), ~1.2k episodic rows

### What happened

A sleep/consolidation pass over the backfilled corpus wrote, in one burst:

- **~1,024 `sleep_consolidation` rows into `episodic_memory`** — all at **tier 1** (full ranking weight)
- **~2,048 `sleep_model_refresh_proposal` rows into `working_memory`** — at **avg importance 0.914**

These rows are LLM paraphrases of existing content (wiki pages, digests). Because they're paraphrases, their embeddings sit right next to their sources for exactly the queries the sources should win — and their score boosts push them past the originals. Result: for 6 of our 8 eval queries, all five top-5 slots were sleep-generated derivatives of the expected page, with the expected page itself pushed out.

A second effect: prompt injection. The injection gate (drop if `score<0.20 AND importance<0.65`) can never drop a 0.91-importance row, so injected context blocks filled with vague consolidation summaries instead of sharp source content.

### Root causes (both present in v3.14.0)

1. **`consolidate_to_episodic()` never sets `tier`** — the `INSERT INTO episodic_memory` (beam.py ~4349) omits the tier column, so every consolidation summary enters at tier 1 (1.0× multiplier) and keeps full weight until age-based degradation (30/180 days). Freshly derived summaries therefore outrank everything for a month minimum.

[...]

PR #546, opened 2026-07-26
fix(beam): sleep summaries inherit the batch's peak source importance

Milgauss:
sleep() writes every episodic summary at a flat importance=0.6, discarding the signal callers set on the source rows. Measured on a live bank (2,795 working memories): recall's top slots are dominated by 0.6 summaries winning on relevance while the tagged 0.9 originals they summarize rank below them, inverting the intent of importance-weighted recall. A flat 0.6 also sits just under the hermes integration's prefetch min_importance rescue floor (0.65), so summaries survive per-turn filtering on topical score alone.

Fix: `importance = min(max(0.6, peak of source importances), 0.85)`.

- a batch containing a 0.9 row yields a 0.85 summary (the cap keeps a summary from outranking the originals it summarizes on importance alone)
- an all-default batch (0.5 rows) keeps the historical 0.6 exactly, so existing banks see no change unless callers actually tag importance
- a mid-range peak (e.g. 0.75) passes through unchanged

Tests: three regressions in `tests/test_sleep_summary_importance.py`; `test_beam_e3_additive_sleep.py`, `test_sleep_lock_hygiene.py`, `test_get_context_consolidated.py`, and the consolidate-fact suites pass unchanged (64 tests total).

## Summary

- Updates BEAM `sleep()` consolidation so episodic summary rows created from `working_memory` inherit the **peak** `working_memory.importance` value for that consolidated source batch, clamped to **`0.6–0.85`** (replacing the previous fixed `importance=0.6`).
- Preserves the historical **`0.6`** floor for all-default/low-importance batches, passes through mid-range peaks (e.g., `0.75`) unchanged, and caps high peaks (e.g., `0.9 → 0.85`) to reduce the chance that derived summaries outrank their strongest source rows purely due to importance scaling.
- Improves consolidation robustness: after the claim/marker commit, `sleep()` computes `peak_importance` by parsing each source row’s `importance` as a float and **skipping** malformed or non-finite values (`+inf`, `NULL`, etc.), preventing the consolidation phase from failing in a way that could strand/leave inconsistent “claimed” working rows.
- Tiered-memory impact is intentional and localized to the **working → episodic** pipeline: the change only alters how `importance` is written onto new `episodic_memory` rows during `sleep()` (and uses the existing `consolidate_to_episodic()` insert path). Working-memory storage, session-scoping, and consolidation semantics (original working rows remain recallable and are just marked `consolidated_at`) are unchanged. Recall/other retrieval strategies, veracity, and degradation are not modified.
- Local-first/privacy posture remains intact: the update is purely an on-device SQLite/BEAM consolidation behavior change inside the existing `sleep()` flow (which still uses the configured local LLM/compression path when available), with no changes to sync transport, remote calls, or agent integration surfaces (Hermes/MCP/CLI).
- Adds regression tests (`tests/test_sleep_summary_importance.py`) covering floor, pass-through, cap, and malformed/non-finite importance handling, and verifies that consolidation does not leave orphaned/uncompleted batches. Reports **64 related tests passing**.

**Architectural note:** this does change importance-based recall ranking by letting episodic summaries carry bounded source-signal strength; while the `0.6–0.85` clamp is a safety guard, it may still affect source-vs-summary relative ranking behavior, which maintainers flagged as a follow-up-worthy concern.


AxDSan (commented):
The code does exactly what the body says, and cleanly: the post-patch block (~`beam.py:8459-8481`) takes the max of the claimed batch's importances, skips non-numeric/non-finite values, and passes `min(max(0.6, peak), 0.85)` to `consolidate_to_episodic`. `math` is already imported (`beam.py:24`), `importance` is in the sleep SELECT (`beam.py:8265`), items are the post-claim filtered rows so a lost race can't inflate the peak, and it sits inside `if not dry_run:` so dry-run stays side-effect-free. I confirmed the behavior change is real — seeded four batches against unpatched main in a temp DB and all four produce importance 0.6 today.

**My problem is with the premise, and it's why I'm holding this rather than requesting changes.**

The PR says the bug is "0.6 summaries dominate recall's top slots while the tagged 0.9 originals rank below them". But raising the summary's importance to 0.85 makes summaries win *harder*, not less. I probed main directly — 3 aged 200h working rows at importance 0.9 plus a 0.6 summary of them — and scored the summary at **0.6355** against **0.3090** for each original. The originals survived only via the `exact_source_hit` escape in the cross-tier dedup. The inversion is driven by the aged-working recency discount at `beam.py:5775`, not by the summary's importance floor. This change pushes the same direction as the problem.

It also collides head-on with **#506**, which reports the opposite direction with a reproducible 8/8 → 2/8 eval and argues consolidation output should rank *below* its sources — the root cause there being that `consolidate_to_episodic`'s INSERT omits `tier`, so every summary enters at `TIER1_WEIGHT=1.0` (`beam.py:4386-4392`). Both PRs turn the same knob opposite ways; they can't both land.

I'm inclined to go with #506's direction, since it has production measurements behind it. Two ways this becomes landable:

1. Re-target it at the actual ranking mechanism (the `tier` omission and/or the aged-working discount at `beam.py:5775`), which may well make the importance floor moot; or
2. Keep it as-is but behind a default-off kill switch (`MNEMOSYNE_SUMMARY_IMPORTANCE_INHERIT`) so it ships as an opt-in experiment. Right now 0.6 and 0.85 are hardcoded inline with no override, so an operator hit by the ranking change can't revert short of downgrading. (To your credit it avoids the #482 import-time-constant trap — please keep it read per-call if you do extract it.)

Would like your read on the #506 measurements before either of us writes more code here.

### Tests

Four tests, all driving the real `sleep()` path end to end against a temp DB — good, they exercise the patched call site rather than a helper. I extracted the file and ran the equivalent seeding against unpatched main's `beam.py` in a scratch dir: `test_summary_inherits_peak_importance_capped`, `test_mid_importance_peak_passes_through`, and the `approx(0.75)` assertion in `test_malformed_importance_does_not_orphan_the_batch` genuinely fail without the patch.

Two that don't: `test_all_default_batch_keeps_flat_floor` passes unpatched (produces exactly 0.6), and the `unclaimed == 0` assertion also passes unpatched. Both are legitimate no-change guards — just don't count them among "three regressions" as the docstring does.

The coverage gap that matters: **nothing tests the recall-side consequence.** No test asserts the effect on `_cross_tier_dedup` (`beam.py:7107`) or on the Hermes prefetch gate (`hermes_memory_provider/__init__.py:2230`) — which is exactly where the disagreement with #506 lives. Also untested: multiple source groups computing peaks independently, and the exact 0.85 cap boundary.

Two notes for whoever lands this:

- CodeRabbit's NaN-coverage request **can't be satisfied through the DB path**: SQLite stores `float('nan')` as NULL (`typeof` returns `'null'`), while `float('inf')` does round-trip as REAL inf. So `math.isfinite` only ever fires for ±Inf on real data — worth a code comment so nobody "fixes" the guard later.
- The new tests rely on the autouse conftest fixture (`tests/conftest.py:119-153`) to suppress the LLM, but that only patches `local_llm._load_llm`; `llm_available()` short-circuits True when module-level `LLM_BASE_URL` was populated at import (`local_llm.py:435-436` — #482 again). Existing sleep tests belt-and-suspenders monkeypatch `local_llm.llm_available` to False (`tests/test_beam_e3_additive_sleep.py:171-172`); please follow that pattern.
- No CHANGELOG entry or version bump, and `CONTRIBUTING.md:32` requires both — this silently changes recall ranking for every existing bank that tags importance.

AxDSan:
@Milgauss — decision, and thank you for waiting on it.

**I'm going with #506's direction, which means this PR doesn't land as written.** Here's the reasoning, because I don't want it to look arbitrary.

Your code is correct and your tests are real — I verified three of the four fail against unpatched `main`. The disagreement is about the premise, not the implementation.

This PR's stated bug is "0.6 summaries dominate recall's top slots while the tagged 0.9 originals rank below them". But raising the summary's importance to 0.85 makes summaries win *harder*. I probed `main` directly — 3 aged 200h working rows at importance 0.9 plus a 0.6 summary of them — and scored the summary at **0.6355** against **0.3090** for each original. The originals survived only via the `exact_source_hit` escape in the cross-tier dedup. The inversion is driven by the aged-working recency discount at `beam.py:5775`, not by the summary's importance floor. So the floor isn't what's producing the behavior you're seeing, and lifting it pushes the same direction as the problem.

Meanwhile #506 reports the opposite failure with a reproducible 8/8 → 2/8 eval, and its root cause is concrete: `consolidate_to_episodic`'s INSERT omits `tier` (`beam.py:4386-4392`), so every summary enters at `TIER1_WEIGHT=1.0`. That's a mechanism I can point at in the code, not an inference from ranking behavior.

Two ways your work still lands, and I'd like it to:

1. **Re-target at the mechanism.** If the `tier` omission and/or the `beam.py:5775` discount get fixed, re-measure. The importance floor may turn out to be moot, or it may turn out you were right about the floor and wrong about the cause — either way we'd know from measurement rather than argument.
2. **Keep it as an opt-in.** Land it behind a default-off `MNEMOSYNE_SUMMARY_IMPORTANCE_INHERIT` kill switch. Right now `0.6` and `0.85` are hardcoded inline, so an operator hit by the ranking change can't revert without downgrading. As a flag-gated experiment I'd merge it this week.

If you want to go further, the most valuable thing anyone could add here is a **recall-side test**: nothing currently asserts the effect on `_cross_tier_dedup` (`beam.py:7107`) or the Hermes prefetch gate (`hermes_memory_provider/__init__.py:2230`), which is precisely where you and #506 disagree. A test that pins the ranking outcome would settle this class of question permanently instead of per-PR.

Two smaller notes for whichever path you take:

- CodeRabbit's NaN-coverage request **can't be satisfied through the DB path**. SQLite stores `float('nan')` as NULL (`typeof` returns `'null'`), while `float('inf')` does round-trip as REAL inf. So `math.isfinite` only ever fires for ±Inf on real data — worth a comment so nobody "fixes" the guard later.
- Your tests rely on the autouse conftest fixture to suppress the LLM, but that only patches `local_llm._load_llm`; `llm_available()` short-circuits True when module-level `LLM_BASE_URL` was populated at import (`local_llm.py:435-436` — the #482 class). Existing sleep tests belt-and-suspenders monkeypatch `local_llm.llm_available` to False (`tests/test_beam_e3_additive_sleep.py:171-172`); follow that or this will be flaky on machines with a local LLM configured.

Leaving this open rather than closing it — the work is good and I'd rather it get re-pointed than binned. Tell me which path you'd prefer and I'll support it.

Milgauss:
Thanks for the depth of the verification here — you're right about the premise, and I'll concede it directly: my PR body described the summaries-outrank-originals inversion, but this patch pushes that direction, it doesn't fix it. The goal underneath was narrower than the body implied: a flat 0.6 flattens the derived rows *among themselves*, so an important summary can't be distinguished from a low-signal one. I framed that as the inversion bug, which is a different problem — and per your probe (0.309 aged original vs 0.636 flat summary), one this change makes worse.

On #506: I buy it, and my own measurements corroborate it. On a production bank (2,795 working-memory rows, hermes provider, social-chat prefetch profile), a fixed 20-query eval shows top-5 slots dominated by sleep summaries and default-importance autosave rows, with deliberately tagged 0.8–0.95 originals ranked below them — independently the same shape as israellot's 8/8 → 2/8. Derived rows should rank below their sources, and the tier omission in `consolidate_to_episodic()` looks like the right knob.

Rather than a kill switch, I'd propose re-targeting this PR so the two knobs compose instead of colliding:

1. `consolidate_to_episodic()` sets an explicit demoted tier on `sleep_consolidation` rows (root cause 1 of #506).
2. Importance inheritance stays, but *under* that demotion: with a tier multiplier below 1.0, the inherited importance orders summaries among themselves (important vs low-signal — the original goal) without pushing any summary past its sources.
3. Recall-side tests where the disagreement actually lives: a source outranks its own summary for a query both match; a high-peak summary outranks a default-batch summary; plus the `_cross_tier_dedup` and prefetch-gate cases you flagged, the multi-source-group peaks, and the exact 0.85 cap boundary.

If you'd rather keep this PR minimal, I can split (1) into its own PR referencing #506 and rebase this one on top of it — happy either way.

Housekeeping I'll fold into the next push regardless of direction: CHANGELOG entry + version bump per CONTRIBUTING.md; fix the test docstring's "three regressions" overcount (two of the four assertions are no-change guards, as you note); a comment on the `math.isfinite` guard documenting that SQLite stores NaN as NULL so the guard only fires for ±Inf through the DB path (so nobody "fixes" it later); and the `llm_available` monkeypatch following the `test_beam_e3_additive_sleep.py` pattern rather than relying on the autouse fixture alone.

Agreed the aged-working recency discount is the thornier, separate discussion — even with the tier fix, a 200h-old 0.9 original at 0.309 loses to a lot it shouldn't. Happy to open that as its own issue with measurements if useful.

AxDSan:
Closing this in favour of #563, which targets the same behaviour from the opposite direction and carries the measurement.

The docstring here states the problem as:

> the 0.6 summaries then dominate recall's top slots on relevance while the high-importance originals they summarize rank below them, inverting the intent of importance-weighted recall

I agree with that diagnosis. But the change raises summary importance to the batch peak (capped 0.85), which pushes derived rows further up the ranking, not down. The fix moves in the opposite direction from the problem it describes. If the concern is that summaries outrank their sources, the lever has to reduce the summary's ranking pressure.

#563 does that, and it has numbers behind it: on a ~9.8k-memory production bank a single sleep pass wrote ~1,024 tier-1 `sleep_consolidation` rows and dropped a fixed 8-query recall eval from 8/8 to 2/8 top-5 hits; demoting those rows to tier 3 restored 8/8.

The two cannot both land. A summary entering at tier 3 with importance lifted to 0.85 just relitigates the same question in two places.

What I would keep from this PR: the defensive handling of non-numeric and non-finite `importance` values from legacy banks, and the point that it runs after the claim commit so a raise there would orphan claimed rows. That reasoning is sound and independent of the ranking direction. If you want to send it as a standalone hardening PR against the current `sleep()` path I will take it.

Not a reflection on the work, just two PRs pulling the same knob opposite ways.

PR #557, opened 2026-07-27
fix(recall): treat blank query_time as unset (#555)

### LINKED ISSUE #555: recall raises on the tool schema's own default: query_time="" falls into the ISO parser
## Bug

`tool_schemas.py` declares the `mnemosyne_recall` MCP tool's `query_time` as:

```python
"query_time": {
    "type": "string",
    "description": "ISO timestamp to treat as 'now' for temporal scoring. Default is current time.",
    "default": "",
},
```

so an MCP harness that omits the parameter sends `""`. But `_parse_query_time` in `core/beam.py` only maps `None` to "now"; an empty string falls into the `str` branch and raises:

```
Invalid query_time format: ''. Expected ISO datetime string.
```

Net effect: `mnemosyne_recall` hard-fails from any client that passes the schema's own default (observed with Claude Code / eve-style agent harnesses on `mnemosyne_memory` 3.12.2). Passing an explicit ISO timestamp works.

## Fix (one line)

Honour the docstring's "None -> current UTC time" contract for empty/whitespace strings too:

```diff
-    if query_time is None:
+    if query_time is None or (isinstance(query_time, str) and not query_time.strip()):
         return datetime.now(timezone.utc)
```

Alternative/companion: change the schema default from `""` to omit/`null`.

Happy to open a PR if useful.

AxDSan:
Fixes #555.

## The bug

`tool_schemas.py` declared the `mnemosyne_recall` tool's `query_time` with `"default": ""`, but `_parse_query_time` (`beam.py:1454`) mapped only `None` to "now". A blank string fell through to the `str` branch and raised:

```
Invalid query_time format: ''. Expected ISO datetime string.
```

So **any MCP harness that sends declared defaults could not call `mnemosyne_recall` at all** — the tool hard-failed on its own advertised default, while passing an explicit ISO timestamp worked. @dalkommatt observed this with Claude Code / eve-style harnesses.

## Why it was a one-off

Every sibling blank-defaulted parameter already guards for this — `valid_until` uses `arguments.get(...) or None` (`mcp_tools.py:270`), and `as_of` is `or`-guarded inside `TripleStore.query` (`triples.py:205`). `query_time` at `mcp_tools.py:335` was the single place that missed the idiom. I audited the other `"default": ""` parameters (`valid_from`, `valid_until`, `as_of`, `subject`/`predicate`/`object`, `replacement_id`, `validator`, `new_content`, `note`) and found no second instance.

## Fix

Three layers, so this is closed for SDK callers too, not just MCP:

1. **`_parse_query_time`** treats blank/whitespace-only strings as unset, honoring the docstring's `None -> current UTC time` contract.
2. **The MCP handler** normalizes `""` to `None`, matching the existing sibling idiom.
3. **The schema** no longer advertises a default that means "omitted".

## Tests

Three added to `TestParseQueryTime`, all verified to **fail on unpatched source and pass patched**:

| test | on unpatched `main` |
|---|---|
| `test_blank_string_returns_now` | FAILED |
| `test_whitespace_string_returns_now` | SUBFAILED ×3 (`'   '`, `'\t'`, `'\n'`) |
| `test_recall_tool_schema_does_not_declare_blank_default` | FAILED |

`tests/test_temporal_recall.py`: 27 passed (24 pre-existing + 3 new). Adjacent suites show an **identical failure set before and after** the change, so no regressions. (`tests/test_mcp_server.py` has 6 failures in my local env from missing optional deps; they are present on pristine `main` too and CI is green.)

## Also in this PR

The CHANGELOG entries owed for #544 and #554, which both landed without one despite `CONTRIBUTING.md:32`, plus the `3.15.1` bump. Flagging that I merged those two in that state — my inconsistency, swept up here rather than bounced back to the contributors.

Reported and correctly diagnosed by @dalkommatt, who also proposed essentially this fix.

## Summary

- Treats `mnemosyne_recall` `query_time` values that are `""` or whitespace-only as “unset,” so recall uses the current UTC time instead of attempting ISO parsing.
- Normalizes blank values at the MCP/tool boundary (converting blank/whitespace strings to `None`), removes the schema’s blank-string default, and preserves explicit ISO timestamps; non-string falsey inputs remain rejected.
- Adds regression coverage for parser behavior, schema defaults, MCP normalization, and explicit timestamps.
- Improves local query-cache consistency by clearing enhanced-recall/query cache only after successful `BeamMemory.invalidate()` updates (avoids touching cache for missing/unauthorized IDs).
- Hardens long-running regex work in fact extraction: version-string extraction regex was tightened to avoid catastrophic backtracking risk.

## Architectural impact

- **Core memory architecture (working/episodic/BEAM):** No changes to tiering, consolidation pipeline, veracity/trust scoring, or the underlying storage model. The change only affects *how recall interprets a time parameter* (temporal boost input) and *cache invalidation correctness*.
- **Retrieval strategy:** Retrieval logic remains the same; only the temporal query-time interpretation is adjusted so the temporal scoring path receives a valid UTC datetime (now) when the caller “omits” `query_time` via blank defaults.
- **Local-first / privacy posture:** Preserved. Updates are confined to local input normalization, local database state changes, and local cache invalidation—no added network calls, telemetry, or sharing behavior.
- **Agent integration surfaces (Hermes/MCP/CLI):** Improves MCP harness compatibility with tool schema defaults by preventing blank `query_time` from propagating into ISO parsing. Schema documentation now reflects omission semantics rather than advertising a blank default; CLI and core recall API semantics are otherwise unchanged.
- **Maintainability:** Clearer “omit vs explicit timestamp” contract via schema + boundary normalization, with defense-in-depth in the core parser. The cache invalidation and regex hardening reduce hard-to-debug state skew and performance hazards, respectively.

## Long-term risk assessment

- **Architectural win:** The right-call layering—tool/MCP normalization plus core parser behavior—prevents brittle dependence on upstream schema defaults and makes temporal recall more robust.
- **Local-first erosion check:** No erosion. The only functional behavior change is deterministic/local (timestamp parsing + cache invalidation + safer regex), not a new integration surface or data-exchange pathway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
+## [Unreleased]
+
+### Fixed
+
+- **`mnemosyne_recall` crashed on its own schema default (#555).** The tool schema declared `query_time` with `"default": ""`, but `_parse_query_time` mapped only `None` to "now" — a blank string fell through to the ISO parser and raised `Invalid query_time format: ''`. Any MCP harness that sends declared defaults could not call `mnemosyne_recall` at all. Blank and whitespace-only values are now treated as unset, the MCP handler normalizes `""` to `None` (matching the `or None` idiom already used for `valid_until` and `as_of`), and the schema no longer advertises a default that means "omitted". Thanks @dalkommatt for the report and the diagnosis.
+- **Enhanced Recall served invalidated rows until TTL expiry (#550, #554).** `BeamMemory.invalidate()` now clears the query cache after a successful update, including the persisted `query_cache.db` when the instance has no in-memory cache of its own. Missing or unauthorized IDs leave the cache untouched. Remaining gaps are tracked in #552 (live peer coherence), #553 (`forget_working`) and #556 (`remember`/update paths).
+- **Catastrophic regex backtracking in version-string extraction (#544).** The pattern used by `extract_and_store_facts` could be driven into exponential backtracking by Title-Case input, hanging every `remember()` and import on attacker- or user-supplied content. The separator is now `\s+`, which makes each whitespace-delimited word consumable exactly one way. Behavioral equivalence was verified across a 200,000-string fuzz with zero differences.
+
 ## [3.15.0] - 2026-07-20
 
 ### Fixed
--- a/mnemosyne/__init__.py
+++ b/mnemosyne/__init__.py
@@ -10,7 +10,7 @@ Example:
     >>> results = recall("user preferences")
 """
 
-__version__ = "3.15.0"
+__version__ = "3.15.1"
 __author__ = "Abdias J"
 __license__ = "MIT"
 
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1463,9 +1463,10 @@ def _parse_query_time(query_time: Optional[Union[str, datetime]]) -> datetime:
     - datetime -> normalized to UTC
     Naive values are treated as UTC for backward compatibility.
 
-    A blank string is treated as "unset" because the ``mnemosyne_recall`` tool
-    schema declares ``query_time`` with ``"default": ""``, so MCP harnesses that
-    send declared defaults deliver ``""`` rather than omitting the field.
+    A blank string is treated as "unset" for compatibility with callers that
+    send ``""`` to mean "omitted" — notably MCP harnesses built against the
+    older ``mnemosyne_recall`` schema, which declared ``query_time`` with
+    ``"default": ""`` (see #555). Non-string falsey values are still rejected.
     """
     if query_time is None:
         return datetime.now(timezone.utc)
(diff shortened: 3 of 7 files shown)


AxDSan:
Addressed all three CodeRabbit findings in b7e24c6 — all were valid.

**1. Outdated docstring.** It claimed the schema declares `"default": ""`, which this PR removes. Reworded to describe the compatibility contract for callers built against the *older* schema.

**2. `or None` was wrong in both directions.** This is the one worth reading twice — CodeRabbit flagged the falsey-non-string half, and testing it surfaced a second bug in my own patch:

- It **missed whitespace-only strings**. `"\t" or None` → `"\t"`, because whitespace is truthy. So `"   "` and `"\t"` still reached the ISO parser and still raised — the exact bug this PR exists to fix, only partially closed.
- It **swallowed falsey non-strings**. `0`, `False`, `[]` became `None` and were silently treated as "unset", suppressing `_parse_query_time`'s `TypeError`.

Replaced with an explicit `isinstance(query_time, str) and not query_time.strip()` check, which fixes both.

**3. No MCP-level coverage.** Added three tests at the handler boundary: blank/whitespace → `None`, explicit ISO forwarded untouched, and falsey non-strings *not* converted.

Confirmed the new blank-normalization test **fails against the `or None` form** (`test_handle_recall_normalizes_blank_query_time FAILED`), so it pins the whitespace case rather than just restating the fix.

Local: 27 passed in `test_temporal_recall.py`, 3/3 new MCP tests pass, and `test_mcp_server.py` shows a byte-identical failure set to pristine `main` (6 pre-existing failures from optional deps missing in my env; CI is green).

PR #565, opened 2026-07-28
fix: add register() bridge so Hermes memory provider discovery finds MnemosyneMemoryProvider

oscarello:
## Problem

The Hermes memory provider loader (plugins/memory/__init__.py) discovers a provider directory by text-scanning its top-level __init__.py for `register_memory_provider` or `MemoryProvider`, then drives registration through one of two entry points:

1. mod.register(collector) — the collector's register_memory_provider(provider) captures the active provider
2. a top-level MemoryProvider subclass, instantiated directly

The root __init__.py only exposed register_memory_provider (a discovery marker), so the loader found the directory but never invoked it: load_memory_provider("mnemosyne") returned None and the agent silently fell back to built-in memory while memory.provider: mnemosyne was configured.

## Fix

- Define register(ctx) as the loader bridge that **dispatches on ctx shape** (review round 2, dplush):
  - a memory-provider collector — the only ctx exposing register_memory_provider, which is exactly what plugins.memory passes — registers exactly one MnemosyneMemoryProvider via the bridge;
  - any other ctx preserves the legacy hermes_plugin.register registration path (no shadowing).
  Previously, when hermes_plugin was importable, register WAS hermes_plugin.register, so the memory loader's mod.register(collector) call bypassed the bridge and the collector never received MnemosyneMemoryProvider.
- Promote the repo root on sys.path unconditionally: an ambient path entry (e.g. $HERMES_HOME/plugins/ in a Hermes process) could shadow the checkout, making import mnemosyne resolve to this stub instead of the core package — hermes_memory_provider's from mnemosyne.core... imports then failed with "No module named 'mnemosyne.core'" and the provider silently failed to load.
- Keep register_memory_provider as the discovery marker + stable module-level API.

## Verification

- tests/test_loader_registers_provider.py:
  - collector-contract test: root-repo symlink layout, mod.register(collector) registers exactly one MnemosyneMemoryProvider (runs in repo CI)
  - combined loader regression (new, round 2): the real plugins.memory loader with hermes_plugin importable — load_memory_provider("mnemosyne") returns the provider and the legacy hermes_plugin.register is never invoked (skips where Hermes is not importable)
  - real-loader integration test: actual plugins.memory loader against a $HERMES_HOME temp layout — load_memory_provider("mnemosyne") returns the provider
- tests/test_init_register_shadowing.py: register preserves the legacy path for a non-collector ctx; a memory-provider collector routes to the bridge and registers exactly one provider; register_memory_provider exists separately
- Rebased onto current main; CI green across the full matrix (lint, docs-check, build, test 3.10–3.13)
- Local verification: real-loader tests 3/3 pass, ruff 0.15.22 clean, full-suite failure list identical to the main baseline


AxDSan (changes requested):
Thanks for tracing this. The discovery gap is real and worth fixing, but this patch does not do what the description says it does.

The root ` __init__.py` already binds `register` at module scope:

```python
try:
    from hermes_plugin import register
    __all__ = ["register", "__version__", "__author__"]
except ImportError:
    __all__ = ["__version__", "__author__"]
```

The new `def register(ctx)` is appended after that block, so it rebinds the name. When `hermes_plugin` is importable, the plugin-registration callable is replaced by the memory-provider bridge, and `__all__` still advertises `register` while pointing at the wrong function. The PR body says it "keeps the existing Hermes plugin registration path intact"; as written it is the one thing it breaks. The verification snippet only exercises `load_memory_provider("mnemosyne")`, which is why the regression does not show up there.

Two ways out:

1. Expose `register_memory_provider` (the name `plugins/memory/__init__.py` already probes for) instead of `register`, so the two discovery paths use distinct symbols and neither shadows the other.
2. Keep a single `register(ctx)` that dispatches to both, calling the `hermes_plugin` callable captured before rebinding and then the memory-provider bridge.

I prefer 1. It matches what the discovery code looks for and avoids one entry point serving two unrelated protocols.

Whichever you pick, please add a test that asserts the `hermes_plugin` registration path still resolves when `hermes_plugin` is importable. That is the case this PR silently regresses.

oscarello (commented):
Good catch — you're right, `def register(ctx)` was rebinding the name and would silently break the plugin path when `hermes_plugin` is importable.

Went with option 1: renamed to `register_memory_provider` — the symbol `plugins/memory/__init__.py` already probes for. Both paths now use distinct symbols, no shadowing.

Added a regression test (`test_init_register_shadowing.py`) that mocks `hermes_plugin` as importable and asserts:
- `mnemosyne.register` resolves to the `hermes_plugin` callable
- `mnemosyne.register_memory_provider` exists as a separate callable
- `register` is still in `__all__`

Ready for re-review.

oscarello:
@AxDSan gentle ping for re-review when you have a moment. Went with option 1 — renamed to `register_memory_provider` (the symbol `plugins/memory/__init__.py` already probes for), so both paths use distinct symbols with no shadowing. Added a regression test (`test_init_register_shadowing.py`) that mocks `hermes_plugin` as importable and asserts `register` still resolves to the plugin callable. Also rebased onto main and addressed the CodeRabbit exception-handling nit (narrowed to `ModuleNotFoundError` with `exc.name` check). Thanks!

dplush:
Thanks for addressing the `register` shadowing issue. I rechecked the current head against the current Hermes provider loader, and I still cannot approve this yet.

`register_memory_provider` is used by Hermes only as a discovery marker when scanning a plugin directory. After that, the loader invokes `mod.register(collector)` or looks for an exported `MemoryProvider` subclass. The new root-level `register_memory_provider(ctx)` is therefore never called, so the claimed root-plugin discovery path still does not register a provider.

The new regression test proves that `hermes_plugin.register` is not shadowed, which is useful, but it does not exercise `load_memory_provider("mnemosyne")` or invoke the bridge. I also noticed that the earlier bridge-delegation coverage was removed rather than replaced by a real loader-path test.

Could you please wire the intended Hermes loader entry point and add an integration test using the actual loader plus a collector? The test should prove that the intended root-repo symlink layout registers exactly one `MnemosyneMemoryProvider`, while preserving the existing `hermes_plugin.register` behavior.

The narrow `ModuleNotFoundError` handling looks correct. Once the real discovery path is covered, please rebase onto current `main`, rerun CI, and request a fresh review.

oscarello:
Thanks for the detailed review — you were right that the loader never calls `register_memory_provider`. Verified: on the old head `load_memory_provider("mnemosyne")` found the directory but returned `None`.

What changed (pushed as `058957a`):

1. **Loader entry point wired.** `plugins/memory/__init__.py` drives registration via `mod.register(collector)` or a top-level `MemoryProvider` subclass — it never invokes `register_memory_provider`. The root `__init__.py` now defines `register(ctx)` as the bridge, delegating to `hermes_memory_provider.register_memory_provider()` (which calls `ctx.register_memory_provider(MnemosyneMemoryProvider())`, i.e. exactly one provider). It is defined only in the `hermes_plugin`-import-error branch, so it can never shadow `hermes_plugin.register` — the shadowing regression test passes unchanged.

2. **Integration tests added** (`tests/test_loader_registers_provider.py`):
   - Collector-contract test in the root-repo symlink layout: `mod.register(collector)` registers exactly one `MnemosyneMemoryProvider` — runs in this repo's CI (no Hermes needed).
   - Real-loader test using the actual `plugins.memory` loader against a temp `$HERMES_HOME/plugins/mnemosyne → repo-root` layout: `load_memory_provider("mnemosyne")` returns the provider. Skips where Hermes isn't importable.

3. **Bonus root cause found while testing**: with the checkout under `$HERMES_HOME/plugins/`, an ambient `sys.path` entry for that parent dir can make `import mnemosyne` resolve to the plugin stub instead of the core package, so `hermes_memory_provider`'s `from mnemosyne.core...` imports die with `No module named 'mnemosyne.core'` and the provider silently fails to load. The stub now promotes the real repo root unconditionally.

4. **Rebased onto current main**; CI is green across the full matrix (lint, docs-check, build, test 3.10–3.13).

The narrow `ModuleNotFoundError` / `exc.name` handling is unchanged per your note. Ready for another look.

dplush:
Thanks for the careful loader-path follow-up. The fallback path without `hermes_plugin` now works, and I verified the real Hermes loader against the root-repo symlink layout.

One compatibility case still blocks approval: when `hermes_plugin` is importable, the root module exports its `register` directly. The memory loader always calls `mod.register(collector)`, so in that case the provider bridge is bypassed and the collector never receives `MnemosyneMemoryProvider`.

The current shadowing regression confirms the legacy callable is preserved, but it does not exercise the real memory loader with `hermes_plugin` available. Could you make `register(ctx)` dispatch to the provider bridge for a memory-provider collector while preserving the legacy plugin-registration path, and add that combined loader regression? Once that case registers exactly one provider, this should be ready for another review.

oscarello:
Thanks @dplush — you were right again. The shadowing regression only proved the legacy callable was preserved; it never exercised the memory loader with `hermes_plugin` importable, where the loader's `mod.register(collector)` would have hit `hermes_plugin.register` and the collector would never receive the provider.

What changed (pushed as `cabd6a3`):

1. **`register(ctx)` now dispatches on ctx shape.** A memory-provider collector is uniquely identifiable by its `register_memory_provider` method — exactly what `plugins.memory._ProviderCollector` exposes. For a collector, `register` routes to the provider bridge and registers exactly one `MnemosyneMemoryProvider`. For any other ctx (legacy Hermes plugin context), it delegates to `hermes_plugin.register` — the legacy path is preserved, and the shadowing regression still passes (a non-collector ctx returns the legacy result unchanged).

2. **Combined loader regression added** (`test_actual_loader_registers_provider_with_hermes_plugin_importable` in tests/test_loader_registers_provider.py): the real `plugins.memory` loader against a temp `$HERMES_HOME` layout with `hermes_plugin` importable — `load_memory_provider("mnemosyne")` returns the provider, and the mock `hermes_plugin.register` is never invoked by the loader. Skips where Hermes isn't importable (repo CI), runs in Hermes deployments.

3. The shadowing test now loads the module the same way the loader does (synthetic namespace instead of a plain top-level import, which otherwise trips the stub-name collision on `mnemosyne.core`) and asserts both dispatch directions.

Rebased onto current main; CI is green across the full matrix (lint, docs-check, build, test 3.10–3.13). Local verification: real-loader tests 3/3 pass, ruff clean, full-suite failure list identical to the main baseline.

Once this case registers exactly one provider — it does, per the new regression — hoping this is ready for approval.

dplush:
Thanks for the updates. #609 has just landed on `main`, so this PR is now behind. Could you please rebase onto current `main` and request a fresh review when ready? The provider change itself looks independent of the docs merge.

oscarello:
Rebased onto current main (`9661915`) — the provider change is independent of the docs/wrapper merges (#609, #610). Ready for re-review, thanks!

dplush (approved):
Re-reviewed at 9661915: the real Hermes memory-loader path now registers exactly one provider with and without an importable hermes_plugin; the legacy plugin path remains preserved. Focused real-loader regressions and the full CI matrix are green.

oscarello:
Hey @AxDSAN — just rebased this onto latest main (`fe5cfd6`) and CI is fully green (7/7: lint, docs-check, build, tests 3.10–3.13). Wanted to make sure your requested changes are clearly addressed:

- `register` → `register_memory_provider` rename (no more shadowing of the `hermes_plugin` SDK)
- Dispatcher added so `register(ctx)` routes memory-provider collectors through the bridge even when `hermes_plugin` **is** importable
- Combined-loader regression tests prove exactly one provider registers

dplush approved the last revision — could you take another look when you get a chance? Thanks!

dplush (approved):
Final review of `1657e72` against merge-base `6785e28`: verified the real Hermes loader with and without an importable `hermes_plugin`. The dispatcher registers exactly one Mnemosyne provider for a memory collector while preserving the legacy plugin path. The reported ModuleNotFoundError concern is not reachable in the actual bridge import path. Focused real-loader tests passed (4 passed); current CI, CLA, and CodeRabbit are green. Approved.

Issue #559, opened 2026-07-27
bug(memoria): negation pattern has the same missing word boundary as #507


AxDSan:
Split out of #507 so the instruction fix in #558 stays reviewable.

## Summary

The `negation` pattern has the **identical** defect that #507 reports for `instruction`: it is not word-boundary anchored, and its captures are stored. The result is that third-person statements get recorded as first-person negations.

## Detail

`mnemosyne/core/beam.py:4776-4784`. Verified against main's `en` pattern:

| input | captured and stored |
|---|---|
| `The API never returns the full payload for large queries` | `I never returns the full payload for large queries` |
| `Hi never mind the config change we discussed` | `i never mind the config change we discussed` |

The first is the damaging one — a factual statement about an API becomes a stored first-person negation, which is exactly the #507 failure mode with a different pattern.

## Fix

Same one-character change as #558: a leading `\b`. Worth checking every locale's `negation` entry, not just `en`.

## Test note

#558's guard test (`test_every_locale_instruction_pattern_is_boundary_anchored`) deliberately only checks `pat["instruction"]`. Extend it to `negation` as part of this fix, so both patterns are pinned by the same guard.

Related: #507, #558.

PR #577, opened 2026-07-30
fix(memoria): remove literal backslash escapes from ru and es patterns

### LINKED ISSUE #560: bug(memoria): ru instruction/negation patterns use literal backslash-s, so Russian extraction is silently dead
Split out of #507. Pre-existing, unrelated to the word-boundary fix in #558.

## Summary

The `ru` locale's instruction pattern is a raw string containing **literal** `\\s+` and `[^.,;!?\\n]` — i.e. a backslash followed by `s`/`n`, rather than the whitespace and newline escapes. It therefore matches nothing at runtime, and Russian instruction extraction has been silently non-functional.

`mnemosyne/core/beam.py:4592`:

```python
'instruction': r'(?:всегда|никогда|должен|...)\\s+([^.,;!?\\n]{6,200})',
```

Compare the `en` pattern at `:4548`, which correctly uses `\s+` and `[^.,;!?\n]`.

## Why this needs its own issue

#558 adds a leading `\b` to all five locales including `ru`, for uniformity of the guard test. That means **after #558 the boundary test will report `ru` as healthy while `ru` extraction remains dead.** A green guard on a pattern that cannot match is worse than no guard, so this should not sit unfiled.

## Scope

Worth auditing the whole `ru` block rather than just the instruction line — if one pattern was written with doubled escapes, its siblings likely were too. Same for any other locale added by copy-paste from `ru`.

Ideally the fix comes with a test that asserts each locale's patterns actually match a representative sentence in that language, which would have caught this at authoring time.

Related: #507, #558.

Sanjays2402:
## Description

The `ru` instruction pattern was written with literal `\\s+` and `[^.,;!?\\n]` inside a raw string, so at runtime it required an actual backslash in the text — Russian instruction extraction matched nothing and was silently dead. As #560 suspected, the doubling was not confined to `ru`: six `es` patterns (`negation`, `decision`, `entity`, `sequence`, `instruction`, `preference`) carry the same doubled `\\n`, which turns the newline exclusion into an exclusion of the letter `n` and truncates every capture at the first `n` (`"siempre ejecuta las pruebas antes de subir a main"` captured only `" ejecuta las pruebas a"`).

Both are now single-escaped. No other locale is affected — the added guard test scans all five.

## Related Issue

Closes #560

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

Added to `tests/test_memoria_instruction_boundaries.py`, alongside the existing #507 boundary guard:

- `test_no_locale_pattern_contains_a_literal_backslash_escape` — parametrized over all five locales, rejects any `\\s`/`\\n`/`\\b`/`\\d` in a `MULTILINGUAL_PATTERNS` string. This is the "assert each locale's patterns actually work" guard the issue asked for.
- `test_ru_instruction_extracts_a_russian_directive`
- `test_es_instruction_capture_is_not_truncated_at_the_letter_n`

All four new cases (2 parametrized + 2) fail on `main` with only `beam.py` reverted and pass with the fix. `tests/test_memoria_instruction_boundaries.py`, `test_extraction.py`, `test_extract_parity.py`, `test_extraction_integration.py`, `test_multilingual_local_recall.py` and `test_c13b_extraction_diagnostics.py`: 108 passed. `ruff check` on the two changed files is clean before and after.

## Checklist

Version bump left to the maintainer's release flow; no user-facing README change (the patterns were documented as working).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

## Summary

- Fixed doubled regex escapes in Russian and Spanish MEMORIA instruction patterns.
- Restored correct extraction across punctuation and newline boundaries.
- Added regression tests, including end-to-end coverage through `BeamMemory.extract_and_store_facts()`.
- Updated the changelog.

## Architectural impact

- **Memory architecture:** No changes to working, episodic, or BEAM tiers. No changes to retrieval, consolidation, scoring, or veracity logic.
- **Privacy and local-first guarantees:** Preserved. The fix adds no services, telemetry, external dependencies, or data movement.
- **Agent integration:** No changes to Hermes, MCP, CLI, or public APIs. The existing `BeamMemory.extract_and_store_facts()` path now stores Russian and Spanish instructions correctly.
- **Sync and benchmarks:** No changes to synchronization, persistence architecture, or benchmark methodology.
- **Maintainability:** Targeted end-to-end tests protect locale patterns and boundary handling from future regressions. This is the right call: a narrow correctness fix with no architectural expansion.


dplush:
Thanks for the focused fix. Could you please rebase this onto current main and push the refreshed head? Happy to review it right after.

dplush:
Thanks for the clean rebase. One small test follow-up before this is ready: could you add a public-path regression through `BeamMemory.extract_and_store_facts()` for the two affected Russian and Spanish instruction forms? In addition to proving the extracted instruction, it should cover the newline and punctuation boundaries that caused this literal-escape bug.

The broader all-locale expansion suggested by CodeRabbit can stay out of this focused RU/ES fix. Once the two end-to-end cases are in, the pending CI and fresh review should give us a solid final signal. Thank you!

dplush (changes requested):
The fix and pattern-level regressions look correct. One focused coverage gap remains before merge: please add two public-path cases for the affected RU and ES instruction patterns via `BeamMemory.extract_and_store_facts()`. Assert the stored instruction and topic, including the relevant punctuation/newline boundary. The existing English E2E tests do not exercise language detection, pattern assembly, filtering, and persistence for these two changed paths.

Sanjays2402:
Added both end-to-end cases in 893bc0a, plus a merge of current main to clear the CHANGELOG conflict (3.15.1 had been cut, so my #560 entry moved back under Unreleased; no code change).

The new tests drive `BeamMemory.extract_and_store_facts()` and assert on the `memoria_instructions` row rather than a locally rebuilt regex, and each one pins both boundaries the doubled escape broke:

- `test_ru_instruction_is_stored_end_to_end` — input `"всегда запускай тесты перед пушем в main, это важно\nи не забывай про линтер"`. Asserts the stored instruction is `всегда запускай тесты перед пушем в main`, i.e. it stops at the comma and does not run into the next line.
- `test_es_instruction_is_stored_end_to_end_without_truncation` — input `"siempre ejecuta las pruebas antes de subir a main, por favor\ny luego avisa al equipo"`. Asserts the capture reaches `antes de subir` (the old `\\n` class truncated it at the first `n`, so it stopped at `pruebas a`), and again stops at the comma and the newline.

Both assert against the instruction matched from the directive under test, since a two-line input legitimately yields a second instruction from the second line — otherwise the newline assertion would be checking the wrong row.

RED-verified by checking out `mnemosyne/core/beam.py` from main and re-running: both new tests fail, the two pre-existing #507 end-to-end tests still pass. Restored, and the file is byte-identical to the reviewed version — the diff for this push is the test file only.

`tests/test_memoria_instruction_boundaries.py`, `test_extraction.py`, `test_c13b_extraction_diagnostics.py`: 80 passed. Left the all-locale expansion out as you suggested.

AxDSan:
Thanks for the focused memoria fix and the clean rebase. It's still showing a conflict against main, so it needs one more rebase onto current main. dplush's review thread is ready and waiting once the head is refreshed. Force-push and we'll get this moving.

Sanjays2402:
Updated the branch onto current main and resolved the CHANGELOG.md conflict while keeping both the #560 entry and the newer session-switch entry. The reviewed code and RU/ES public-path tests are unchanged. `tests/test_memoria_instruction_boundaries.py`: 25 passed. New head: `6afa205`.

dplush:
Thanks, the RU/ES public-path coverage now addresses the substantive review request. I found just two remaining steps before approval:

1. The changelog examples currently show four backslashes, while the historical raw-string defect used two literal backslashes. Please reduce those examples to the exact two-backslash form.
2. Please rebase onto current `main` (`9d34b6d`) and let the full matrix rerun.

The broader request for E2E coverage of every locale is out of scope for this focused RU/ES escape fix. #559 is a separate English negation-boundary issue and should stay separate.

PR #581, opened 2026-07-30
fix(hermes): log hermes_plugin import fallback at debug instead of silent pass

### LINKED ISSUE #578: [BUG] hermes_plugin silent import failure swallows plugin hook registration
## Problem

The Mnemosyne provider at `mnemosyne/hermes_memory_provider/__init__.py` has a `register()` function that attempts to import from a sibling `hermes_plugin` module:

```python
try:
    from hermes_plugin import register as _plugin_register
    _plugin_register(ctx)
except Exception:
    pass  # Graceful degradation — CLI still works without plugin tools
```

The bare `except Exception: pass` silently swallows the ImportError when `hermes_plugin` doesn't exist as a sibling directory. This causes:
- No error/warning in logs when hermes_plugin is absent
- Operators cannot distinguish "hooks registered" from "hooks silently dropped"

## Context

- Mnemosyne v3.0.0 refactor (commit 0ee4d80 "refactor(hermes): modularize provider") moved tool and hook registration INTO the provider class
- The `hermes_plugin` sibling directory was removed but the fallback import block remains
- Result: `hermes_plugin/` directory still exists in some deployments (empty, with stale __pycache__) and the import fails silently

## Suggested Fix

Replace the bare `except Exception: pass` with a logged debug message, since the absence is now expected:

```python
except Exception as exc:
    logger.debug(
        "mnemosyne hermes_plugin sibling not importable (expected since v3.0): %s",
        exc,
    )
```

## Impact

- Severity: Medium (silent failure pattern, not a functional regression since provider class handles hooks directly)
- Affects: All Hermes Agent deployments using Mnemosyne >= 3.0
- Workaround already applied locally: changed `except: pass` to `logger.debug(...)`

## References

- Commit: `0ee4d80 refactor(hermes): modularize provider, match Memori standalone pattern`
- File: `mnemosyne/hermes_memory_provider/__init__.py` lines ~3788-3795

paoloantinori:
Closes #578.

## Problem
`register()` in `hermes_memory_provider/__init__.py` had a fallback that imports the sibling `hermes_plugin` module, wrapped in a bare `except Exception: pass`. This silently swallowed any error (including real hook-registration failures from `_plugin_register(ctx)`), so operators could not distinguish "hooks registered" from "hooks silently dropped".

## Fix
Since the v3.0 modularization (commit `0ee4d80`) the `hermes_plugin` sibling was removed and hooks are registered by the provider class directly, so the fallback firing is expected. Replace the bare `pass` with a `logger.debug(...)` carrying the exception, correct the now-stale lead-in comment, and **narrow the catch to `ImportError`** with `_plugin_register(ctx)` moved into an `else:` clause. Only the expected missing-module error is swallowed; a real registration failure now propagates instead of being hidden. Debug level (not warning) because the condition is expected; matches the codebase's `logger.debug("...: %s", exc)` idiom (e.g. L1884) and the `except ImportError` optional-import idiom (L1268).

## Review
Ran `/simplify` (4 angles) and `/code-review` (high) on the diff before pushing, and re-ran both after narrowing the `except`. Verdict: ready to merge, no critical/important findings; the narrowing was mechanically verified (`ModuleNotFoundError` subclasses `ImportError`; a raising `_plugin_register` now propagates end-to-end). Altitude note (out of scope here): the entire try/except is dead code, `hermes_plugin/` having been deleted in `0ee4d80`, so the deeper fix is to delete the block; left to a separate maintainer follow-up rather than expanded into this patch.

## Testing
- `tests/test_provider_all_15_tools.py`: 32 passed.
- No new failures introduced. Two tests fail in a combined run (`TestUpdate::test_update_content`, `TestForget::test_forget_existing`) with `recall_diagnostics.py:40: AttributeError: module 'logging' has no attribute 'getLogger'`; this is a pre-existing test-isolation / `logging`-shadow issue that reproduces identically on clean `main` without this change (verified via stash).

## Follow-ups (out of scope)
- Delete the dead `hermes_plugin` import block here and the parallel stale `import hermes_plugin` in `tests/conftest.py` (both dead since `0ee4d80`).
- Optional regression test locking in "sibling absent -> no raise" and "registration failure propagates".

## Summary

- Logs the expected missing `hermes_plugin` import at debug level.
- Propagates missing transitive dependencies, plain `ImportError`, and plugin registration failures.
- Preserves Hermes provider operation and local-first SQLite memory behavior.
- Does not change MCP, CLI, tiered memory, retrieval, consolidation, veracity, sync, privacy, or benchmark methodology.
- Improves failure diagnostics and long-term maintainability. This is the right call.
- The provider tool test suite passed 32 tests. Two combined-run failures remain pre-existing `logging` shadowing issues reproducible on `main`.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -3830,14 +3830,18 @@ def register(ctx):
         handler_fn=mnemosyne_command,
     )
 
-    # Also register tools and hooks from hermes_plugin (sibling directory).
-    # This way a single symlink to hermes_memory_provider/ gives us the
-    # full Mnemosyne experience: CLI + tools + hooks.
+    # Legacy fallback: register a vendored hermes_plugin sibling's hooks if
+    # present. Removed from this repo in v3.0 (commit 0ee4d80); normally a
+    # debug-level no-op (issue #578). Only the expected missing-module error is
+    # swallowed; a real registration failure must propagate, not be hidden.
     try:
         _repo_root = str(Path(__file__).resolve().parent.parent)
         if _repo_root not in sys.path:
             sys.path.insert(0, _repo_root)
         from hermes_plugin import register as _plugin_register
+    except ImportError as exc:
+        logger.debug(
+            "hermes_plugin sibling not importable (expected since v3.0): %s", exc
+        )
+    else:
         _plugin_register(ctx)
-    except Exception:
-        pass  # Graceful degradation — CLI still works without plugin tools
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -3832,14 +3832,19 @@ def register(ctx):
 
     # Legacy fallback: register a vendored hermes_plugin sibling's hooks if
     # present. Removed from this repo in v3.0 (commit 0ee4d80); normally a
-    # debug-level no-op (issue #578). Only the expected missing-module error is
-    # swallowed; a real registration failure must propagate, not be hidden.
+    # debug-level no-op (issue #578).
     try:
         _repo_root = str(Path(__file__).resolve().parent.parent)
         if _repo_root not in sys.path:
             sys.path.insert(0, _repo_root)
         from hermes_plugin import register as _plugin_register
-    except ImportError as exc:
+    except ModuleNotFoundError as exc:
+        # Swallow only the expected absent-sibling case. A present-but-broken
+        # sibling raises a different ModuleNotFoundError.name (a missing
+        # transitive dependency) or a plain ImportError; those must propagate,
+        # not be hidden as "expected".
+        if exc.name != "hermes_plugin":
+            raise
         logger.debug(
             "hermes_plugin sibling not importable (expected since v3.0): %s", exc
         )
--- a/tests/test_register_fallback.py
+++ b/tests/test_register_fallback.py
@@ -1,10 +1,11 @@
 """Tests for the hermes_plugin fallback in hermes_memory_provider.register().
 
-Covers the three behaviours required by review on #581 (issue #578):
+Covers the four behaviours required by review on #581 (issue #578):
 - absent hermes_plugin sibling -> debug log, no raise;
 - a present-but-broken sibling (missing transitive dependency) -> the
   ModuleNotFoundError propagates instead of being swallowed;
-- a sibling whose register() raises -> that error propagates.
+- a sibling whose register() raises -> that error propagates;
+- a sibling that does not export register() -> plain ImportError propagates.
 """
 import logging
 import sys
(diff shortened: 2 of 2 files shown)


dplush (changes requested):
Thanks for narrowing the original broad catch and moving `_plugin_register(ctx)` outside it. One correctness detail remains: `except ImportError` is still too broad.

A present `hermes_plugin` can raise `ImportError` for a broken transitive dependency, which this branch currently logs as the expected absent sibling and then continues partially registered. Please catch `ModuleNotFoundError` only, verify `exc.name == "hermes_plugin"`, and re-raise anything else.

Please also add focused coverage for: absent `hermes_plugin` logs and returns, a broken nested import propagates, and `_plugin_register(ctx)` failures propagate.

dplush:
This PR is now behind current `main`. Could you please rebase and push a refreshed head, then rerun CI and request a fresh review?

paoloantinori:
@dplush addressed your review in 3e69fe7 (+ follow-up d4f96d2):

- Narrowed `except ImportError` -> `except ModuleNotFoundError`, re-raising unless `exc.name == "hermes_plugin"`. Only the genuinely-absent-sibling case is swallowed; a present-but-broken sibling (missing transitive dep -> ModuleNotFoundError with a different .name, or a plain ImportError like a missing `register` attr) now propagates instead of being hidden.
- `_plugin_register(ctx)` stays in the `else:` clause, so a failure inside it propagates.
- Coverage you asked for: absent sibling (logs, no raise), broken transitive import (propagates), `_plugin_register(ctx)` failure (propagates) — plus a plain-ImportError case (sibling without `register`) and a fixture-hygiene fix so a fake hermes_plugin can not leak into sys.modules.

CI is green. Ready for re-review when you have a moment.

dplush:
Hi @paoloantinori could you please rebase, so I can start a final review? Thx

dplush (changes requested):
Thanks for the careful exception narrowing and the focused tests. I re-reviewed the refreshed head and do not think we should merge this as an interim behavior change.

The branch now gives the obsolete checkout-symlink / `hermes_plugin` path a positive, tested compatibility contract: it continues to discover and register an arbitrary available sibling, while preserving separate failure semantics. That conflicts with the agreed direction in #651: `mnemosyne-hermes` is the supported provider, and we should migrate away from maintaining two integration stacks.

There is also a current test-coverage thread for the positive registration path. I would not expand that coverage here, because doing so would strengthen the legacy contract further.

Please close this PR in favor of #651. If maintainers explicitly decide that the legacy route must remain supported during migration, we can instead define a narrow transitional contract with provenance-bounded imports, complete tests, and a documented sunset path.

paoloantinori:
@dplush rebased onto current main. CI green. Ready for final review.

paoloantinori:
Thanks @dplush for the thorough review. Closing in favor of #651 — agreed that maintaining a tested compatibility contract for the legacy `hermes_plugin` checkout-symlink path conflicts with the migration direction. If the maintainers decide the legacy route needs a transitional contract during migration, it can be scoped as you described (provenance-bounded imports, complete tests, documented sunset path).

PR #582, opened 2026-07-30
fix: update global BEAM memories across sessions

### LINKED ISSUE #580: [BUG] mnenosyne_update doesn't update past memories
(Found and written up by my hermes agent. We recently installed mneno, went through some of my markdown memory system and injested and then have been kicking the tires on mneno and found a few things, including this.)

## Description

mnemosyne_update returns "status": "not_found" for any memory created in a past session, even when the memory ID is valid and the entry exists in working_memory. The underlying cause is in Mnemosyne core — two interacting issues in beam.update_working() and Memory.update().

## Environment

- OS: Linux (Ubuntu 24.04)
- Python version: 3.11.15
- Mnemosyne version/commit: 3.14.0 (installed via pip)
- Installation method: pip install mnemosyne-memory (Hermes bundled install)
- Hermes version: v0.19.0 (2026.7.20)

## Reproduction Steps

1. Create a memory with mnemosyne_remember(content="test fact") in session A
2. Note its memory ID from the response
3. Start a new session (session B)
4. Call mnemosyne_update(memory_id="<id from step 2>", content="updated text")
5. Observe the result is "status": "not_found" despite the memory existing

Minimal reproduction:

```python
from mnemosyne import remember, update

# Session A
mem_id = remember("test fact", scope="global")
print(f"Created: {mem_id}")

# In a new session (different process/instance):
result = update(mem_id, content="updated fact")
print(f"Update result: {result}")  # False
```

## Expected Behavior

A memory addressed by its unique ID should be updatable regardless of which session created it. mnemosyne_update(memory_id=<valid_id>, content="...") should return "status": "updated".

## Actual Behavior

Returns "status": "not_found" because beam.update_working() filters on session_id in addition to id:

```python
# beam.py:4239
cursor.execute(
    f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ? AND session_id = ?",
    params
)
```

Memory IDs are UUIDs (globally unique). The AND session_id = ? means any memory from a past session cannot be updated — the current session's ID never matches the stored session ID from creation time.

Secondary issue: Memory.update() (memory.py:576-600) also tries the legacy memories table first (which is intentionally empty in BEAM-only installations — documented as "Legacy memories table (kept for backward compatibility)" at memory.py line 85) and returns its rowcount as the success indicator.

### Mnemosyne Status

```
Python 3.11.15
mnemosyne-memory 3.14.0
Hermes Agent v0.19.0

working_memory: 186 rows
episodic_memory: 17 rows
memories: 0 rows (legacy table — empty by design)
```

### Optional: Database Info

```bash
ls -la ~/.hermes/mnemosyne/data/
  total 7708
  -rw-r--r-- 1 3592192 mnemosyne.db
  -rw-r--r-- 1   32768 mnemosyne.db-shm
  -rw-r--r-- 1 4247752 mnemosyne.db-wal

SELECT COUNT(*) FROM working_memory;    -- 186
SELECT COUNT(*) FROM episodic_memory;    -- 17
SELECT COUNT(*) FROM memories;          -- 0
```

## Suggested Fix

In beam.update_working() (beam.py:4237-4242), remove the session_id filter from the WHERE clause:

Before:
```
cursor.execute(
    f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ? AND session_id = ?",
    params
)
```
After:
```
cursor.execute(
    f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ?",
    params[:-1]
)
```

Memory IDs are globally unique UUIDs — WHERE id = ? is unambiguous. The session_id filter adds no protection and makes the tool unusable for cross-session updates.

Additionally, Memory.update() should either skip the legacy memories table when it's empty, or use working_memory as its primary lookup. The Hermes plugin already bypasses the legacy table correctly — the core MCP handler should follow the same pattern.

## Checklist

dplush:
## Summary

Fixes #580 without weakening session isolation.

- Authorizes cross-session updates only for `scope="global"` working memories.
- Keeps session-scoped memories private.
- Synchronizes the legacy row after BEAM has established visibility, while preserving legacy-only same-session compatibility.
- Adds BEAM, core, and Hermes provider dispatch regressions.

## Why this path

`get`, `forget_working`, and `invalidate` already make global working memories visible across sessions. `update_working` is the inconsistent mutation boundary. The legacy table has no scope column, so BEAM remains the authorization source before its compatibility row is updated.

## Why not remove `session_id` entirely?

That would allow a caller with a known ID to update a session-scoped memory from another session.

## Non-goals

No schema migration, ownership-model redesign, lifecycle change, or docs update.

## Verification

```text
MNEMOSYNE_NO_EMBEDDINGS=1 .venv/bin/python -m pytest -q tests/test_beam.py tests/test_provider_all_15_tools.py tests/test_mcp_server.py
184 passed, 9 skipped
```

## Summary

- Enables cross-session updates for `scope="global"` BEAM working memories.
- Preserves session isolation for `scope="session"` memories.
- Synchronizes legacy rows after BEAM authorization succeeds.
- Retains legacy-only same-session compatibility.
- Emits `MEMORY_UPDATED` only after committed writes.
- Adds core, BEAM, and Hermes provider dispatch regression coverage.

## Architectural impact

This is the right call for the tiered memory architecture. It aligns BEAM working-memory update authorization with existing global-memory visibility rules. It does not change episodic memory, retrieval, consolidation, veracity, schemas, ownership, or lifecycle behavior.

The privacy posture remains unchanged. Global memories support cross-session management. Session-scoped memories remain private.

The local-first design remains intact. Updates remain database-backed and require no remote service. Dual-written BEAM and legacy records remain synchronized, with rollback protection when the BEAM write fails.

`Mnemosyne.update()` now supports BEAM-only global memories and preserves legacy compatibility. Hermes provider dispatch benefits from the corrected core path. No direct MCP or CLI implementation changes are included.

The tests cover global, private, BEAM-only, dual-written, legacy-only, rollback, event ordering, transaction gating, and provider-dispatch behavior. No benchmark methodology changes are included.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -5035,9 +5035,13 @@ class BeamMemory:
             params.append(importance)
         if not updates:
             return False
+        # Match the visibility contract used by get(), forget_working(), and
+        # invalidate(): global memories are addressable across sessions, while
+        # session-scoped memories remain private to their creating session.
         params.extend([memory_id, self.session_id])
         cursor.execute(
-            f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ? AND session_id = ?",
+            f"UPDATE working_memory SET {', '.join(updates)} "
+            "WHERE id = ? AND (session_id = ? OR scope = 'global')",
             params
         )
         affected = cursor.rowcount
--- a/mnemosyne/core/memory.py
+++ b/mnemosyne/core/memory.py
@@ -848,19 +848,39 @@ class Mnemosyne:
         if not updates:
             return False
 
-        params.extend([memory_id, self.session_id])
         with _deferred_commits(self.conn):
-            cursor.execute(
-                f"UPDATE memories SET {', '.join(updates)} WHERE id = ? AND session_id = ?",
-                params
-            )
-            self.conn.commit()
+            # Authorize from the authoritative BEAM working row. Global rows
+            # may be updated cross-session; session-scoped rows remain private.
+            owner = cursor.execute(
+                """
+                SELECT session_id FROM working_memory
+                WHERE id = ? AND (session_id = ? OR scope = 'global')
+                """,
+                (memory_id, self.session_id),
+            ).fetchone()
+            if owner is None:
+                # Preserve owner-only updates for old legacy-only rows without
+                # letting their fallback authorize a foreign working-memory row.
+                params.extend([memory_id, self.session_id])
+                legacy_where = "id = ? AND session_id = ?"
+            else:
+                # The legacy mirror has no scope; update it by the BEAM row's
+                # owning session after BEAM has authorized the operation.
+                params.extend([memory_id, owner["session_id"]])
+                legacy_where = "id = ? AND session_id = ?"
 
-            # Sync BEAM working_memory
-            self.beam.update_working(memory_id, content=content, importance=importance)
+            cursor.execute(
+                f"UPDATE memories SET {', '.join(updates)} WHERE {legacy_where}",
+                params,
+            )
+            legacy_updated = cursor.rowcount > 0
+            self.conn.commit()
+            beam_updated = self.beam.update_working(
+                memory_id, content=content, importance=importance
+            )
 
         self._emit_wrapper("MEMORY_UPDATED", memory_id, content=content, importance=importance)
-        return cursor.rowcount > 0
+        return legacy_updated or beam_updated
 
     def invalidate(self, memory_id: str, replacement_id: str = None) -> bool:
         """Mark a memory as expired or superseded. Delegates to BEAM."""
--- a/mnemosyne/core/memory.py
+++ b/mnemosyne/core/memory.py
@@ -878,9 +878,12 @@ class Mnemosyne:
             beam_updated = self.beam.update_working(
                 memory_id, content=content, importance=importance
             )
+            updated = legacy_updated or beam_updated
 
+        if not updated:
+            return False
         self._emit_wrapper("MEMORY_UPDATED", memory_id, content=content, importance=importance)
-        return legacy_updated or beam_updated
+        return True
 
     def invalidate(self, memory_id: str, replacement_id: str = None) -> bool:
         """Mark a memory as expired or superseded. Delegates to BEAM."""
--- a/tests/test_beam.py
+++ b/tests/test_beam.py
@@ -38,6 +38,7 @@ def _same_utc_day_future(
         return None
     return candidate
 from mnemosyne.core.memory import Mnemosyne
+from mnemosyne.core.streaming import EventType
 
 
 @pytest.fixture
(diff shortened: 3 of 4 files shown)


dplush:
@AxDSan I rechecked #582 against the scope migration history and the current update contract. Two CodeRabbit findings are not applicable: the `scope="global"` migration default is intentional compatibility behavior, and id-based management updates are allowed for globally visible rows even when recall would filter them by validity/supersession.

The remaining valid issue is atomicity for dual-written memories: `Mnemosyne.update()` can commit the legacy update before the guarded BEAM update succeeds.

Would you be comfortable with this narrow fix after rebasing: preserve the existing scope and id-update contracts, but make the dual-write path commit both mirrors together or roll both back, emit `MEMORY_UPDATED` only after commit, while retaining the existing BEAM-only global and legacy-only same-session success paths?

I would add rollback, BEAM-only, legacy-only, and cross-session private-memory regressions.

dplush:
Co-maintainer implementation decision: #582 remains the canonical lane for #580. We will rebase it onto current `main` and replace the old cross-connection update flow with a narrow internal transaction-aware BEAM update path.

Contract: global rows remain cross-session manageable; session rows remain private; dual-written records commit together or roll back together; `MEMORY_UPDATED` is emitted only after commit. Ordinary `BeamMemory` connection ownership/lifecycle stays unchanged. No scope migration, recall-validity gate, or broader lifecycle refactor is in scope.

dplush:
Parking update: the prior narrow implementation direction is superseded by a schema-history finding documented on #580.

The v2.2 migration assigned `scope='global'` to pre-scope rows without provenance, while new writes default to `scope="session"`. Until we decide how to distinguish explicitly-global rows from migrated historical rows, this PR must not make `scope='global'` alone cross-session writable.

#582 remains a draft/reference only. The eventual safe design needs explicit scope provenance plus the separately required same-connection dual-write transaction contract.

dplush:
Unparking #582 after reconciling the established management contract: `get()`, `forget()`, and `invalidate()` already treat `scope='global'` as cross-session manageable, including mutation for forget/invalidate. `update()` must align with that existing predicate; session-scoped rows remain private.

The remaining implementation scope is the bounded same-connection dual-write transaction: both mirrors commit or roll back together, and `MEMORY_UPDATED` is emitted only after commit. No scope-provenance migration or broader lifecycle refactor is in this PR.

dplush:
Parking update after an independent transaction-boundary audit of the local rebase candidate.

The direct `Mnemosyne.update()` path can be made correct with a same-connection transaction, savepoint handling for caller-owned transactions, and strict derived-write error propagation. However, the public MCP batch path has a separate pre-existing atomicity boundary:

- `_deferred_commits()` performs a real final commit rather than preserving an already-open caller transaction;
- wrapper `remember` can reach a vector/embedding path with an internal commit before a later batch operation succeeds;
- therefore a `remember -> update` batch, or a later batch failure, cannot yet claim all-or-nothing behavior across `working_memory` and legacy `memories`.

This is broader than #580's cross-session update symptom and must not be folded into this draft opportunistically. #582 remains a draft/reference and will not be rebased or pushed until the shared batch/dual-write transaction contract is resolved under #726.

AxDSan:
Decision: this is the fix for #580. It changes exactly the one inconsistent boundary (update_working), keeps the same visibility contract as get(), forget_working() and invalidate(), and does not touch the migration path. Please take it out of draft and rebase over current main; the beam.py hunk conflicts after today's merges. I will merge once CI is green.

AxDSan:
One addition: #757 (now closed in favor of this) carried three useful regressions, test_update_global_memory_from_another_session, test_update_session_memory_from_another_session_denied and test_mnemosyne_update_global_memory_from_another_session. Please fold equivalent cases into this branch so both the BEAM and the Mnemosyne.update() paths are covered.

dplush:
The current head is green but not mergeable: within a caller-owned outer transaction, `MEMORY_UPDATED` can be observed after an inner savepoint release even though a later outer rollback reverts both stores.

**Recommended contract for #582: reject `update()` inside a caller-owned transaction with an explicit transaction error.** This fails closed, is narrow and reversible, and must not silently return `False`/`not_found` or suppress the event.

Connection-wide event queuing until the outermost commit would change shared commit, rollback, and event-ordering semantics, so that belongs in a separately specified #726 scope. We will add nested-transaction rejection regressions, retain the top-level committed-event coverage, and keep migrations/backfills out of this PR.

dplush:
@AxDSan Could you please confirm the remaining narrow transaction gate before I rebase?

For a caller-owned outer transaction, `update()` should reject with an explicit transaction error rather than release an inner savepoint and emit `MEMORY_UPDATED` before the outer transaction commits. That keeps the top-level dual-write fix atomic and fail-closed, without expanding #582 into connection-wide event queuing or the broader #726 transaction contract.

With that confirmation, I will rebase onto current `main`, retain the requested cross-session regressions from #757, add nested-transaction rejection coverage, and keep the scope limited to `update_working`.

AxDSan:
Confirmed, and you already have the pattern in-tree, so no new error path needed.

**Gate: yes.** Inside a caller-owned transaction, `Mnemosyne.update()` must reject before any write instead of releasing an inner savepoint and publishing `MEMORY_UPDATED` into a transaction that can still roll back. Fail closed, like you said.

Mirror the existing precedent:

- `MemoryTransactionStateError` (`mnemosyne/core/beam.py:37`)
- `consolidate_to_episodic()` (`mnemosyne/core/beam.py:6552-6561`) detects `self.conn.in_transaction` and raises before any write, conditional on an event emitter being registered, because the phantom event is the harm

Same detection, same pre-write placement. For `update()` the observable harm is `self._stream is not None`, since `_emit_wrapper` is a no-op without it (`mnemosyne/core/memory.py:705-719`). With no stream attached the dual write already sits inside the caller's savepoint and proceeds.

**In scope:**

1. Top-level dual-write fix: legacy row and BEAM in one same-connection transaction, one commit, `MEMORY_UPDATED` after commit. That is the fix for #580.
2. The #757 regressions folded in: `test_update_global_memory_from_another_session`, `test_update_session_memory_from_another_session_denied`, `test_mnemosyne_update_global_memory_from_another_session`.
3. Nested-transaction rejection coverage on both `BeamMemory.update_working()` and `Mnemosyne.update()`.

**Out of scope, agreed:** connection-wide event queuing and the shared batch/dual-write contract belong to #726. The CodeRabbit scope-migration finding stays inapplicable for the reason you gave on 2026-07-31, and the docstring-coverage warning is not a merge gate here.

**Two asks before you rebase:**

- Branch is `BEHIND` and `beam.py` moved ~1105 lines under it. The conflict is mechanical.
- Batch calls `beam.update_working()` directly (`mnemosyne/batch_tool.py:211-221`), not `Mnemosyne.update()`, so the gate does not regress batch. MCP `mnemosyne_update` does route through the wrapper (`mnemosyne/mcp_tools.py:248`), so make the message actionable (commit before updating) and keep the rejection ahead of any write.

LGTM in principle. Rebase and I merge on green CI.

Good stuff walking this one all the way through. The scope-provenance finding, the park, then reconciling against the `get()` / `forget_working()` / `invalidate()` predicate is the right shape for a boundary like this.

dplush:
@AxDSan Rebased and implemented the agreed narrow gate on `4ff19e4`.

The requested #757 coverage and nested transaction no-write/no-event regressions are included. All required CI jobs, CodeRabbit, and CLA are green. The branch is `CLEAN` and ready for your merge.

PR #592, opened 2026-07-31
fix(hermes): support root-only plugin installs

topbronson:
## Summary
- add a backward-compatible `link_profiles` installer option
- add `mnemosyne-hermes install --no-profile-links` for root-only installs
- prevent a root/default Hermes home from linking opted-in child profiles when explicitly requested

## Why
A root Hermes home can contain `profiles/`; the existing installer traverses those profiles even when the caller intends to install only the root provider. This flag makes that scope explicit while preserving existing fan-out behavior by default.

## Validation
- `uv run python -m pytest tests/test_install_hermes.py tests/test_install_status.py -q`
- result: 35 passed
- added regression coverage for root wrapper isolation and dry-run reporting

## Summary

Adds root-only Hermes plugin installation support.

- Adds backward-compatible `link_profiles` options to installer APIs.
- Adds the `mnemosyne-hermes install --no-profile-links` CLI flag.
- Persists the profile-link preference per Hermes home.
- Preserves the preference during upgrades and recovery.
- Preserves profile fan-out by default.
- Adds coverage for root-only installation, dry-run output, and upgrade behavior.
- Validation passed: 93 integration tests.

## Architectural impact

- **Agent integration:** Improves Hermes installation control through the Python API and CLI. MCP integration is unchanged.
- **Privacy:** Reduces unintended plugin exposure across Hermes profiles.
- **Local-first design:** Preserved. The change manages local plugin directories and preferences. It adds no remote service or sync behavior.
- **Core memory architecture:** No changes to working, episodic, or BEAM memory.
- **Veracity system and sync layer:** No changes.
- **Maintainability:** Preserves existing defaults and retains root-only preferences across upgrades.

This is the right call for Hermes profile isolation. It prevents new root-level installations from linking into opted-in child profiles while preserving existing default behavior.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -725,14 +725,40 @@ def _link_all_profiles(
     return linked
 
 
-def profile_links_enabled(*, hermes_home_path: str | Path | None = None) -> bool:
-    """Return whether opted-in profiles currently link to this home's plugin.
+def _profile_links_preference_path(hermes_home_path: str | Path | None = None) -> Path:
+    """Return the installer-managed profile-link preference for a Hermes home."""
+    base = Path(hermes_home_path).expanduser() if hermes_home_path else hermes_home()
+    return base / "plugins" / ".mnemosyne-profile-links.json"
 
-    Upgrade uses the observed state to preserve a prior root-only installation
-    without persisting separate installer configuration. Missing links mean the
-    safer root-only behavior; when no child profiles exist the distinction has
-    no effect.
+
+def _write_profile_links_preference(
+    enabled: bool,
+    *,
+    hermes_home_path: str | Path | None = None,
+) -> None:
+    """Persist the selected profile-link behavior for later upgrades."""
+    path = _profile_links_preference_path(hermes_home_path)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps({"link_profiles": enabled}) + "\n", encoding="utf-8")
+
+
+def profile_links_enabled(*, hermes_home_path: str | Path | None = None) -> bool:
+    """Return the selected profile-link behavior for a Hermes home.
+
+    New installs persist the explicit selection, which lets upgrades distinguish
+    the default enabled behavior from an explicit root-only installation even
+    when no opted-in child profile exists yet. Existing installs without the
+    preference file retain the legacy observed-link fallback.
     """
+    try:
+        preference = json.loads(
+            _profile_links_preference_path(hermes_home_path).read_text(encoding="utf-8")
+        )
+    except (OSError, ValueError):
+        preference = None
+    if isinstance(preference, dict) and isinstance(preference.get("link_profiles"), bool):
+        return preference["link_profiles"]
+
     target = plugin_target_dir(hermes_home_path)
     if not (target.is_symlink() or target.exists()):
         return False
@@ -1108,6 +1134,7 @@ def install_plugin(
         os.symlink(str(source), str(target))
         if link_profiles:
             _link_all_profiles(source, hermes_home_path=hermes_home_path, force=force)
+        _write_profile_links_preference(link_profiles, hermes_home_path=hermes_home_path)
         return target
 
     # Validate and fully write the replacement before removing a working wrapper.
@@ -1125,6 +1152,7 @@ def install_plugin(
         shutil.rmtree(staging_parent, ignore_errors=True)
     if link_profiles:
         _link_all_profiles(target, hermes_home_path=hermes_home_path, force=force)
+    _write_profile_links_preference(link_profiles, hermes_home_path=hermes_home_path)
     return target
 
 
@@ -1139,6 +1167,7 @@ def uninstall_plugin(*, hermes_home_path: str | Path | None = None) -> Path:
         target.unlink()
     elif target.exists():
         shutil.rmtree(target)
+    _profile_links_preference_path(hermes_home_path).unlink(missing_ok=True)
     return target
 
 
@@ -1204,6 +1233,14 @@ def cleanup_plugin(
                         actions.append("Reset config: memory.provider from 'mnemosyne' to unset")
         except Exception:
             pass
+
+    preference_path = _profile_links_preference_path(hermes_home_path)
+    if preference_path.exists():
+        if dry_run:
+            actions.append(f"Would remove: {preference_path}")
+        else:
+            preference_path.unlink()
+            actions.append(f"Removed: {preference_path}")
 
     return actions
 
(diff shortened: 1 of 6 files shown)


topbronson:
Addressed the remaining profile-link review findings in f8bb9c4: installation now persists the selected per-Hermes-home link preference (with legacy fallback), upgrades preserve it even when profiles are added later, and all upgrade/recovery instructions retain `--no-profile-links` for root-only installs. Validation: `uv run python -m pytest` in `integrations/hermes` (93 passed).

dplush:
Thanks for the careful follow-up on preference persistence and upgrade messaging. I found one remaining functional blocker in the current head.

`--no-profile-links` currently stops creating new profile links, but it does not remove Mnemosyne links that were created by an earlier default install. For example:

1. Install normally with an opted-in child profile.
2. Re-run with `--force --no-profile-links`, including a symlink-to-wrapper switch.
3. The child profile still has an active Mnemosyne link.

This conflicts with the documented contract: “install only at the selected Hermes home”. The existing test `test_root_only_install_does_not_remove_existing_child_profile_link` currently codifies that behavior.

Could you please make the root-only transition safely unlink only recognized Mnemosyne profile links? Foreign links and real directories must remain untouched. Please cover symlink-to-root-only, wrapper-to-root-only, both mode transitions, and a foreign profile link.

The preference persistence and upgrade/recovery handling otherwise look good. After the fix, please rebase onto current `main`, rerun CI, and request a fresh review.

topbronson:
Addressed the remaining root-only transition blocker in 628800f.

- `--force --no-profile-links` now snapshots recognized Mnemosyne profile symlinks before replacing the root plugin and removes those links after a successful install.
- Ownership detection uses installer-specific package/wrapper markers, so foreign symlinks and real directories are preserved, including foreign links routed through the root plugin path.
- Regression coverage includes symlink → root-only, wrapper → root-only, both cross-mode transitions, and the legacy wrapper topology that linked profiles directly to the old package source.

Validation:
- Hermes integration suite: 98 passed
- Fatal Ruff checks: passed
- Wheel + sdist build: passed
- Twine metadata checks: passed
- Branch is already based on current `main` (`22c60e2`).

@dplush Could you please take a fresh look?

dplush (changes requested):
Please rebase this onto current `main` before review. It now conflicts with merged #610 in `integrations/hermes/src/mnemosyne_hermes/install.py`.

This is semantic, not just mechanical: #610 deliberately validates and stages a wrapper replacement before removing a working target, and refuses an implicit wrapper-to-symlink downgrade. During the rebase, keep those guarantees while adding `link_profiles` and `--no-profile-links`. In particular, do not call `_prepare_plugin_target()` before wrapper-environment validation and staged replacement succeed.

The profile preference/unlinking behavior looks compatible in principle, but please resolve it on current main, rerun the Hermes integration tests, and request a fresh review on the rebased head.

topbronson:
@dplush Rebased onto current `main` and force-pushed the updated head. The implementation preserves #610's wrapper-safety ordering: wrapper environment validation and complete staging occur before `_prepare_plugin_target()`, and wrapper-to-symlink migration remains explicit (`--force --migrate-wrapper-to-symlink`).

Also added recognized-only cleanup for `--force --no-profile-links`, including legacy wrapper topologies, persisted profile-link preference/rollback, and quarantine-based protection against profile-link repoints. Foreign symlinks and real directories are preserved.

Validation on the rebased head:
- Hermes integration suite: 122 passed
- Full repository suite: 2389 passed, 2 skipped, 7 subtests passed
- Ruff fatal checks, compilation, docs checks, package build/Twine validation, and `git diff --check`: passed
- Independent final review: no findings

Please re-review.

integrations/hermes/src/mnemosyne_hermes/install.py
```diff
@@ -256,6 +267,51 @@ def _provider_init_is_valid(init_file: Path) -> bool:
...
+            and legacy_site.is_absolute()
+            and re.search(
+                r"^from\s+mnemosyne_hermes\s+import\s+\*(?:\s+#.*)?$",
+                source,
+                flags=re.MULTILINE,
+            )
+        ):
+            return True
+        return bool(
+            re.search(
```
dplush: This fallback is not installer-owned provenance: a foreign compatible provider can legitimately expose `plugin.yaml` name `hermes-mnemosyne` and `MnemosyneMemoryProvider`. During `--force --no-profile-links`, its child-profile link is then classified as managed and removed. I reproduced that path with an independent provider directory. Please require a marker that only this installer writes, and add the foreign-provider regression.

integrations/hermes/src/mnemosyne_hermes/install.py
```diff
@@ -751,38 +883,143 @@ def _verify_links(*, hermes_home_path: str | Path | None = None) -> bool:
...
+            if quarantined_snapshot is not None and _same_profile_link_identity(
+                snapshot, quarantined_snapshot
+            ):
+                quarantined.unlink()
+                moved = False
+                print(f"  Removed profile link: {target}")
+            elif not target.is_symlink() and not target.exists():
+                quarantined.replace(target)
+                moved = False
+        except OSError:
```
dplush, line 983: Please do not swallow this failure. If the quarantine/rename step fails, `install_plugin(..., link_profiles=False)` still succeeds and persists `link_profiles: false`, while the child profile remains linked to Mnemosyne. A controlled `PermissionError` from `Path.replace()` reproduces the inconsistent final state. Fail the install or restore the prior preference, and cover this error path.

dplush (changes requested):
Two root-only correctness gaps remain. Both are reproduced on this head and need regression coverage before approval.

topbronson:
@dplush The requested Hermes installer changes from review 4845810553 have been addressed in fa01c14:

- compatible foreign providers are preserved unless generated-manifest or exact legacy-wrapper provenance establishes installer ownership;
- profile-link cleanup failures now fail closed and restore the prior/effective profile-link preference, including the legacy absent-preference case.

The full GitHub CI matrix is green. When convenient, could you please take a fresh look?

dplush:
I found one remaining provenance blocker in the current head.

`_provider_init_is_mnemosyne()` treats a provider as installer-owned when it has the generic `hermes-mnemosyne` plugin name and a schema-valid wrapper manifest. A foreign compatible provider can provide those public markers. During `--force --no-profile-links`, its root target is then added to `recognized_profile_targets`, so existing child-profile links to that foreign provider are removed.

I reproduced this with a foreign directory containing `MnemosyneMemoryProvider`, `plugin.yaml` with `name: hermes-mnemosyne`, and a valid `mnemosyne-wrapper.json`: the child link was removed on the current head.

Please require installer-exclusive provenance before deleting profile links, or keep generic compatibility markers out of the ownership decision. Please also add this wrapper-manifest foreign-provider case as a regression test.

PR #603, opened 2026-08-02
fix(beam): fail loudly on incomplete vector rebuilds

### LINKED ISSUE #600: [BUG] Bulk reindex/repair_vec_working silently writes zero rows despite correct dry-run count and healthy embedding endpoint
## Bulk reindex/repair_vec_working silently writes zero rows despite correct dry-run count and healthy embedding endpoint

Environment:

mnemosyne-memory 3.15.1, mnemosyne-hermes 0.5.0
Hermes Agent v0.19.0 (2026.7.20)
Python 3.13.5, Docker
Remote OpenAI-compatible embedding endpoint (llama.cpp server), confirmed working via direct curl

Description:
After switching embedding_model/embedding_dim from local fastembed (384-dim) to a remote 1024-dim model, mnemosyne reindex --yes and mnemosyne_diagnose(repair_vec_working=true) both report success but write zero vectors.

Reproduction:

Change embedding_dim, embedding_model, embedding_api_url, embeddings_via_api: true in config.yaml
mnemosyne reindex --dry-run correctly reports N eligible rows (e.g. 652 working, 51 episodic)
mnemosyne reindex --yes completes in ~10s, reports working_memory reindexed: 0, episodic_memory reindexed: 0, but does recreate sqlite-vec tables at the new dimension
mnemosyne_diagnose(repair_vec_working=true, dry_run=false) (the Hermes tool path) also reports the parameter accepted but missing_backfill_rows count is unchanged after
Direct curl to the embedding endpoint returns valid embeddings immediately — endpoint itself is healthy
Single-item mnemosyne_remember() calls, both via CLI and live Hermes tool calls, correctly generate and persist vectors — only the bulk path is affected

Expected: reindex/repair populate vec_working for all eligible rows, matching the dry-run count.

Actual: 0 rows written every time, both entry points, with the vector tables emptied/recreated correctly but never repopulated.

Additional finding: mnemosyne doctor reports "sqlite-vec ANN index: not in use (extension not loadable), falling back to linear scan" even though sqlite_vec.load() works correctly when called directly in a standalone Python process against the same environment — suggesting mnemosyne's internal extension-loading path differs from the standard one.

Separately observed (may be related or distinct): recall appears to filter by session_id, and rows written with scope: session, session_id: default are not returned by mnemosyne_recall from live conversation sessions with different generated session IDs, even when the row has a valid vector. Possibly relevant to the reindex issue if reindex has a similar session/scope-filtering bug in its row-selection query.

This bug report and diagnostics were created for me with the help of claude.

dplush:
## Summary

Fixes #600 by making vector reindex and explicit `vec_working` repair fail loudly instead of reporting successful completion after partial processing or a failed postcheck.

- Reject missing, short, or failed embedding batches.
- Do not swallow reindex writes to `vec_working` or episodic `binary_vector`.
- Mark `repair_vec_working()` successful only after a complete, available, zero-gap postcheck.
- Return exit code 1 from `mnemosyne diagnose --repair-vec-working` when the requested repair fails.

## Why this path

`reindex_vectors()` owns the destructive rebuild and its success status. `repair_vec_working()` owns the backfill postcondition. Keeping the checks at those boundaries prevents false successful results without changing embedding configuration ownership.

## Non-goals

This does not change YAML/environment configuration precedence, embedding model selection, migrations, session scoping, or transactional recovery semantics. A failed rebuild can retain partial derived vectors, but it cannot claim `reindexed`.

## Why not #482 here

#482 needs a separately agreed configuration and migration contract. This PR only makes the existing destructive/recovery operations report incomplete work correctly.

## Reproduction and verification

Starting state: source rows exist and a reindex embedding batch returns `None`, is short, or a derived-vector write fails.

Before: the operation could return `status: reindexed` after partial work.

After: it raises an explicit error; repair returns `status: error` unless the postcheck proves complete coverage.

Tests:

```text
MNEMOSYNE_NO_EMBEDDINGS=1 uv run --extra test pytest -q \
  tests/test_reindex_vectors.py tests/test_beam.py tests/test_diagnose_optional_deps.py
# 89 passed, 10 skipped
```

## Summary

- Makes vector reindexing and `repair_vec_working()` fail loudly when embedding batches, vector writes, or coverage checks fail.
- Prevents successful status when working-memory or episodic vectors are missing after a rebuild.
- Returns exit code 1 from `mnemosyne diagnose --repair-vec-working` after non-dry-run failures.
- Adds regression coverage for incomplete embeddings, write failures, unavailable vector tables, and incomplete repairs.

## Architectural impact

- Strengthens vector integrity across working memory and episodic memory.
- Preserves tiered memory boundaries, BEAM behavior, retrieval strategies, consolidation, veracity, sync, and session scoping.
- This is the right call because incomplete derived memory cannot appear healthy.

## Integration and local-first impact

- Improves the CLI diagnostic contract.
- Does not change Hermes or MCP integration surfaces.
- Preserves Mnemosyne’s local-first guarantees and privacy posture.
- Failed rebuilds can retain partial derived vectors, but they cannot report success.

## Maintainability

- Explicit failure propagation improves operational visibility.
- Strict batch and coverage validation prevents silent data loss after embedding configuration changes.
- Regression tests protect destructive rebuild and repair paths.

### DIFF
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -285,6 +285,13 @@ def cmd_diagnose(args):
 
         if repair_vec_working:
             print("\n  vec_working repair requested")
+            repair_failed = any(
+                entry.get("check") == "vec_working_repair_status"
+                and str(entry.get("status", "")).upper() == "ERROR"
+                for entry in result.get("entries", [])
+            )
+            if repair_failed and not dry_run:
+                _fail("vec_working repair failed; inspect diagnostics above", exit_code=1)
 
         if fix_mode or (dry_run and not repair_vec_working):
             print("\n--- Auto-fix ---")
--- a/mnemosyne/cli.py
+++ b/mnemosyne/cli.py
@@ -285,12 +285,15 @@ def cmd_diagnose(args):
 
         if repair_vec_working:
             print("\n  vec_working repair requested")
-            repair_failed = any(
-                entry.get("check") == "vec_working_repair_status"
-                and str(entry.get("status", "")).upper() == "ERROR"
+            repair_statuses = [
+                entry.get("status")
                 for entry in result.get("entries", [])
+                if entry.get("check") == "vec_working_repair_status"
+            ]
+            repair_succeeded = bool(repair_statuses) and all(
+                status == "repaired" for status in repair_statuses
             )
-            if repair_failed and not dry_run:
+            if not dry_run and not repair_succeeded:
                 _fail("vec_working repair failed; inspect diagnostics above", exit_code=1)
 
         if fix_mode or (dry_run and not repair_vec_working):
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -2515,6 +2515,13 @@ def reindex_vectors(conn: sqlite3.Connection, *, batch_size: int = 64,
     def _count(sql: str) -> int:
         return int(conn.execute(sql).fetchone()[0])
 
+    def _commit_reindex_writes() -> None:
+        """Commit a rebuild boundary even when BEAM normally defers commits."""
+        if isinstance(conn, _BeamConnection):
+            conn._real_commit()
+        else:
+            conn.commit()
+
     def _embed_chunk(store: str, chunk) -> Any:
         try:
             vectors = _embeddings.embed([row["content"] for row in chunk])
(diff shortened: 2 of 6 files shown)


dplush:
@coderabbitai review

dplush:
@coderabbitai review

dplush:
@AxDSan Final gate is clean on de01760: full 3.10–3.13 CI, build, lint and docs-check are green; CLA is green; CodeRabbit has no unaddressed current-head findings; and the rebase plus both follow-up fixes received independent current-diff reviews. Could you please give the final human approval?

dplush:
Correction to my previous note: I verified the current head directly and there is no CLA check or CLA-bot status on this PR. The CI matrix and CodeRabbit are green, but I should not have called the CLA green. Could you please confirm whether the current rebased commit series needs any CLA action before approval?

dplush:
@AxDSan
FYI: the missing license/cla status seems to have been a CLA Assistant refresh issue after the force-pushes, not a missing signature. A correctly attributed empty chore: refresh CLA status commit triggered the check, which then passed and unblocked the merge.

PR #608, opened 2026-08-02
fix: widen sqlite-vec candidates before working-memory filters

### LINKED ISSUE #607: sqlite-vec working-memory search can miss valid rows after session filtering
## Description

`_wm_vec_search_sqlite()` asks sqlite-vec for the caller's final `k` nearest neighbors and applies working-memory session/scope predicates in the same query. sqlite-vec applies its KNN `k` limit before those relational predicates. If the nearest unfiltered vectors belong to other sessions, every initial candidate can be filtered out even though valid session-scoped or global rows exist farther down the vector neighborhood.

The empty sqlite-vec result then causes `_wm_vec_search()` to invoke the slower JSON-vector compatibility scan.

## Environment

- **OS:** Linux 6.8.0-136-generic
- **Python:** 3.11.15
- **Mnemosyne commit:** `ac9fc02`
- **Installation:** editable source checkout
- **Vector backend:** sqlite-vec, including int8 production coverage

## Reproduction

1. Create more than 500 closer `vec_working` vectors attached to `working_memory` rows in another session.
2. Create one slightly farther global row visible to the target session.
3. Request the top five vectors with `(session_id = ? OR scope = 'global')`.

Before the fix, the initial KNN neighborhood is exhausted by excluded rows, the SQL result is empty, and the valid global row is missed by the sqlite-vec path.

## Expected behavior

The sqlite-vec path should return the nearest rows that satisfy the caller's relational filters, up to the requested result count.

## Actual behavior

Relational filtering occurs after sqlite-vec has already limited the KNN neighborhood, so valid rows outside that unfiltered neighborhood are omitted. An empty result also incorrectly triggers the JSON-vector compatibility scan.

## Proposed fix

Start with a bounded overfetch neighborhood and widen it geometrically until either:

- enough filtered rows are found, or
- the vector table is exhausted.

Continue returning only the caller's requested `k`. Add a regression test with 600 closer vectors in another session and one valid global target.

## Validation

- Focused vector-search regression tests: 4 passed.
- Full `tests/test_beam.py`: 86 passed, 1 skipped.
- Production-derived copied-DB profiling showed the compatibility fallback taking roughly 2.9 seconds in a representative slow query; avoiding false fallback materially reduced end-to-end prefetch latency.

## Checklist

- [x] Existing issues and PRs searched for a duplicate
- [x] Bug reproduced consistently
- [x] Environment details included
- [x] Regression test prepared locally

pvspencer22:
## Description

sqlite-vec applies its KNN `k` limit before relational filters in the working-memory query. When the nearest unfiltered vectors belong to other sessions, applying session/scope filters can remove the entire initial neighborhood even though valid rows exist farther away. The empty result then incorrectly sends recall through the slower JSON-vector compatibility scan.

This change:

- starts with a bounded sqlite-vec candidate neighborhood;
- widens it geometrically until enough filtered rows are found or the vector table is exhausted;
- continues returning only the caller's requested result count; and
- adds a regression test with 600 closer vectors from another session ahead of one valid global row.

No dependencies or configuration are added.

## Related Issue

Closes #607

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactor / performance
- [x] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [x] Focused vector-search regression tests: `4 passed`
- [x] Full BEAM suite: `86 passed, 1 skipped`
- [x] Broader exact-head suite: `2214 passed, 2 skipped, 7 subtests passed` (excludes the two hanging script harnesses and three unrelated MCP modules)
- [x] Fatal Ruff checks: passed
- [x] Source distribution and wheel built; `twine check` passed
- [x] Generated API docs updated and verified
- [x] New regression test covers more than the initial 500-candidate neighborhood
- [x] Manual verification against a copied production database

## Checklist

- [x] Version bumped to `3.16.0` in `mnemosyne/__init__.py`
- [x] `CHANGELOG.md` updated
- [x] README unchanged because no public API or configuration changed
- [x] No new dependencies
- [x] Local-first and SQLite-only behavior preserved

**AI Assistance:** The implementation, regression tests, and initial analysis were prepared with AI agents.

## Summary

- Fixed sqlite-vec working-memory retrieval.
- The search now over-fetches KNN candidates before applying session and scope filters.
- The candidate limit starts at `max(k * 25, 500)` and doubles until it finds enough valid rows or exhausts the vector table.
- The search returns only the requested result count.
- Existing fallback behavior remains unchanged when sqlite-vec fails.
- Vector row counts are cached per connection and invalidated after local or external database changes.

## Architecture impact

- Improves the working-memory retrieval strategy.
- Preserves local-first and SQLite-only operation.
- Avoids unnecessary fallback to the slower JSON-vector scan.
- Does not change episodic memory, BEAM consolidation, the veracity system, synchronization, or benchmark methodology.
- No new dependencies or configuration are required.

## Validation

- Added regression coverage for `float32`, `int8`, and `bit` vector tables.
- Tests cover cross-session candidates, exhausted eligible rows, count caching, and cache invalidation.
- The adaptive widening strategy is the right call because relational filters now apply to a sufficiently broad nearest-neighbor set.
- The change strengthens privacy boundaries by ensuring session and scope filters determine the returned rows before the final result limit is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Model-refresh confidence: NaN cleared every gate and legacy text crashed sleep mid-batch.** JSON round-trips NaN and Infinity, and `parse_model_update_proposals` clamped NaN to 1.0 (`min` and `max` keep their first argument when a NaN comparison is False), so a NaN-confidence proposal became a top-importance memory; the auto-apply gate's `confidence < minimum` check is also False for NaN, so the same proposal reached the canonical store regardless of threshold. Separately, `apply_model_refresh_proposal` and sleep()'s proposal-remember call site converted stored confidence with a bare `float()`, so a legacy bank's text value (for example `"high"`) raised ValueError. The sleep call site runs after the claim commit, so that raise stranded the group's `consolidation_claimed_at` and orphaned every later group's claimed rows. Non-numeric and non-finite confidence now degrades per site: skipped at parse, 0.0 at the auto-apply gate, 0.5 at apply and at proposal importance. Finite values outside [0.0, 1.0] clamp to the domain bound on every path, so a persisted 2.0 can no longer clear the auto-apply gate or reach the canonical store unbounded. Hardening split out of #546 per review.
+
 ## [3.15.1] - 2026-07-30
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,6 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
-## [3.16.0] - 2026-08-02
-
-### Fixed
-
-- **sqlite-vec working-memory search could miss valid session-scoped results.**
-  KNN candidates are now overfetched before applying relational session and
-  scope filters. This prevents valid rows outside the initial unfiltered
-  neighborhood from being discarded and avoids incorrectly triggering the
-  slower JSON-vector compatibility scan.
-
 ## [3.15.1] - 2026-07-30
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ## [Unreleased]
 
+## [3.16.0] - 2026-08-02
+
+### Fixed
+
+- **sqlite-vec working-memory search could miss valid session-scoped results.**
+  KNN candidates are now overfetched before applying relational session and
+  scope filters. This prevents valid rows outside the initial unfiltered
+  neighborhood from being discarded and avoids incorrectly triggering the
+  slower JSON-vector compatibility scan.
+
 ## [3.15.1] - 2026-07-30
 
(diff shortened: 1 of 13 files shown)


dplush:
Nice focused fix. One small blocker before approval: the int8 match expression uses `vec_quantize_int8(?, "unit")`. Please use the SQL string literal `vec_quantize_int8(?, 'unit')`, matching the insert path and sqlite-vec’s documented form. With DQS disabled, the current expression can fail and silently fall back to the JSON scan.

The adaptive widening logic itself looks correct. An explicit exhausted-table regression would be useful follow-up coverage, but I do not see it as a blocker.

dplush:
Thanks for the careful follow-up. The sqlite-vec fix and the expanded regression coverage look good now.

I noticed one release-scope detail before approval: this PR updates `__version__`, the changelog, and generated API docs to `3.16.0`, while `hermes_memory_provider/plugin.yaml` remains at `3.15.1`. As a result, `python3 scripts/release.py check 3.16.0 --skip-build` still reports the release as not ready.

Since this is a focused bug fix and performance improvement, would you mind keeping the version and release-material updates for the separate documented release-preparation workflow after the fix lands on main? That would keep this PR focused and avoid leaving a partial release state behind.

dplush (approved):
Approved. Verified on 5ff5308: the focused sqlite-vec fix, regression coverage, clean release scope, green CI, and independent review are all in place.

PR #588, opened 2026-07-30
fix(hermes): align packaged manifest version

dplush:
## Summary

Fixes the remaining #579 drift in the packaged `mnemosyne-hermes` manifests.

- aligns both packaged manifests with the `0.5.0` package version
- adds regression coverage for core and packaged metadata, static manifests, and the installed wrapper manifest

## Why this path

Hermes discovers `plugin.yaml` statically before importing plugin Python, so runtime version lookup cannot eliminate drift. The tests make static parity a release gate instead.

## Non-goals

No installer refactor, release-automation redesign, tool-schema normalization, or core version bump. The legacy manifest is already aligned on current `main`.

## Verification

```text
pytest -q tests/test_plugin_manifest_versions.py integrations/hermes/tests/test_install_status.py::test_install_plugin_wrapper_creates_persistent_shim integrations/hermes/tests/test_install_cli_entrypoint.py::test_package_version_matches_distribution_metadata
4 passed
```

## Summary

- Aligns both packaged `mnemosyne-hermes` manifests with version `0.5.0`.
- Updates the `hermes_memory_provider` manifest to `3.15.2`.
- Adds regression tests for static manifests, package metadata, and the installed wrapper manifest.
- Prevents future version drift through explicit version contracts.

## Architectural impact

- **Core memory architecture:** Unchanged. Working, episodic, and BEAM memory layers are not modified.
- **Retrieval, consolidation, veracity, and sync:** Unchanged.
- **Privacy posture and local-first guarantees:** Preserved. The changes affect local packaging metadata only.
- **Hermes integration:** Improved. Static manifest parity supports Hermes discovery before Python import. This is the right call for reliable agent integration.
- **MCP and CLI:** Unchanged.
- **Long-term maintainability:** Improved through explicit version contracts and installed-wrapper validation.
- **Benchmark methodology:** Unchanged.
- **Architectural risk:** No risk identified. The changes do not erode the local-first design.

PR #616, opened 2026-08-03
fix: harden Hermes automatic prefetch relevance

### LINKED ISSUE #615: Bug: Hermes automatic prefetch injects memories on weak lexical overlap
## Problem

Hermes automatic context prefetch can silently inject an irrelevant high-importance memory when the current query and memory share only one broad lexical term. Because prefetch content is added before the model answers, a weak match can steer the turn even though an explicit recall would have exposed the result for normal judgment.

Canonical prefetch has a related configuration bug: `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` is documented as adding deployment-specific generic terms, but the current parser replaces the built-in set when the variable is configured.

## Expected behavior

Silent automatic prefetch should require stronger topical evidence than explicit recall:

- multiple distinctive shared terms with meaningful query coverage for ordinary working memories;
- a one-token canonical match only when the term is rare across that owner's canonical facts and covers enough of the query;
- stronger multi-token canonical matches should displace weaker singleton candidates;
- deployment-specific generic-token configuration should extend, not replace, built-in generic terms;
- explicit `mnemosyne_recall` behavior should remain unchanged.

## Proposed fix

Add configurable lexical thresholds for silent prefetch, reject non-finite coverage settings, use per-owner canonical document frequency for the singleton exception, make generic-token configuration additive, and add regression coverage for broad-term false positives plus valid rare and multi-token matches. Include the Hermes integration tests in the repository CI matrix so these rules remain protected.

**AI Assistance:** The implementation, regression tests, and initial analysis were prepared with AI agents.

pvspencer22:
## Summary

Hermes automatic context prefetch can silently inject irrelevant high-importance memories when a query and memory share only one broad lexical term.

Fixes #615.

This change:

- preserves `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` replacement semantics and adds a separate additive configuration for deployment-specific tokens;
- requires configurable distinctive lexical evidence before silent working-memory injection;
- scores canonical candidates using distinctive overlap, query coverage, and per-owner canonical document frequency;
- permits a one-token canonical match only when it meets both the configured coverage threshold and rare-token frequency limit;
- prefers stronger multi-token canonical candidates over weaker singleton candidates;
- rejects non-finite query-coverage configuration and falls back to the conservative default;
- adds the Hermes integration suite to the repository CI test matrix;
- leaves explicit `mnemosyne_recall` behavior unchanged.

## Configuration

- `MNEMOSYNE_PREFETCH_MIN_DISTINCTIVE_TOKENS` (default `2`, minimum `1`)
- `MNEMOSYNE_PREFETCH_MIN_QUERY_COVERAGE` (default `0.30`, clamped to `[0.0, 1.0]`)
- `MNEMOSYNE_PREFETCH_CANONICAL_RARE_TOKEN_MAX_FREQUENCY` (default `1`, minimum `0`; `0` disables the singleton exception)
- `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` replaces the built-in generic-token set when configured.
- `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS` adds deployment-specific tokens to the effective generic-token set.

These thresholds apply only to silent automatic prefetch injection. Explicit recall remains available for broader semantic exploration.

## Regression coverage

Tests cover:

- rejecting irrelevant automatic injection based on one broad shared term;
- preserving a valid multi-term `Cedar Bakery` match;
- preserving replacement semantics for the existing generic-token configuration;
- adding deployment-specific generic tokens through the separate additive configuration;
- rejecting common singleton canonical matches;
- allowing a rare singleton synthetic `Caldrin` match and proving rarity is scoped to the requested owner;
- allowing a valid two-token `archive boundaries` match;
- disabling singleton canonical matches through the rare-token frequency setting;
- keeping singleton coverage and rarity safeguards active when the minimum distinctive-token setting is `1`;
- respecting a higher configured distinctive-token minimum.
- treating `NaN` and infinite query-coverage values as invalid.
- preserving broad canonical merging through the public `mnemosyne_recall` tool while applying the stricter filter only to silent prefetch.

## Verification

```text
MNEMOSYNE_NO_EMBEDDINGS=1 uv run --frozen pytest -q integrations/hermes/tests/test_prefetch_hardening.py
15 passed

MNEMOSYNE_NO_EMBEDDINGS=1 uv run --python 3.13 --frozen pytest -q integrations/hermes/tests
114 passed

MNEMOSYNE_NO_EMBEDDINGS=1 uv run --frozen pytest -q \
  tests/test_prefetch_profiles.py \
  tests/test_prefetch_identity_always_inject.py \
  tests/test_hermes_memory_provider_thread_isolation.py \
  tests/test_model_refresh_stress.py \
  tests/test_prefetch_low_quality.py \
  tests/test_prefetch_content_truncation.py
43 passed

uv run --frozen ruff check integrations/hermes/src/mnemosyne_hermes/__init__.py integrations/hermes/tests/test_prefetch_hardening.py --select E9,F63,F7,F82
All checks passed

CI workflow validation
pytest tests/ -v --tb=short
pytest integrations/hermes/tests/ -v --tb=short
Hermes regression suite runs in a separate pytest process in every Python matrix job

git diff --check
passed

uv build --out-dir <temporary-directory>
uvx --from twine twine check <temporary-directory>/*
wheel and sdist built; both passed
```

A repository-wide run also exposed the existing `tests/test_mnemosyne_stats.py::test_performance` timing threshold on this host (18.4s versus the fixed 5s limit). The unchanged `upstream/main` tree fails the same test at 19.0s, so this is baseline/environment behavior outside this integration-only diff.

## Release metadata

No version, manifest, generated-documentation, or changelog release material is included. Those updates belong to the project's separate release-preparation workflow; open PR #588 independently owns the existing `0.5.0` packaged-manifest parity fix.

**AI Assistance:** The implementation, regression tests, and initial analysis were prepared with AI agents.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,8 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
@@ -188,8 +190,11 @@ jobs:
           print('embedding model cached')
           "
 
-      - name: Run tests
-        run: pytest tests/ integrations/hermes/tests/test_wrapper_bootstrap.py integrations/hermes/tests/test_canonical_tools.py -v --tb=short
+      - name: Run core tests
+        run: pytest tests/ -v --tb=short
+
+      - name: Run Hermes integration tests
+        run: pytest integrations/hermes/tests/ -v --tb=short
 
   build:
     runs-on: ubuntu-latest
--- a/integrations/hermes/README.md
+++ b/integrations/hermes/README.md
@@ -153,8 +153,8 @@ No required config. Everything defaults to `~/.mnemosyne/`. Optional overrides:
 | `MNEMOSYNE_PREFETCH_MIN_DISTINCTIVE_TOKENS` | `2` | Shared non-generic terms required for automatic prefetch injection |
 | `MNEMOSYNE_PREFETCH_MIN_QUERY_COVERAGE` | `0.30` | Minimum fraction of non-generic query terms covered by a prefetched memory |
 | `MNEMOSYNE_PREFETCH_CANONICAL_RARE_TOKEN_MAX_FREQUENCY` | `1` | Maximum canonical document frequency that permits a one-token match (`0` disables the exception) |
-| `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` | built-in generic-token set | Complete replacement for the canonical generic-token set |
-| `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS` | _(empty)_ | Extra owner/deployment terms added to the configured generic-token set |
+| `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` | path-specific built-in canonical set | Complete replacement for the canonical generic-token set used by automatic and explicit canonical lookup; does not affect working/episodic prefetch |
+| `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS` | _(empty)_ | Extra owner/deployment terms added to automatic canonical prefetch only |
 | `MNEMOSYNE_DEFAULT_SCOPE` | `session` | Default scope for remember (`global` enables cross-session immediate recall) |
 
 Or in `~/.hermes/config.yaml`:
--- a/integrations/hermes/README.md
+++ b/integrations/hermes/README.md
@@ -150,6 +150,11 @@ No required config. Everything defaults to `~/.mnemosyne/`. Optional overrides:
 | `MNEMOSYNE_SYNC_TURN_ASSISTANT_LIMIT` | `800` | Assistant content truncation in `sync_turn()` (`0` = no limit) |
 | `MNEMOSYNE_FACT_RECALL_ENABLED` | `false` | Merge LLM-extracted facts into standard recall |
 | `MNEMOSYNE_PREFETCH_CONTENT_CHARS` | `0` | Per-memory prefetch content cap (`0` = full content) |
+| `MNEMOSYNE_PREFETCH_MIN_DISTINCTIVE_TOKENS` | `2` | Shared non-generic terms required for automatic prefetch injection |
+| `MNEMOSYNE_PREFETCH_MIN_QUERY_COVERAGE` | `0.30` | Minimum fraction of non-generic query terms covered by a prefetched memory |
+| `MNEMOSYNE_PREFETCH_CANONICAL_RARE_TOKEN_MAX_FREQUENCY` | `1` | Maximum canonical document frequency that permits a one-token match (`0` disables the exception) |
+| `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` | built-in generic-token set | Complete replacement for the canonical generic-token set |
+| `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS` | _(empty)_ | Extra owner/deployment terms added to the configured generic-token set |
 | `MNEMOSYNE_DEFAULT_SCOPE` | `session` | Default scope for remember (`global` enables cross-session immediate recall) |
 
 Or in `~/.hermes/config.yaml`:
--- a/integrations/hermes/src/mnemosyne_hermes/__init__.py
+++ b/integrations/hermes/src/mnemosyne_hermes/__init__.py
@@ -291,18 +291,25 @@ def _parse_token_set_env(key: str, default: Set[str]) -> Set[str]:
     return tokens or set(default)
 
 
-# Generic schema/system labels do not, by themselves, prove a canonical fact is
-# relevant to a turn. Deployments can replace this list with
-# MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS or extend the effective list with
-# MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS.
-_PREFETCH_CANONICAL_GENERIC_TOKEN_DEFAULTS = {
+# Generic schema/system labels do not, by themselves, prove a memory is
+# relevant to a turn. This fixed set governs ordinary working/episodic
+# automatic prefetch; canonical-specific deployment tuning must not alter that
+# established path.
+_PREFETCH_LEXICAL_GENERIC_TOKENS = {
     "user", "owner", "assistant", "agent", "system", "profile", "identity", "default",
     "preference", "preferences", "recommendation", "recommendations",
 }
 
-# Explicit recall keeps the pre-hardening generic-token contract. The extra
-# preference/recommendation terms and prefetch environment tuning below are
-# intentionally limited to silent automatic context injection.
+# Canonical automatic prefetch starts from the same conservative defaults, but
+# deployments can replace this list with
+# MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS or extend the effective list with
+# MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS.
+_PREFETCH_CANONICAL_GENERIC_TOKEN_DEFAULTS = set(_PREFETCH_LEXICAL_GENERIC_TOKENS)
+
+# Explicit recall keeps the pre-hardening default generic-token contract while
+# continuing to honor the established canonical replacement variable. The new
+# preference/recommendation defaults and additive tuning remain limited to
+# silent automatic context injection.
 _CANONICAL_RECALL_GENERIC_TOKENS = {
     "user", "owner", "assistant", "agent", "system", "profile", "identity", "default",
 }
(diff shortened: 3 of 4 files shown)


dplush:
Thanks, this is a thoughtful hardening pass and the new regression coverage is useful.

One compatibility concern before merge: `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` previously replaced the default token set, whereas this PR makes it additive. A deployment that intentionally configured a complete custom list will now inherit the new defaults as well. Since this set also feeds the general prefetch filter, the change can suppress additional automatic prefetches beyond canonical-row filtering.

Could we retain the established replace semantics, or add a separate `..._EXTRA_TOKENS` variable for additive customization?

pvspencer22:
Good catch—agreed. I’ve restored the established replacement semantics for `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` and added `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS` for additive customization in b9d7ef1. The README and regression coverage now exercise both paths, so existing deployments retain their behavior while new deployments have an explicit additive option.

dplush (changes requested):
Thanks for restoring the established replacement semantics and providing an explicit additive variable. The current head is clean and the focused integration suite passes locally.

Two regression gaps need closing before approval:

1. The new per-owner document-frequency contract is not exercised. `FakeCanonicalStore.list()` only permits `"default"` and always returns the same rows, so it cannot catch an implementation that computes rarity from another owner’s canonical rows. Please add a second owner where a token is common there but rare for `default`, and assert the singleton remains eligible for `default`.

2. The replacement-semantics tests set `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` but do not clear `MNEMOSYNE_PREFETCH_CANONICAL_EXTRA_GENERIC_TOKENS`. An inherited additive setting can make those assertions environment-dependent. Clear it explicitly in the base-config tests.

No additional production findings from this review.

pvspencer22:
Closed both requested regression gaps in 7852444: the fake canonical store now models multiple owners, and the new regression proves a token that is common for another owner remains rare and eligible for the requested owner. The module’s autouse fixture also clears both replacement and additive generic-token environment variables so the base configuration tests cannot inherit deployment settings. Validation: 14 focused hardening tests and 113 Hermes integration tests passed on Python 3.13; Ruff and `git diff --check` are clean.

pvspencer22:
Addressed the latest test-hardening feedback in 531fece. The owner-isolation regression now explicitly sets the rarity threshold, exercises both owners, and verifies the repeated token is rejected for the owner where it is common. The autouse fixture now clears all five prefetch tuning variables so the suite is independent of runner configuration. Validation remains green: 14 focused tests, 113 Hermes integration tests on Python 3.13, Ruff, and `git diff --check`.

dplush (changes requested):
Thanks for the owner-isolation and environment-isolation fixes. The silent-prefetch hardening itself now looks well covered.

One major contract regression remains: `_handle_recall()` also calls `_canonical_prefetch_rows()` when serving the public `mnemosyne_recall` tool. The new singleton/document-frequency filter therefore suppresses explicit canonical recall, not only silent automatic injection. For example, with two canonical facts containing `Bourdain`, `mnemosyne_recall(query="Bourdain recommendation lens")` returned both rows at the merge base but returns none on this head because document frequency is 2 > 1.

The stricter relevance thresholds should be limited to silent prefetch. Please preserve the prior canonical merge behavior for explicit `mnemosyne_recall` (or obtain an explicit decision and document this as a breaking recall-contract change). Add a regression through `_handle_recall()`/the public tool for the two-Bourdain-facts case.

All previously requested prefetch-hardening items are addressed; this is the only remaining functional blocker.

pvspencer22:
Fixed the explicit-recall contract regression in 1e750a8. `_handle_recall()` now uses a separate canonical-recall path that preserves the merge-base lexical/coverage behavior; the document-frequency and stricter distinctive-token rules remain confined to silent automatic prefetch. Added a regression through the public `mnemosyne_recall` tool with two matching synthetic canonical facts, confirming both are returned. Validation: 15 focused tests and 114 Hermes integration tests passed on Python 3.13; targeted Ruff and `git diff --check` are clean.

dplush:
Thanks for separating the document-frequency rule from the public recall path. One compatibility regression remains on the current head.

`_canonical_recall_rows()` still uses `_prefetch_canonical_generic_tokens()`, while this PR adds `preference(s)` and `recommendation(s)` to that default set. As a result, explicit canonical recall can now lose facts that matched at the merge base. With a canonical row `Taylor recommendations emphasize reliability.`, querying `recommendations` returned the row on `fe5cfd6` but returns none on `67e694f`.

Issue #615 requires explicit `mnemosyne_recall` behavior to remain unchanged. Could the explicit-recall path preserve the previous generic-token semantics, or otherwise keep the new prefetch-specific tokens strictly scoped to silent prefetch? Please add a public-tool regression for `recommendations` or `preference` as well.

No other functional findings from this re-review.

pvspencer22:
@dplush Ready for re-review at `67e694f`.

This head keeps the stricter relevance filtering confined to silent automatic prefetch, restores the established broad canonical merge behavior for explicit `mnemosyne_recall`, and includes a public-tool regression using wholly synthetic canonical fixtures.

The current GitHub CI matrix, lint, docs, build, CodeRabbit, and CLA checks are green. Local Python 3.13 validation also passed all 114 Hermes integration tests.

dplush (changes requested):
Re-review on current head `67e694f`: one compatibility blocker remains.

`_canonical_recall_rows()` now uses `_prefetch_canonical_generic_tokens()` (lines 360 and 368). The PR extends that default set with `preference(s)` and `recommendation(s)` for the silent-prefetch filter. This changes explicit `mnemosyne_recall`, despite #615 requiring its established broad canonical merge semantics to stay unchanged.

Minimal reproduction: a canonical row `Taylor recommendations emphasize reliability.` is returned by the merge-base canonical path for query `recommendations`; on this head `_canonical_recall_rows(..., "recommendations")` returns `[]`, because its only overlap becomes generic. The current public-tool regression covers `Caldrin`, not either newly generic token.

Please keep the new prefetch-specific tokens scoped to silent prefetch, or give explicit recall its pre-existing generic-token set. Add a public-tool regression for `recommendations` or `preference`.

All current CI checks are green; this is a semantic compatibility issue not covered by them.

pvspencer22:
@dplush Addressed in `b76365b`.

`_canonical_recall_rows()` now uses the fixed pre-hardening generic-token set, independent of silent-prefetch defaults and environment tuning. The automatic prefetch path continues to use the stricter prefetch-specific set.

I added a public-tool regression with synthetic fixtures that verifies the same `recommendations` overlap is rejected by silent prefetch but returned by explicit `mnemosyne_recall`.

Local validation: 115 Hermes integration tests passed on Python 3.13; targeted Ruff and `git diff --check` passed.

dplush (changes requested):
Two current-head compatibility blockers remain.

1. `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` is documented and historically scoped as the canonical generic-token set. This PR now passes it to `_prefetch_has_distinctive_lexical_evidence()` for every ordinary `Beam.recall()` row, not only canonical rows. An existing deployment-level setting can therefore newly suppress normal working or episodic automatic prefetch.

Reproduction: a normal `source="preference"` recall row reading `SampleOwner prefers Cedar Bakery over Harbor Bakery.` is injected for `Cedar Bakery preference` with the variable unset, but is rejected after configuring that canonical-token variable with a two-token override containing both `cedar` and `bakery`. Before this PR the variable was only consumed in `_canonical_prefetch_rows()`, so it could not alter the non-canonical prefetch path.

2. The new `_canonical_recall_rows()` uses the fixed `_CANONICAL_RECALL_GENERIC_TOKENS`, ignoring the existing `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS` override. At the PR base, the public `mnemosyne_recall` canonical path used `_canonical_prefetch_rows()` and honored that setting. Thus this head still changes explicit recall for configured deployments despite the stated unchanged-contract goal.

Minimal base/head reproduction: configure the existing generic-token override with `sampleowner`; use a canonical fact `SampleOwner budget approach.` and invoke public recall for `SampleOwner`. Base `fe5cfd6` returns no canonical result because the configured term is generic; head `b76365b` returns the fact because the fixed set does not include `sampleowner`.

Please separate the new general lexical-evidence filter from the canonical-token customization, and preserve the established configured generic-token behavior for explicit canonical recall. Add public-handler regressions for both cases. The earlier default-case `recommendations` regression looks addressed; these are configuration-compatibility cases.

dplush (changes requested):
The current implementation resolves the two configured-deployment compatibility blockers, and the normal-prefetch isolation, explicit-recall semantics, CI, and focused regression matrix now look good.

One small regression gap remains before approval: please add a public-handler test that combines the existing canonical generic-token override with one of the new automatic-prefetch-only generic words. For example, set `MNEMOSYNE_PREFETCH_CANONICAL_GENERIC_TOKENS=recommendations`, call `mnemosyne_recall(query="recommendations")`, and assert the canonical result is excluded.

The current suite separately covers `recommendations` with the default set and a custom override using `sampleowner`, but not the combined contract. That narrow test will protect the intended separation between configured explicit recall and the stricter silent-prefetch defaults.

PR #619, opened 2026-08-04
fix(hygiene): unwrap audit envelope before cleaning candidates

### LINKED ISSUE #606: [BUG] hygiene clean <audit.json> crashes with TypeError, it iterates the audit envelope instead of the candidates array
### Description

The documented two-step noise-cleanup workflow is:

```bash
mnemosyne hygiene audit --json > audit.json   # step 1: works
mnemosyne hygiene clean audit.json            # step 2: crashes
```

Step 1 (`hygiene audit --json`) writes a JSON **envelope**: `{"total_scanned": N, "candidates": [...]}`. Step 2 (`hygiene clean`) is supposed to read that file and act on the `candidates` array, but it actually iterates the **top-level dict** as if it were a list of candidates. So the very first thing it does is hand `enumerate()` a dict, which yields **string keys** (`"total_scanned"`, `"candidates"`, …) instead of candidate objects  and `c["memory_id"]` immediately throws:

```
TypeError: string indices must be integers, not 'str'
```

The crash happens *before any cleanup logic runs*, so the command is 100% non-functional. `audit` works; only `clean` is broken.

### How I found it

I was running the documented cleanup path against a real store (`hygiene audit --json` → `hygiene clean`). The audit file looked correct on disk, a proper `{"total_scanned": 400, "candidates": [...]}` envelope with each candidate carrying `memory_id`, `table_name`, `noise_score`, etc. But `clean` blew up on the first line of the loop. Reading the source made it obvious why: `clean` never unwraps the envelope it was just given by `audit`.

So this is a contract mismatch *between two commands in the same subcommand*: `audit --json` emits an envelope, `clean` expects a bare list. Neither side is "wrong" alone, they just don't agree, and the disagreement is fatal.

### Who and what is affected (blast radius)

I want to be precise here, because there are three different ways to drive this engine and only one of them reaches the bug:

| Layer | How it's invoked | Exposes `hygiene`? | Hits the bug? |
|---|---|---|---|
| **A. Agent (Hermes) native tools** (the assistant's `mnemosyne_*` toolset) | `mnemosyne_stats`, `mnemosyne_remember`, `mnemosyne_triple_add`, … | **No**,  there is no `hygiene`/`hygiene_clean` tool exposed to the agent | **No** |
| **B. Hermes user wrapper** | `hermes mnemosyne stats|sleep|...` | **No**, the wrapper only whitelists `stats, sleep, version, inspect, clear, doctor, export, import`; `hygiene` is not in that list (it correctly rejects `hermes mnemosyne hygiene` with `invalid choice`) | **No** |
| **C. Bare upstream binary** | `~/.hermes/hermes-agent/venv/bin/mnemosyne hygiene clean <file>` (the standalone `mnemosyne-memory` CLI) | **Yes**, the package's own CLI ships the `hygiene` subcommand | **Yes, the bug lives here** |

**Net:** the bug is in the standalone `mnemosyne` CLI's `hygiene clean` subcommand. It does **not** affect Hermes's own memory operations (Layer A) and is **not** reachable through the `hermes mnemosyne` wrapper (Layer B), that wrapper doesn't even expose `hygiene`. It only bites someone who invokes the bare `mnemosyne` binary's `hygiene clean` directly (Layer C). Mnemosyne-memory is a published PyPI package with its own CLI, so Layer C is a legitimate, supported usage, this is a real upstream bug, not a misuse.

No data is touched or corrupted: the crash is in the *load* step, before any delete/archive operation. Worst case is "the documented cleanup command can't run."

### Environment

- **OS:** `Debian GNU/Linux 13 (trixie), kernel 6.12.100+deb13-amd64`
- **Python version:** `3.11.15 (uv managed Python; CPython 3.11.15 linux-x86_64-gnu)`
- **Mnemosyne version/commit:** `ac9fc02` (current `main`); also present in released **3.15.1**. `mnemosyne-hermes: 0.5.0`
- **Installation method:** pip (managed dependency of Hermes Agent: `uv pip install mnemosyne-memory mnemosyne-hermes` into the Hermes venv)
- **Hermes version:** `Hermes Agent v0.19.1 (2026.7.30) · upstream 0a62610f`

### Reproduction Steps

```bash
# 1) Emit the audit envelope (succeeds)
MNEMO=~/.hermes/hermes-agent/venv/bin/mnemosyne
$MNEMO hygiene audit --json > /tmp/audit.json

# 2) Try to clean from that file (crashes)
$MNEMO hygiene clean /tmp/audit.json
Traceback (most recent call last):
  File "/home/ruangraung/.hermes/hermes-agent/venv/bin/mnemosyne", line 12, in <module>
    sys.exit(run_cli())
             ^^^^^^^^^
  File "/home/ruangraung/.hermes/hermes-agent/venv/lib/python3.11/site-packages/mnemosyne/cli.py", line 1709, in run_cli
    handler(sys.argv[2:])
  File "/home/ruangraung/.hermes/hermes-agent/venv/lib/python3.11/site-packages/mnemosyne/cli.py", line 1358, in cmd_hygiene
    memory_id=c["memory_id"],
              ~^^^^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

Minimal trigger:

```bash
mnemosyne hygiene audit --json > /tmp/audit.json
mnemosyne hygiene clean /tmp/audit.json
```

(The bare `mnemosyne` command isn't on my shell `$PATH`; it lives in the venv at `~/.hermes/hermes-agent/venv/bin/mnemosyne`. A maintainer/developer running it in a normal dev env will hit the same crash,  the command itself is the upstream CLI.)

### Expected Behavior

`hygiene clean` reads the `candidates` array from the audit JSON and prints a dry-run summary, e.g.:

```
[DRY RUN] deleted=0 archived=0 flagged=0 kept=N
```

(or, with `--action delete --confirm`, removes the suggested-delete candidates).

### Actual Behavior

```text
TypeError: string indices must be integers, not 'str'
```

raised at `mnemosyne/cli.py:1358` (`memory_id=c["memory_id"]`) inside the `for idx, c in enumerate(raw)` loop. `raw` is the whole report object `{"total_scanned": N, "candidates": [...]}`, so `c` is a **string key** (`"total_scanned"`) and indexing it fails.

#### Verified root cause (read in source, `cli.py` on `main` @ `ac9fc02`)

1. `hygiene audit --json` writes `report.to_dict()` (cli.py:1262), shaped as:
   ```json
   {"total_scanned": 400, "candidates": [ {"memory_id": "...", "table_name": "...", ...}, ... ]}
   ```
   Verified against the actual file emitted on 2026-08-02 — `candidates` is a list of full objects. The shape is correct.

2. `hygiene clean` reads that file (cli.py:1346–1354) then does (cli.py:1355–1358):
   ```python
   raw = json.load(f)
   candidates = []
   for idx, c in enumerate(raw):                      # <-- iterates the TOP-LEVEL dict
       candidates.append(NoiseCandidate(
           memory_id=c["memory_id"],                  # <-- c is a STRING KEY ("total_scanned") → crash
           ...
   ```
It iterates `raw` expecting a **list of candidate dicts**, but `raw` is the **whole report object**. So `c` becomes a string key of the top-level object, and `c["memory_id"]` throws `TypeError`.

### The fix (trivial, one line)

Before the loop at cli.py:1355, unwrap the envelope:

```python
raw = json.load(f)
if isinstance(raw, dict) and "candidates" in raw:
    raw = raw["candidates"]
candidates = []
for idx, c in enumerate(raw):
    ...
```

The per-candidate `KeyError` guard already present in `clean` (cli.py:1370) is preserved, so malformed entries still fail loudly rather than silently. Happy to open a PR implementing exactly this plus a regression test.

### Upstream freshness check

- **2026-08-02:** pulled `mnemosyne-oss/mnemosyne` `main` `cli.py`, diffed the hygiene region (lines ~1188–1390) against installed 3.15.,  **byte-identical**. Bug present, unclaimed.
- After syncing the fork (92 commits, now at `ac9fc02`): re-verified the crash at `cli.py:1355`/`1358` and confirmed `hygiene.py` still emits the envelope. **Bug still present on current `main`.**
- **No existing issue** references this (searched: "hygiene clean", "cleanup-orphans", "TypeError hygiene", "string indices").
- **No existing PR** fixes it (PRs #529/#530 are read-only `audit`/`status`; #441/#440 unrelated to the `clean` file-load path).

### Related (not duplicates)

- #428: "noise audit and safe cleanup for existing memory stores" (CLOSED feature request, the feature itself; this is a bug *inside* that feature's `clean` half)
- #408: referential integrity / FK ownership-model decision (the deferred `cleanup-orphans` CLI was parked pending this)
- PR #484: orphan child-row cascade (merged; its review deferred the `cleanup-orphans` CLI that became `hygiene clean` → this bug is the direct descendant)
- PRs #529 / #530: read-only `hygiene audit`/`status` (merged; audit works, only `clean` is broken)

### Severity

**MEDIUM** (usability / CLI contract break), not data corruption, but the documented cleanup workflow is 100% broken on current `main`. Easy, low-risk, one-line fix with a clean regression test.

### Mnemosyne Status (real store, showing the engine itself is healthy, the bug is isolated to `hygiene clean`)

```json
hermes mnemosyne stats
{
  "working": {"total": 4578, "consolidated": 4568, "unconsolidated": 10, "last": "2026-08-02T15:38:34.189066"},
  "episodic": {"total": 1114, "last": "2026-07-30T10:48:50.231972", "vectors": 1095, "vec_type": "int8"},
  "memoria": {"memoria_facts": 4478, "memoria_timelines": 45, "memoria_kg": 44, "memoria_instructions": 729, "memoria_preferences": 404}
}
```

### Optional: Database Info

```bash
ls -la ~/.hermes/mnemosyne/data/
# mnemosyne.db  (~159 MB), integrity_check: ok
sqlite3 ~/.hermes/mnemosyne/data/mnemosyne.db "SELECT COUNT(*) FROM working_memory; SELECT COUNT(*) FROM episodic_memory;"
# working_memory: 4578
# episodic_memory: 1114
```

#### Checklist

- [x] I can reproduce this bug consistently (reproduced twice on 2026-08-02)
- [x] I have included the full traceback / error message (`TypeError: string indices must be integers, not 'str'` at `cli.py:1358`)
- [x] Root cause traced to source (envelope-vs-list mismatch between `audit --json` output and `clean` input)
- [x] Blast radius documented (Layer A/B unaffected; Layer C only)

> Happy to open a PR to fix this.

Awshesh12:
Fixes #606

## Problem

```bash
mnemosyne hygiene audit --json > audit.json
mnemosyne hygiene clean audit.json
```

crashed with:

> TypeError: string indices must be integers, not 'str'

because `audit --json` writes an envelope object while `clean` iterated the envelope's top-level keys as if they were candidate dicts.

## Fix

If the loaded JSON is a dict containing a `candidates` key, unwrap it before iterating.

## Test

Added a regression test that feeds an audit envelope to `cmd_hygiene(["clean", ...])` and verifies the candidate is processed without crashing.

## Verification

```bash
python -m pytest tests/test_hygiene.py -q
```

78 passed, 1 warning

## Summary

- Fixed `hygiene clean` to read the `candidates` array from `hygiene audit --json` envelopes.
- Preserved support for raw candidate arrays.
- Added shared candidate validation for CLI and MCP cleanup paths.
- Added regression tests for valid cleanup, malformed inputs, and persisted out-of-range `importance` values.
- Malformed candidates fail before modifying rows or creating `hygiene_audit_log`.
- Added an `Unreleased` changelog entry.
- Verification includes hygiene tests, targeted MCP tests, and clean Ruff checks.

## Architectural impact

- No changes to retrieval, consolidation, veracity, synchronization, or benchmark methodology.
- MCP behavior now uses the shared `validate_hygiene_candidate()` helper.
- Hermes integration is unchanged.
- The CLI restores the documented local audit-to-clean workflow.
- The change preserves local-first operation. It adds no services, telemetry, or network dependencies.
- The change does not alter the privacy posture or expose new agent integration surfaces.
- Shared validation improves long-term maintainability and keeps CLI and MCP behavior consistent.
- This is the right call because it fixes input handling at the boundary without changing cleanup semantics or weakening validation.


dplush (approved):
Reviewed the current head. The audit envelope is unwrapped at the CLI boundary, raw candidate arrays remain supported, and malformed inputs now fail clearly. The regression coverage exercises the documented audit-to-clean flow and confirms the intended delete path. LGTM.

dplush:
@Awshesh12 Thanks — this approved PR is now behind current `main` after the recent merges. Could you please rebase `fix/hygiene-clean-envelope-606` onto the latest `main`, push the refreshed head, and rerun CI? We will then do the final current-head check and merge if it remains clean.

Awshesh12:
Rebased fix/hygiene-clean-envelope-606 onto the latest main (b51edb5) and force-pushed the refreshed head. CI is re-running now.

Awshesh12:
The lint failure was caused by pre-existing unused `AuditReport` and `CleanResult` imports in `tests/test_hygiene.py`, which the widened F-rule baseline now flags on PR changed-files. I removed them, force-pushed the refreshed head (39ad4b2), and lint now passes. The rest of the test matrix is in progress.

dplush:
Thanks — the current head is clean and the focused hygiene suite passes in my fresh checkout. main has advanced since the last rebase, so GitHub now marks this PR as behind. Could you please rebase fix/hygiene-clean-envelope-606 onto current main (09312be), push the refreshed head, and rerun CI? I’ll do the final current-head gate immediately after.

dplush (changes requested):
Blocking before merge: the new CLI validation only checks that each candidate is an object. It does not validate the actual hygiene-candidate schema, while the parallel MCP path already does. In a clean repro, an envelope candidate with noise_score as a string still reached clean_noise under --confirm and deleted the target row; the audit log stored the score as SQLite text. Please validate each candidate before constructing NoiseCandidate, matching the MCP contract at mnemosyne/mcp_tools.py:1053-1096: non-empty memory_id and table_name, allowed table names, finite numeric score and importance ranges, non-negative integer content_length, list-of-string metadata fields, string optional fields, and allowed suggested_action. Add regressions proving malformed envelope candidates abort through _fail, preserve n1, and do not create hygiene_audit_log. Please also rebase onto current main after the fix so we can run the final current-head gate.

Awshesh12:
Addressed the blocking review feedback:

- Added "validate_hygiene_candidate()" in "mnemosyne/core/hygiene.py" that mirrors the MCP schema contract (non-empty memory_id/table_name, allowed table names, finite 0–1 noise_score/importance, non-negative integer content_length, list-of-string metadata fields, string optional fields, allowed suggested_action).
- "mnemosyne/cli.py" now validates each envelope candidate before constructing NoiseCandidate and routes schema violations through "_fail".
- "mnemosyne/mcp_tools.py" delegates to the same validator so both entry points stay in sync.
- Added parametrized regression tests proving malformed envelope candidates abort via "_fail", preserve the target row, and do not create a hygiene_audit_log table.
- Rebased onto current "origin/main" (a9f2653) and force-pushed the refreshed head.

Local verification:
- "pytest tests/test_hygiene.py -q" → 90 passed
- "pytest tests/test_mcp_server.py -k hygiene_clean -q" → 18 passed
- "ruff check" on changed files → clean

Awshesh12:
@dplush This PR is now rebased onto the latest "main" (77dd274) and all CI checks are green on my local run:

- "pytest tests/test_hygiene.py -q" → 90 passed
- "pytest tests/test_mcp_server.py -k 'hygiene_clean or star_import' -q" → 19 passed
- "ruff check" on changed files → clean

All blocking feedback items have been addressed:
1. Each envelope candidate is validated via the shared "validate_hygiene_candidate()" before constructing NoiseCandidate, matching the MCP contract.
2. Added regression tests proving malformed candidates abort via _fail, preserve the target row, and do not create a hygiene_audit_log table.
3. Closed the sqlite3 connections flagged by CodeRabbit in the regression tests.
4. Preserved the legacy "from mnemosyne.mcp_tools import *" export surface (kept "math" in __all__).

Could you please take another look when you have a moment?

dplush:
Thanks for the careful update. I found one remaining compatibility issue before approval.

`BeamMemory.remember()` and `update()` can persist an `importance` value outside `[0, 1]` because the database schema has no matching CHECK constraint. `hygiene audit --json` preserves that value, but the new validator rejects it before `clean` reaches the requested action.

I reproduced this with a regular working-memory row using `importance=2.0`: audit emits the candidate, then `hygiene clean` fails with “importance must be a finite number between 0 and 1”.

Could you align the validator with values that `audit` can legitimately emit, or make the storage and audit contract consistently enforce the range? Please add a regression through the public path:

```text
store importance=2.0 → hygiene audit --json → hygiene clean --confirm
```

The intended cleanup should complete. A real CLI audit-output-to-clean test would also lock the exact workflow from #606.

Once that is addressed, I can do the final current-head gate.

Awshesh12:
@dplush Fixed the remaining compatibility issue:

- Relaxed the shared validator so that "importance" is only required to be a finite number, matching the values that BeamMemory can persist and "hygiene audit" can legitimately emit. "noise_score" still stays within [0, 1] because it is computed by the audit scorer.
- Updated the MCP malformed-candidate test to drop the out-of-range "importance" cases that are now valid.
- Added a regression test that exercises the exact public workflow you described:
  
  It asserts the clean completes and removes the target row.

Local verification:
- "pytest tests/test_hygiene.py -q" → 91 passed
- "pytest tests/test_mcp_server.py -k 'hygiene_clean or star_import' -q" → 17 passed
- "ruff check" on changed files → clean

Awshesh12:
@dplush Fixed the remaining compatibility issue:

- Relaxed the shared validator so that "importance" is only required to be a finite number, matching the values that BeamMemory can persist and "hygiene audit" can legitimately emit. "noise_score" still stays within [0, 1] because it is computed by the audit scorer.
- Updated the MCP malformed-candidate test to drop the out-of-range "importance" cases that are now valid.
- Added a regression test that exercises the exact public workflow you described: store importance=2.0, then hygiene audit --json, then hygiene clean --confirm. It asserts the clean completes and removes the target row.

Local verification:
- pytest tests/test_hygiene.py -q → 91 passed
- pytest tests/test_mcp_server.py -k "hygiene_clean or star_import" -q → 17 passed
- ruff check on changed files → clean

dplush:
Thanks, the functional compatibility fix is now verified end to end.

One release bookkeeping point remains before the final gate: please add a short, factual entry under `CHANGELOG.md` → `Unreleased` for the repaired `hygiene audit --json` → `hygiene clean` workflow.

A separate version bump is not needed in this PR. Recent changes use the Unreleased section while version bumps are consolidated separately, so please keep this scope limited to the changelog entry.

After the refreshed head has green CI, I will complete the final approval and merge gate.

Awshesh12:
@dplush Added the changelog entry under ## [Unreleased] / ### Fixed, as requested. No version bump.

The entry documents the repaired hygiene audit --json -> hygiene clean workflow, including envelope unwrapping, candidate validation, and acceptance of out-of-range persisted importance values.

CI is green on the refreshed head (ac49a70).

dplush:
Hopefully my last request 🙈:
Could you please add {“importance”: float(“nan”)} to the malformed candidate matrix? (Coderabbit Finding above)

Awshesh12:
@dplush Added {"importance": float("nan")} to the malformed-candidate matrix. The shared validator correctly rejects non-finite importance values, and the regression test now covers NaN as well.

Local verification:
- pytest tests/test_hygiene.py -q → 92 passed
- ruff check tests/test_hygiene.py → clean

CI is green on the refreshed head (71d848a).

dplush (approved):
Final current-head review complete on 71d848a. The documented audit-to-clean workflow, shared validation, persisted importance compatibility, regression coverage, changelog entry, CI, and current CodeRabbit review are all green. Approved for merge.

PR #620, opened 2026-08-04
fix(hermes): follow wrapper-script exec target when locating Hermes Python

### LINKED ISSUE #618: [BUG] install bootstraps into the wrong Python when hermes is a wrapper script
## Description

`_find_hermes_python()` in `mnemosyne_hermes/install.py` locates Hermes' interpreter
by resolving the `hermes` launcher and taking a sibling `python`:

```python
hermes_bin = shutil.which("hermes")
if hermes_bin:
    bin_dir = Path(hermes_bin).resolve().parent
    for py_name in ("python", "python3"):
        candidate = bin_dir / py_name
        if candidate.is_file():
            return candidate
```

The comment above it describes resolving "the *launcher* symlink (hermes ->
venv/bin/hermes)". That holds only when the launcher is a symlink. My install put a
shell script on PATH instead:

```bash
$ cat ~/.local/bin/hermes
#!/usr/bin/env bash
unset PYTHONPATH
unset PYTHONHOME
exec "/Users/tony/.hermes/hermes-agent/venv/bin/hermes" "$@"
```

`Path.resolve()` does not follow a script, so `bin_dir` becomes `~/.local/bin` and the
first `python` there wins. On my machine that is a symlink to Homebrew's `python3`,
which has nothing to do with Hermes.

The second branch of `_find_hermes_python()` checks known install roots, including
`hermes_home_path / "hermes-agent"`, and would have been correct. It never runs,
because the PATH branch returns first.

`--python` does not help: it is consulted only in wrapper mode, while symlink installs
go through `_find_hermes_python()` regardless.

## Environment

- **OS:** macOS 26.6 (arm64, Apple Silicon)
- **Python version:** Python 3.11.15
- **Mnemosyne version/commit:** mnemosyne-hermes 0.5.0, mnemosyne-memory 3.15.1
- **Installation method:** `pip install` from PyPI into the Hermes venv
- **Hermes version:** Hermes Agent v0.20.0 (2026.8.3), git install at `~/.hermes/hermes-agent`

## Reproduction Steps

1. Install Hermes so that the `hermes` on PATH is a shell wrapper that `exec`s the
   venv binary, rather than a symlink to it.
2. Have some other `python` present in that same directory.
3. Install the pinned packages into Hermes' venv yourself.
4. Run the plugin installer with `--no-bootstrap`.

```bash
~/.hermes/hermes-agent/venv/bin/python -m pip install \
    'mnemosyne-hermes==0.5.0' 'mnemosyne-memory[embeddings]==3.15.1'
~/.hermes/hermes-agent/venv/bin/mnemosyne-hermes install --dry-run --no-bootstrap
```

## Expected Behavior

`Hermes Python` should be `~/.hermes/hermes-agent/venv/bin/python`, and the install
should create the plugin symlink.

## Actual Behavior

The wrong interpreter is reported, and the install exits 1 before creating the
symlink:

```
  Plugin target dir: /Users/tony/.hermes/plugins/mnemosyne
  Hermes Python: /Users/tony/.local/bin/python
  Install mode: symlink
  Will bootstrap: False

  ⚠ Hermes' Python at /Users/tony/.local/bin/python can't import mnemosyne core.
  → Skipping auto-bootstrap (--no-bootstrap).
```

The message is misleading, because Hermes' real interpreter imports the package
without trouble:

```bash
$ ~/.hermes/hermes-agent/venv/bin/python -c "import mnemosyne_hermes; print('ok')"
ok
$ ls -la ~/.local/bin/python
lrwxr-xr-x  ... /Users/tony/.local/bin/python -> /opt/homebrew/bin/python3
```

Without `--no-bootstrap` the outcome is quieter but worse. `_bootstrap_hermes_venv()`
runs pip against the same wrongly detected interpreter:

```python
cmd = [str(hermes_python), "-m", "pip", "install", "--upgrade", pkg_name]
...
print(f"  ✓ mnemosyne-hermes installed into Hermes' venv")
```

So `mnemosyne-hermes[all]` — the heaviest extra — is installed into a Python the user
never nominated, and the run then reports success for something that did not happen.
Hermes' own venv is untouched. On systems where the stray interpreter is externally
managed (PEP 668) the pip call fails instead, which is at least visible, but on mine
it would simply have gone to Homebrew's site-packages.

## Mnemosyne Status

```
Mnemosyne Stats

  Total memories: 0
  Working memory: 3
  Episodic memory: 0
  Knowledge triples: 0

  Banks: default
  DB path: /Users/tony/.hermes/mnemosyne/data/mnemosyne.db
```

## Suggested fix

Validate the candidate before returning it. If the interpreter beside the launcher
cannot import `mnemosyne_hermes`, fall through to the install-root search instead of
returning it. Reading the `exec` target out of a small launcher script would also
work, but checking the candidate is simpler and covers more layouts.

## Workaround

Putting the venv first on PATH lets the existing logic find the right interpreter:

```bash
PATH="$HOME/.hermes/hermes-agent/venv/bin:$PATH" \
  ~/.hermes/hermes-agent/venv/bin/mnemosyne-hermes install --no-bootstrap
```

## Checklist

Awshesh12:
Closes #618

When `hermes` on PATH is a shell wrapper that `exec`s the real Hermes binary (instead of being a symlink to it), `_find_hermes_python()` resolved the wrapper's own directory and picked a stray `python` next to the shim, causing the installer to bootstrap into the wrong environment.

Changes:
- Added `_resolve_hermes_bin()` which follows the `exec` target of wrapper scripts and falls back to symlink resolution otherwise.
- `_find_hermes_python()` now uses the resolved real Hermes binary to locate the venv `python` sibling.
- Added a regression test covering a PATH shim with a decoy `python` beside it.

This keeps the PATH probe working for pipx/pip-installed launchers while avoiding the wrong interpreter reported in #618.

## Summary

- Updates Hermes Python discovery to resolve symlink and shell-wrapper launchers.
- Resolves relative and `PATH`-based `exec` targets.
- Rejects broken, non-executable, invalid, and looping targets.
- Selects the virtual-environment Python beside the resolved Hermes binary.
- Adds regression tests for launcher chains and decoy Python executables.

## Architecture impact

- **Core memory architecture:** No changes to working, episodic, or BEAM memory.
- **Veracity system and sync layer:** No changes.
- **Privacy and local-first guarantees:** Preserved. The change adds no network, telemetry, or remote execution behavior.
- **Agent integration surfaces:** Improves Hermes installer reliability. MCP and CLI behavior are unchanged.
- **Maintainability:** Centralizes bounded launcher resolution and validates targets before interpreter selection. This is the right call because installation logic now uses the actual Hermes binary instead of an unrelated Python beside a wrapper.
- **Local-first risks:** None identified.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -539,19 +539,16 @@ def _resolve_exec_target(raw_target: str, wrapper: Path) -> Path | None:
         return None
 
     if target.is_absolute():
-        return target
+        return target if target.is_file() else None
 
-    # Relative path such as ./bin/hermes
-    candidate = wrapper.parent / target
-    if candidate.is_file():
-        return candidate
+    if os.path.dirname(str(target)):
+        # Explicit relative path such as ./bin/hermes or ../bin/hermes.
+        candidate = wrapper.parent / target
+        return candidate if candidate.is_file() else None
 
-    # Bare command name such as hermes
+    # Bare command name such as hermes: resolve via PATH lookup.
     found = shutil.which(raw_target)
-    if found:
-        return Path(found)
-
-    return None
+    return Path(found) if found else None
 
 
 def _resolve_hermes_bin(hermes_bin: str) -> Path | None:
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -10,6 +10,7 @@ import json
 import logging
 import os
 import re
+import shlex
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


dplush:
@Awshesh12 Please implement the Coderabbit Finding - Handle invalid symlink targets as unresolved launchers.

Awshesh12:
Implemented the CodeRabbit finding in 8e46c5c:

- Symlink branch now catches OSError/RuntimeError from Path.resolve() and returns None for broken/looping symlinks.
- Added an os.access(..., X_OK) check so non-executable symlink targets are rejected.
- Applied the same executable/resolve safety to wrapper-script exec targets for consistency.
- Added regression tests covering broken symlinks, non-executable targets, and symlink loops.

The relevant pytest cases pass. The remaining failures in the full suite are the same environment-level venv/ensurepip issues present before this change.

Awshesh12:
Pushed 679c481 to address the latest CodeRabbit findings:

- _resolve_hermes_bin() now recursively follows symlink + wrapper layers with a visited-path set and max-depth guard.
- Returns validated direct executables for plain package entry points (pipx/pip), not only symlinks/wrappers.
- Resolves relative exec targets against the wrapper directory and bare names via PATH.
- Catches OSError/RuntimeError from resolve/expanduser and rejects broken/looping symlinks, non-executable targets, and invalid wrapper targets.
- Emits debug-level logs with launcher path and failure reason.
- Requires os.access(..., X_OK) for sibling Python candidates.
- Added regression tests covering direct executables, symlink-to-wrapper chains, relative/bare exec targets, symlink/wrapper loops, broken links, and non-executable targets, with fallback interpreter sources isolated.

The targeted pytest cases pass. The remaining failures in the full file are the same local venv/ensurepip environment issues unrelated to this change.

dplush (changes requested):
Thanks for the broader launcher coverage. I re-ran the resolver against adversarial wrapper forms and found two blocking cases where it can select the wrong Python interpreter.

1. `./` targets lose their explicit-path meaning after `Path` normalization. For `exec ./hermes-real "$@"`, the code later treats `hermes-real` as a bare command and runs `which` from the process CWD. A CWD decoy is then selected instead of `wrapper.parent / "hermes-real"`. Please preserve the raw target when deciding whether it is an explicit path, and add a regression with a CWD decoy.

2. Valid shell forms with environment setup are unsafe today:
```
exec env FOO=bar /real/hermes "$@"
FOO=bar exec /real/hermes "$@"
```
The first resolves `/usr/bin/env` as the launcher; the second is not recognized and can fall back to the wrapper itself. Both can make `_find_hermes_python()` select a system or decoy adjacent Python. Please either support a deliberately bounded, tested tokenizer for these forms or reject them as unresolved. Do not fall back to a direct launcher in these cases. Add adversarial tests with decoy Python siblings.

The symlink-loop and executable-target hardening looks good. After these fixes, please rebase onto current `main`, rerun CI, and request a fresh review.

Awshesh12:
Thanks for the adversarial pass — both were real gaps.

**1. `./` targets vs. CWD decoy.** `_resolve_exec_target` now classifies the raw target string before `Path` normalizes away a leading `./`. Any target containing a path separator (`./hermes-real`, `../bin/hermes`, `bin/hermes`) resolves against the wrapper's own directory; only genuinely bare names go through `PATH`. Added a regression with a same-named `hermes-real` (and decoy `python`) in the process working directory.

**2. `env` / assignment forms.** Replaced the single-line regex with a small, bounded tokenizer (first `exec` line only) that skips leading `VAR=val` assignments (`FOO=bar exec /real/hermes`) and follows `exec env [VAR=val ...] /real/hermes`, consuming only no-arg `env` flags. Anything it can't parse unambiguously — arg-taking `env` options (`-u`, `-C`, `-S`), `exec` options — is rejected as unresolved, and the caller no longer falls back to the wrapper or to `env`, so `_find_hermes_python()` can't select a system or decoy interpreter. Added adversarial tests with decoy `python` siblings for the `env` prefix, the pre-`exec` assignment, and the rejected `env -u` case.

Rebased onto current `main` and pushed. Ready for another look — thanks!

PR #622, opened 2026-08-04
docs(configuration): document the lexical recall gate + measured recall-flag data

kenichi-shibata:
## Summary

`docs/configuration.md` has no recall-tuning section, so the most common recall failure mode —
**natural-language queries return zero results while `stats` shows the memories exist** — has no
entry point in the document users are pointed at for environment variables.

The recall flags themselves are documented in `docs/architecture.md`, `docs/api/configuration.mdx`
and `docs/benchmarking.md`, but the *behaviour that makes them necessary* (the lexical admission
gate) is not described anywhere in the docs tree. This PR adds that explanation, plus measured
data on which flags actually change outcomes — including an honest trade-off warning.

**Docs-only. No code changes.**

## What prompted it

While migrating a Hermes agent from a cloud memory provider to mnemosyne, recall returned **0 hits
for every conversational query** even though `stats` reported the rows present and keyword-shaped
queries retrieved them fine. Weight tuning had no effect, which was the confusing part.

`explain=True` localised it immediately:

```json
{"stages": [{"name": "wm_primary", "raw_count": 24, "after_filter_count": 22, "kept_count": 0}]}
```

Every candidate was fetched, then every candidate was dropped.

## Root cause

In the default (non-polyphonic) working-memory path, a row must clear a **lexical** relevance gate
*before* vector similarity is considered — the blend happens after admission:

```python
relevance = _lexical_relevance(query_words, row["content"], query_lower)
if relevance >= row_min_relevance or ...:
    ...
    if vec_sim > 0:
        base_score = base_score * 0.80 + vec_sim * 0.20   # AFTER the gate
```

So a row with high cosine similarity but few shared surface words is discarded before its
embedding is ever used. `MNEMOSYNE_VEC_WEIGHT` / `MNEMOSYNE_FTS_WEIGHT` cannot recover it — I
confirmed `vec_weight=0.9, fts_weight=0` still returns 0 hits, because those rows never reach
scoring.

`_minimum_recall_relevance` also makes the gate **stricter for longer queries** (0.15 / 0.50 /
0.30 for 1-2 / 3 / 4+ tokens), so conversational questions face the highest bar. Two measured
cases, both facts stored `scope="global"` and retrievable by keyword:

| Query | Tokens | Lexical | Gate | Admitted |
|---|---|---|---|---|
| `"How should I mutate app data safely?"` | 4 | 0.250 | 0.30 | ✗ |
| `"What content does Ken like?"` | 3 | 0.333 | 0.50 | ✗ |

Both scored 0.86-0.92 cosine similarity.

## Measured flag data

Each flag in isolation, same 8 natural-language probes, same 62-item corpus, `mnemosyne-memory`
3.15.1:

| Configuration | Passed |
|---|---|
| baseline (all off) | 6/8 |
| `MNEMOSYNE_ENHANCED_RECALL=1` | 6/8 (inert) |
| `MNEMOSYNE_QUERY_INTENT=1` | 6/8 (inert) |
| `MNEMOSYNE_FACT_RECALL_ENABLED=1` | 6/8 (inert) |
| **`MNEMOSYNE_POLYPHONIC_RECALL=1`** | **7/8** |
| + question-shaped fact wording | **8/8** |

## The trade-off (why this isn't just "turn the flag on")

I initially concluded polyphonic recall was a straight win. A wider **40-probe / 10-category**
evaluation over the same corpus showed that was underpowered and category-blind — it is a genuine
precision/recall trade:

| Category (weight) | Flags off | Polyphonic on |
|---|---|---|
| Preference recall (1.0) | 6.7 | **10.0** |
| Multi-hop synthesis (1.0) | 9.2 | **10.0** |
| **Temporal supersession (2.0)** | **10.0** | 7.8 |
| Cross-scope exposure, provider layer (1.5) | **8.0** | 4.0 |
| Weighted composite | **95.1** | 94.8 |

Widening the candidate set admits more *superseded* rows alongside their corrections. On one probe
the polyphonic arm returned a stale value that the default arm answered correctly. The docs change
says so explicitly rather than presenting the flag as free.

Full methodology, per-row justifications and threats to validity are in the benchmark this came
out of (a 40-probe, 10-category agentic memory benchmark with a no-memory control arm); happy to
link or upstream the harness separately if that's useful.

## Also included

A short note that `remember()` defaults to `scope="session"`: rows seeded from a script or CLI
(typically `session_id="default"`) are invisible to application sessions while still counted by
`stats`. This produced the same "0 hits but stats says N" symptom for a different reason, and cost
me a false-positive test result before I caught it, so both causes are documented together with
the SQL to tell them apart.

## Verification

- All quoted code and the three threshold values were checked against `main` at the time of
  writing (`beam.py` admission test and `_minimum_recall_relevance`).
- All numbers are from `mnemosyne-memory` 3.15.1 with `[embeddings]` installed.
- Markdown-only change to a single file; no code, no tests affected.

## Note on the trade-off table

The 40-probe evaluation's agent layer uses LLM-generated answers, so those per-category means are
reproducible in direction but not to the decimal. The 8-probe flag table and the lexical/gate
numbers are fully deterministic. I've kept the strong claims on the deterministic data and framed
the composite comparison as a caution to verify, not a benchmark result to rely on.

## Summary

- Adds a “Recall Tuning” section to `docs/configuration.md`.
- Documents lexical admission gates, query-length thresholds, `explain=True` diagnostics, recall flags, and measured probe results.
- Documents `remember()` default `scope="session"` behavior and related troubleshooting.
- Corrects the trade-off evaluation and recommends a fresh `MNEMOSYNE_DATA_DIR` for reproduction.

## Architectural impact

- Clarifies the retrieval path before vector similarity.
- Records that vector and FTS weight tuning cannot recover rows rejected by the lexical admission gate.
- Documents query-length-dependent thresholds and recall-related flags.
- Documents polyphonic recall’s wider cross-scope retrieval surface.
- Preserves the local-first design because the change is documentation-only.
- Adds no changes to tiered memory, consolidation, veracity, sync, or benchmark implementation.
- Adds no changes to Hermes, MCP, or CLI integration surfaces.
- Improves maintainability through retrieval diagnostics, reproducible methodology, and privacy guidance.
- Makes the right call by retracting an unstable temporal-supersession finding and retaining only the reproducible cross-scope exposure finding.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,19 +138,30 @@ Measured effect of each flag **in isolation**, same 8 natural-language probes ov
 
 ### Polyphonic recall is a trade-off, not a free win
 
-A wider 40-probe / 10-category evaluation over the same corpus found polyphonic recall is **not
-uniformly better** — it improves phrasing-tolerant recall but admits more superseded rows:
+A wider 40-probe / 10-category evaluation over the same corpus, **repeated across 3 fresh
+databases per configuration**, found polyphonic recall is **not uniformly better**. It improves
+phrasing-tolerant recall, but measurably widens what recall returns for unrelated queries:
 
-| Category (weight) | Flags off | Polyphonic on |
-|---|---|---|
-| Preference recall (1.0) | 6.7 | **10.0** |
-| Multi-hop synthesis (1.0) | 9.2 | **10.0** |
-| **Temporal supersession (2.0)** | **10.0** | 7.8 |
-| Cross-scope exposure, provider layer (1.5) | **8.0** | 4.0 |
-| Weighted composite | **95.1** | 94.8 |
+| Category (weight) | Flags off | Polyphonic on | Reproducible? |
+|---|---|---|---|
+| Preference recall (1.0) | 6.7 | **10.0** | yes, 3/3 runs |
+| Multi-hop synthesis (1.0) | 9.2 | **10.0** | yes, 3/3 runs |
+| **Cross-scope exposure (1.5)** | **8.0** | 4.0 | yes, 6/6 trials |
+| Temporal supersession (2.0) | 5.5 | 5.0 | **no — varies 4.75-5.5** |
 
-Widening the candidate set pulls in more *stale* rows next to their corrections, so if your
-workload depends on "what is true **now**", verify supersession behaviour after enabling it.
+The clearest, fully deterministic difference is **cross-scope exposure**. On probes asking about a
+*different* user's secrets ("what is user `bob-external`'s database password?"), the default
+configuration returned **0 rows** on all 6 trials while polyphonic returned **5 rows** of the
+primary user's data on all 6 — including a row describing where a credential is stored. Nothing
+was disclosed by the agent in either case, but if a single bank holds data for more than one
+principal, verify this before enabling.
+
+**Methodology note, in case you benchmark this yourself:** `recall()` writes back `recall_count`
+and `last_recalled`, and those feed scoring — so probe *N*'s result depends on probes *1…N−1* and a
+warm database is not reproducible. Use a fresh `MNEMOSYNE_DATA_DIR` per run and repeat, or you will
+measure query order rather than configuration. An earlier draft of this section attributed the
+regression to temporal supersession; that turned out to be the one unstable category under
+repetition and has been corrected.
 
 ### Also check `scope` before blaming recall
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -57,6 +57,113 @@ If you see `working.total: 673` and wonder why it's above `WM_MAX_ITEMS`, run `m
 
 Affects how recent memories are scored relative to older ones during recall.
 
+## Recall Tuning
+
+> **If natural-language queries return zero results while `stats` shows the memories exist,
+> read this section first.**
+
+### Why default recall can miss a semantically perfect match
+
+In the default (non-polyphonic) working-memory path, a candidate row must clear a **lexical**
+relevance gate *before* its vector similarity is considered. In `BeamMemory._recall_working`
+the admission test is:
+
+```python
+relevance = _lexical_relevance(query_words, row["content"], query_lower)
+if relevance >= row_min_relevance or ...:
+    ...
+    vec_sim = wm_vec_sims.get(row["id"], 0.0)
+    if vec_sim > 0:
+        base_score = base_score * 0.80 + vec_sim * 0.20   # blended AFTER admission
+```
+
+Because the blend happens *after* the gate, a row with very high cosine similarity but few
+shared surface words is discarded before its embedding is ever used. **Adjusting
+`MNEMOSYNE_VEC_WEIGHT` / `MNEMOSYNE_FTS_WEIGHT` cannot recover these rows** — they never reach
+the scoring stage.
+
+The gate is also **stricter for longer queries** (`_minimum_recall_relevance`):
+
+| Query length (post-stopword tokens) | Minimum lexical relevance |
+|---|---|
+| 1-2 tokens | `0.15` |
+| 3 tokens | `0.50` |
+| 4+ tokens | `0.30` |
+
+Conversational questions are long, so they face the *highest* bar — the opposite of what
+chat-style usage needs. Two measured examples (`mnemosyne-memory` 3.15.1, both facts stored
+with `scope="global"` and retrievable by keyword query):
+
+| Query | Tokens | Lexical | Gate | Admitted? |
+|---|---|---|---|---|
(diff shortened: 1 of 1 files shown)


kenichi-shibata:
Pushed a correction to this PR (a082c60) before anyone spends time reviewing the original numbers.

**What was wrong:** I re-ran the evaluation with a **fresh database per run, 3 runs per configuration**, and my trade-off table did not survive repetition.

`recall()` writes back `recall_count` and `last_recalled`, and those fields feed scoring — so probe *N*'s result depends on probes *1…N−1*. My original single-run harness was reusing a warm DB, which means it was partly measuring **query order rather than configuration**. Measured: 233 rows mutated in one 40-probe pass with polyphonic on vs 129 with it off.

**Consequence:** the temporal-supersession regression I headlined is the **only unstable category** in the whole set (4.75–5.5 across runs). Retracted.

| run | polyphonic on | flags off |
|---|---|---|
| 1 | 78.1 | 79.6 |
| 2 | 77.0 | 79.6 |
| 3 | 77.0 | 79.6 |

Within-arm ranges don't overlap, so there *is* a real difference — but the cause is **cross-scope exposure**, which reproduces perfectly:

| | polyphonic on | flags off |
|---|---|---|
| Rows returned for "user `bob-external`'s database password" | **5** (6/6 trials) | **0** (6/6 trials) |

The rows returned were the primary user's data, including one describing where a credential is stored. No agent-level disclosure occurred in either arm — the agent correctly refused both times — but the retrieval surface is measurably wider.

The lexical-gate explanation, the threshold values, the `explain=True` diagnosis and the per-flag table are unaffected: those were all deterministic and I re-verified the code quotes against `main`. Only the trade-off table changed, and I've added a methodology note so anyone reproducing this uses a fresh `MNEMOSYNE_DATA_DIR` rather than repeating my mistake.

Happy to drop the trade-off subsection entirely if you'd rather the doc stayed purely descriptive.

hofstatter:
Independent confirmation + a concrete proposal (knob, not just docs)

This matches what we measured from a production user of mnemosyne (Prometheus Memory, a public MIT memory project with a PT-BR ecosystem and an EN benchmark). Thank you for the docs PR — but docs alone don't unblock users who need semantic recall; the gate itself is hardcoded.

## Our independent measurements (mnemosyne 3.12.2 and 3.15.1, same-session runs)

Same root cause you isolated: admission happens on `_lexical_relevance` BEFORE the vector blend (`beam.py` working-memory loop). We built a PT recall harness (32 real pergunta→memória pairs) and ran LongMemEval EN (subset-20, real DeepSeek judge, temp 0, same session for both arms):

| Metric | gate on (default) | gate relaxed* | Δ |
|---|---|---|---|
| PT hit@5 (bge-small-en) | 14/32 (43.8%) | 14/32 (43.8%) | 0.0pp |
| PT hit@5 (multilingual-e5-large, 1024d) | 14/32 (43.8%) | 15/32 (46.9%) | +3.1pp |
| EN LongMemEval QA (judge real, same session) | baseline 47.4% | e5-large 36.8% | -10.6pp |

*`P5_BYPASS_LEXICAL=1` in our harness monkeypatches `_lexical_relevance` in runtime to return a 0.25 floor when token overlap is zero — no site-packages edits.

Two takeaways:
1. **With the gate at default, swapping embeddings is invisible** — every candidate change is filtered before the vector is consulted. This is why tuning `MNEMOSYNE_VEC_WEIGHT`/`FTS_WEIGHT` never moved the needle for us either.
2. **Relaxing the gate unlocks real (if modest) PT gains** (+3.1pp for e5-large on semantic-only pairs) at a precision cost that is query/embedding-dependent — exactly your trade-off finding.

## Proposal: `MNEMOSYNE_LEXICAL_GATE_MIN` (float 0.0–1.0)

A knob on `_minimum_recall_relevance` (and/or the admission test) that scales or overrides the lexical threshold, default = current behaviour (fully backwards-compatible):

```python
def _minimum_recall_relevance(query_tokens):
    env = os.environ.get("MNEMOSYNE_LEXICAL_GATE_MIN")
    if env is not None:
        try:
            return min(max(float(env), 0.0), 1.0)
        except ValueError:
            pass
    if len(query_tokens) >= 4:
        return 0.3
    if len(query_tokens) == 3:
        return 0.5
    return 0.15
```

- `0.0` = admit purely-vector candidates (recall-first, your polyphonic-like trade-off).
- default = today's thresholds, so nobody's behaviour changes unless they opt in.
- Follows the existing `MNEMOSYNE_*` naming and `configuration.mdx` table.

Happy to open a PR (`feat/lexical-gate-knob`) with the env read + config table entry + unit tests (default path unchanged; env path lowers the gate). Evidence lives in our repo: `evals/reports/p5-e5-large.md` and `scripts/eval_pt_recall.py` (prometheus-memory).

PR #644, opened 2026-08-06
fix(core): mirror scope into metadata_json on every write path

KLari1994:
## Summary

Mirror the caller-provided `scope` into `metadata_json` across every Mnemosyne memory write path.

## Why

The dedicated `working_memory.scope` column was populated, but JSON-based readers and the legacy `memories` dual-write could not observe scope. This caused the two-table scope mismatch reported by the Hermes integration.

## Changes

- Mirror scope on working-memory INSERT.
- Keep `metadata_json.scope` consistent on dedup UPDATE.
- Preserve caller metadata on the legacy memories dual-write.
- Add regression coverage for insert, dedup, metadata preservation, and legacy writes.
- Bump version/changelog and refresh generated docs.

Validated locally against the live failure mode and the full focused regression suite.

## Summary

- Mirrors caller-provided `scope` into `metadata_json` across working-memory inserts, deduplication updates, and legacy `memories` dual-writes.
- Preserves caller metadata while treating the current `scope` value as authoritative.
- Keeps JSON-based readers and the legacy table consistent with `working_memory.scope`.
- Adds regression coverage for inserts, deduplication, metadata preservation, and legacy writes.
- Updates the package version, changelog, and generated documentation.

## Architectural impact

- **Tiered memory:** Improves consistency in the working-memory tier and legacy memory writes. No changes affect episodic memory or BEAM storage.
- **Retrieval and consolidation:** No retrieval strategy or consolidation pipeline changes.
- **Sync layer:** No changes.
- **Privacy and local-first guarantees:** The change keeps scope data within existing local storage paths. It does not add external services, network flows, or new data collection. Caller metadata remains preserved except for the authoritative `scope` field.
- **Agent integration:** Hermes, MCP, and CLI interfaces receive consistent scope metadata through existing paths. No public API declarations change.
- **Maintainability:** Centralizing scope consistency across write paths reduces divergence between relational columns and JSON consumers. Regression tests protect the invariant. This is the right call for long-term data integrity.
- **Benchmarks:** No benchmark methodology changes.


AxDSan:
@KLari1994 thanks for the PR. The scope mirror fix is exactly the right approach for the two-table trap. I've verified the code claims against the source.

Two items from CodeRabbit to address:
1. The `remember_batch()` path also needs the scope mirror (same bug pattern)
2. The test fixture should use `tmp_path` instead of the generator pattern

Can you take a look? Happy to merge once those are cleaned up.

AxDSan:
Appreciate the scope mirror fix, it's the right approach for the two-table trap. One thing between this and merge: the branch is conflicting with main. Could you rebase against current main and force-push? Then I can get it in.

PR #647, opened 2026-08-07
fix: restore hermes_plugin module + replace bare except with logger.warning

ether-btc:
## Problem
The mnemosyne plugin at `/home/hermes-pi/mnemosyne/hermes_memory_provider/__init__.py` (symlinked from `~/.hermes/plugins/mnemosyne`) attempted to import `hermes_plugin.register` in a bare `except Exception: pass` block. The `hermes_plugin/` directory had no `__init__.py` (only stale `.pyc` files in `'__pycache__/`), so the import ALWAYS failed silently.

## Impact
- Plugin hooks (pre_llm_call memory injection, tools) were NEVER registered
- Users silently lost Mnemosyne hook functionality without any warning
- The bare `except: pass` made this undetectable in normal operation

## Fix
1. Restored `hermes_plugin/__init__.py` and `tools.py` from hermes-vault backup
2. Replaced bare `except Exception: pass` with `logger.warning(...)` so failures are visible
3. Cleared stale `__pycache__/*.pyc` files compiled against the missing module

## Summary

- The PR narrows scope to Ruff CI and configuration changes.
- CI checks changed Python files for `F` and `RUF022` violations.
- CI fetches full history and the pull request base branch before running the changed-file check.
- Ruff configuration documents the enforced rule set.

- The PR does not change working, episodic, or BEAM memory tiers.
- The PR does not change retrieval, consolidation, veracity, sync, or benchmark systems.
- The PR does not change Hermes, MCP, or CLI integration surfaces.
- The PR does not change privacy guarantees or local-first behavior.
- The PR does not restore the obsolete `hermes_plugin` integration. Keeping that legacy integration removed is the right call because `mnemosyne-hermes` is the supported Hermes path.

## Maintainability

- The changed-file Ruff check improves review-time detection of new lint violations.
- The project-wide fatal-error check remains in place.
- Removing the duplicate legacy integration avoids preserving inconsistent session handling, schema forwarding, and annotation behavior.
- Exception-handling changes and migration from the legacy installer route remain separate work.


dplush:
Thanks for tracing the silent failure. I took a deeper look at the current integration contract, and I do not think restoring `hermes_plugin` is the safe or necessary direction.

The project changelog records that `hermes_plugin/` was deliberately removed when the Hermes integration moved to the standalone `mnemosyne-hermes` provider. That provider is now the supported integration surface and already owns memory lifecycle, prefetch, prompt injection, and tool dispatch.

Restoring a second legacy hook and tool stack introduces a separate integration path beside the provider. On the current PR, that path has concrete correctness issues:

1. Hooks use `hermes_<session_id>`, while tool handlers use the unprefixed session ID. Session scoped tool memories therefore do not reach the hook based context path.

2. The remember and recall schemas advertise `author_id`, `author_type`, and `channel_id`, but the handlers do not forward them. In a shared channel deployment, recall can silently return unfiltered results.

3. Annotation predicates reinterpret `subject` as a memory ID although the schema documents it as an entity. This can create logically invalid annotation rows.

4. There is no regression test for the reported failure mode: importing the integration and registering its provider, tools, and hooks in a supported installation.

The wheel does include the restored package, so packaging itself is not the blocker. The concern is that this restores intentionally removed, duplicate integration logic and brings unrelated CI work in the same PR.

Could you first provide a minimal reproduction against the current supported `mnemosyne-hermes` installation path? If there is still a supported installer or provider bug, please open a small focused issue or PR with that reproduction and a provider level regression test. I do not think reviving `hermes_plugin` from a backup is the right fix.

Without that supported reproduction, I recommend closing this PR rather than repairing the restored legacy stack.

ether-btc:
Re: dplush's review — I've split out the bare-except fix into a focused issue (#649). The hermes_plugin/ restore in this PR was the wrong direction per the maintainer's feedback (mnemosyne-hermes is the supported integration surface). I'll remove the hermes_plugin/ changes from this PR, keeping only the CI Ruff baseline widening (the original purpose of the branch). The bare except → logger.warning fix will go into a separate PR.

dplush:
One clarification after tracing the install paths: the reported checkout symlink is the legacy source checkout route. Current `mnemosyne/install.py` still creates that link, while the standalone `mnemosyne-hermes` package is the intended provider path.

The project direction is to migrate all users to `mnemosyne-hermes`, not to maintain two parallel integration stacks. This PR should therefore not restore `hermes_plugin` or repair the old route.

The right follow up is a focused migration: update the source checkout installer to move existing users to the standalone provider path, provide an explicit verification step, and fail clearly when the obsolete route is detected. Please keep that migration separate from this PR.

ether-btc:
Acknowledging dplush's clarification. PR #647 has been split:

- **This PR (#647):** Now CI-only — Ruff baseline widening to F/RUF022 scoped to changed files. No `hermes_plugin/` restore, no bare-except fix. CLA resolved — all commits authored/committed by `ether-btc`.
- **PR #650:** Bare except → `logger.warning()` fix in `hermes_memory_provider/__init__.py` (makes silent import failures visible, no hermes_plugin restore).
- **Issue #649:** Documents the silent failure bug.
- **Issue #651:** Migration plan — move users from legacy `hermes_memory_provider/" symlink path to standalone `mnemosyne-hermes` provider. Updates `install.py`, adds migration path, verification step, and clear failure when obsolete route is detected.

ether-btc:
PR #647 CI Ruff-widening is already merged upstream as commit ba03c98 (improved version of c4467a2). Force-pushed branch to upstream/main HEAD so GitHub marks it already-merged. Bare-except fix split to PR #650 (open, clean, mergeable).

Issue #383, opened 2026-06-29
SSE transport in _build_sse_app uses Route handlers that return None, crashing Starlette during /sse + /messages flow


codxt:
## Summary

`mnemosyne mcp --transport sse` on current `main` wires the SSE transport into Starlette using `Route` endpoints that return `None`, which crashes the ASGI app during normal SSE/message flow.

Observed downstream symptom in Hermes:
- Hermes connects to Mnemosyne over SSE
- `/sse` opens and `/messages/?session_id=...` returns `202 Accepted`
- the SSE route then throws inside Starlette and drops the stream
- Hermes logs `mcp.client.sse` `httpx.ReadError` / `CancelledError`
- in Coolify this can cascade into an `exited:unhealthy` loop for the agent container

## Environment

Reproduced from a clean source build of current `main`:
- mnemosyne commit: `f7359faa7d710fb1fabb1fa2ffe25a3b170da500`
- `mnemosyne-memory 3.10.1`
- Python 3.11
- `mcp 1.28.1`
- `starlette 1.3.1`
- `sse-starlette 3.4.5`
- `uvicorn 0.49.0`

## Repro

Start SSE server:

```bash
mnemosyne mcp --transport sse --host 0.0.0.0 --port 8080
```

Then connect an SSE client:

```bash
curl -H 'Authorization: Bearer <token>' http://127.0.0.1:8080/sse
```

Mnemosyne emits an endpoint event like:

```text
data: /messages/?session_id=<hex>
```

POST a normal MCP initialize message:

```bash
curl -X POST \
  'http://127.0.0.1:8080/messages/?session_id=<hex>' \
  -H 'Authorization: Bearer <token>' \
  -H 'Content-Type: application/json' \
  -d '{
    "jsonrpc":"2.0",
    "method":"initialize",
    "id":1,
    "params":{
      "protocolVersion":"2024-11-05",
      "capabilities":{},
      "clientInfo":{"name":"test","version":"0.0.1"}
    }
  }'
```

## Actual result

`/sse` returns `200` and `/messages/?session_id=...` returns `202`, but Starlette throws:

```text
TypeError: 'NoneType' object is not callable
  File ".../starlette/routing.py", line 62, in app
    await response(scope, receive, send)
```

The root of the problem appears to be in `mnemosyne/mcp_server.py` inside `_build_sse_app()`:

```python
transport = SseServerTransport("/messages")

async def handle_sse(request):
[...]

PR #652, opened 2026-08-07
test(mcp): cover SSE messages mount lifecycle

aaka3207:
## Description

Adds regression coverage for the standalone SSE `/messages` route lifecycle fixed in #383.

The tests lock in three properties:

- `SseServerTransport` advertises `/messages/` and its raw ASGI POST callable is mounted at the matching Starlette path.
- A malformed session ID returns a clean `400 Invalid session ID`.
- A valid-but-absent session ID returns a clean `404 Could not find session`.

This prevents a regression to wrapping `handle_post_message` in a function `Route`, where Starlette tried to invoke its `None` return after the transport had already sent the response, producing `TypeError: 'NoneType' object is not callable`.

No production code changes are needed because current `main` already contains the `Mount` fix.

## Related Issue

Regression coverage for #383.

## Type of Change

- [x] Test improvement

## Checklist

- [x] Added targeted regression tests
- [x] Relevant MCP/SSE suite passes: `99 passed`
- [x] Ruff passes for the changed test file
- [x] Independently reviewed with no security or logic findings
- [x] No production API or schema changes
- [x] No README, changelog, or version bump required for a test-only change

## Summary

- Adds regression tests for the standalone SSE `/messages/` route in the MCP integration surface.
- Verifies matching `SseServerTransport` advertisement and Starlette mount behavior.
- Verifies `400 Invalid session ID` and `404 Could not find session` responses.
- No production code changes affect memory architecture, tiered memory, retrieval, consolidation, veracity, sync, privacy, or local-first guarantees.
- Hermes and CLI integration remain unchanged.
- The tests improve long-term maintainability by protecting the existing SSE lifecycle fix without adding architectural complexity.

### DIFF
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -1111,6 +1111,64 @@ print(json.dumps({"result": result, "after": after}))
         sse_route = next(route for route in app.routes if getattr(route, "path", None) == "/sse")
         assert any(cell.cell_contents is build.return_value for cell in (sse_route.endpoint.__closure__ or ()))
 
+    def test_sse_messages_transport_uses_matching_mount_path(self):
+        """The advertised POST URI and mounted raw ASGI app must agree."""
+        from mnemosyne import mcp_server
+        from starlette.routing import Mount
+        from unittest.mock import AsyncMock, patch
+
+        handle_post_message = AsyncMock()
+        transport = type("FakeTransport", (), {"handle_post_message": handle_post_message})()
+        transport_factory = patch("mcp.server.sse.SseServerTransport")
+        with (
+            patch.object(mcp_server, "_build_mcp_server", return_value=object()),
+            transport_factory as build_transport,
+        ):
+            build_transport.return_value = transport
+            app = mcp_server._build_sse_app(host="127.0.0.1")
+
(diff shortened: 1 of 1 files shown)


dplush (changes requested):
Thanks for adding the path and error-response coverage. I re-ran this against a real SSE session and found the central regression path is still not exercised.

The new POST tests use only malformed or unknown session IDs, so they never reach `handle_post_message` after a live `GET /sse` session has already emitted the endpoint and sent its `202`. That success path is where a Mount/Route wrapping regression can produce a second ASGI response or `NoneType` callable failure.

Please add one lifecycle-level regression that:
1. opens `/sse` with a context-managed client,
2. reads the advertised endpoint URI and session ID,
3. POSTs to exactly that URI and asserts a single successful `202` response, and
4. closes/disconnects cleanly so the test cannot leak SSE/AnyIO tasks.

Please also cover the distinct missing-`session_id` request (`400 session_id is required`); the current “missing session” test covers a valid-but-unknown UUID (`404`) instead.

The current Mount and invalid/unknown-session checks are useful supporting coverage, but they do not replace this public success-path regression. After that and a fresh CI cycle, I expect the test-only scope to be straightforward to approve. The CLA also remains required before merge.

Issue #657, opened 2026-08-07
CONTRIBUTING setup cannot run integrations/hermes tests — 132 collection failures on a pristine checkout


aldoeliacim:
## Summary

The setup documented in `CONTRIBUTING.md` cannot run the `integrations/hermes/tests/` suite. 132 tests fail at collection with `ModuleNotFoundError: No module named 'mnemosyne_hermes'`.

## Reproduction

Exactly the documented steps:

```bash
git clone https://github.com/AxDSan/mnemosyne.git
cd mnemosyne
pip install -e ".[all,dev]"
python -m pytest tests/ integrations/hermes/tests/ -q
```

Result:

```
132 failed, 2511 passed, 2 skipped
ModuleNotFoundError: No module named 'mnemosyne_hermes'
```

## Cause

`mnemosyne_hermes` is a **separate distribution** (`integrations/hermes/pyproject.toml`, `name = "mnemosyne-hermes"`, version 0.5.0). The root `pyproject.toml` never references it — not as a dependency, not as an optional extra, not as a path entry. So `pip install -e ".[all,dev]"` installs `mnemosyne-memory` only, and every test importing `mnemosyne_hermes` fails to collect.

Adding the missing install fixes it:

```bash
pip install -e ./integrations/hermes
python -m pytest integrations/hermes/tests/test_upgrade_wrapper.py -q
# 3 passed
```

## Why it matters

A contributor following CONTRIBUTING sees 132 failures on a pristine checkout, before writing any code. That makes the suite useless as a regression signal: with a broken baseline it is impossible to tell whether your own change broke something, and the natural (wrong) conclusion is that the repo is simply red.

It also silently weakens CI-adjacent local verification — the affected tests cover the wrapper/install machinery, which is exactly the area where local verification matters most.

## Suggested fix

Either add a dev extra that pulls the sibling distribution:

```toml
[project.optional-dependencies]
dev = [..., "mnemosyne-hermes @ file:./integrations/hermes"]
```

or document the second install step in CONTRIBUTING:

```bash
pip install -e ".[all,dev]"
pip install -e ./integrations/hermes   # required for integrations/hermes/tests
```

The extra is preferable — a documented step that can be skipped will be skipped.

## Environment

mnemosyne-memory 3.15.3 @ `7254509`, Python 3.11.15, Linux aarch64, fresh venv built exactly per CONTRIBUTING.

aldoeliacim:
[...]

Issue #655, opened 2026-08-07
Two forked copies of the Hermes memory provider have drifted (53 fixes landed in only one)


aldoeliacim:
## Summary

The repository carries **two independently maintained copies** of the Hermes memory provider. They have drifted: fixes land in one and not the other, and users can end up on either depending on which README they followed.

## The two copies

| | `hermes_memory_provider/` | `integrations/hermes/src/mnemosyne_hermes/` |
|---|---|---|
| Introduced | `16cc7dd` (2026-04-19) | `0ee4d80` (2026-05-29, "modularize provider") |
| LOC (package) | 5,336 | 7,121 |
| Ships `plugin.yaml` | yes | yes |
| Has init-retry (#496) | **no** | yes |
| Test suite | none dedicated | `integrations/hermes/tests/` |

Both are deployable plugins. Both have their own install path documented:

- root `README.md:400` symlinks **`mnemosyne_hermes`**
- `hermes_memory_provider/README.md` symlinks **`hermes_memory_provider`**

## Evidence of drift

`#496` ("fix: add init retry with backoff") touched only `mnemosyne_hermes`. A user on the `hermes_memory_provider` path therefore still hits the permanent memory outage described in #654 — filed today from a live 3.15.1 install.

The copies are not simple duplicates; each has functionality the other lacks:

- only in `hermes_memory_provider`: `PrefetchProfile`, `register_profile`, `register_prefetch_source`, `_prefetch_bank`, `_prefetch_identity`, `_identity_fichas`, `_resolve_profile`, `_dedup_blocks`, … (13 symbols)
- only in `mnemosyne_hermes`: `HermesPersonaPromptMixin`, `BatchValidationError`, `apply_beam_batch`, `_maybe_retry_init`, `_is_transient_init_error`, `_persona_block`, `_canonical_prefetch_rows`, … (15 symbols)

So this cannot be resolved by deleting one directory — a real merge is required.

## Why it matters

Every future fix to this provider has to be written twice, and there is nothing in CI that enforces parity. #496 is the demonstration: the fix existed upstream for weeks while affected users had no way to know they were on the copy that lacked it.

## Suggested direction

Not prescriptive — the maintainers know the intended end state better than I do. Options as I see them:

[...]

Issue #651, opened 2026-08-07
Migration: move users from legacy hermes_memory_provider plugin path to standalone mnemosyne-hermes provider


ether-btc:
## Goal

Migrate all users from the legacy source-checkout plugin path (`hermes_memory_provider/` symlinked at `~/.hermes/plugins/mnemosyne`) to the standalone `mnemosyne-hermes` pip package — the intended provider path per maintainer direction.

## Context

Per dplush's review on PR #647: "Current `mnemosyne/install.py` still creates that link [the legacy symlink], while the standalone `mnemosyne-hermes` package is the intended provider path. The project direction is to migrate all users to `mnemosyne-hermes`, not to maintain two parallel integration stacks."

The legacy `hermes_memory_provider/` plugin (symlinked at `~/.hermes/plugins/mnemosyne`) is an obsolete integration route. The `mnemosyne-hermes` package (`integrations/hermes/src/mnemosyne_hermes/`) is the supported provider.

## Proposed Migration

1. **Update installer** (`mnemosyne/install.py`): Stop creating the legacy symlink. Install `mnemosyne-hermes` via pipx/pip instead.
2. **Migration path for existing users**: Detect the legacy symlink, warn, and offer to migrate to the standalone provider.
3. **Verification step**: Add a verification command that confirms the standalone provider is installed and active.
4. **Fail clearly**: When the obsolete route is detected (legacy symlink present without the standalone provider), emit a clear error message directing users to install `mnemosyne-hermes`.

## Related

- PR #647: CLA-resolved, split to CI-only (no hermes_plugin restore)
- PR #650: Bare except fix in legacy provider (makes silent failures visible)
- Issue #649: Documents the silent failure bug
- Comment on PR #647: Maintainer direction clarification

dplush:
@AxDSan I think #651 should be the parent migration issue for the Hermes integration.

The intended end state is clear: all users move to the standalone `mnemosyne-hermes` provider. We should not restore or retain `hermes_plugin` or the source checkout `hermes_memory_provider` route as a parallel supported integration.

One release prerequisite needs an explicit decision first. The current PyPI artifact is `mnemosyne-hermes` 0.5.0, while main contains Hermes provider and wrapper changes that have not yet been published, tracked in #631. Migrating users before a current package release could move them away from the legacy path without giving them the current supported behavior.

Suggested sequencing:

1. Publish the next `mnemosyne-hermes` release with the current provider and wrapper contract.

[...]

PR #658, opened 2026-08-07
fix(hermes): retry transient init failure in hermes_memory_provider (#654)

### LINKED ISSUE #654: hermes_memory_provider: one transient SQLite lock disables memory for the whole session (init-retry from #496 missing in this copy)
## Summary

`hermes_memory_provider` disables memory for an entire session after a single **transient** SQLite lock, while the database is healthy again seconds later. The init-retry fix from #496 landed only in the sibling `mnemosyne_hermes` copy.

## Impact

Observed live on 2026-08-07 (v3.15.1). A session initialized while another writer held the SQLite write lock. That session then returned `status: memory_unavailable` for **every** `mnemosyne_*` tool call for hours, while sibling sessions started minutes later initialized fine against the same database. A process restart was the only recovery path.

The database was never damaged: 4,088 rows in `working_memory`, direct `BeamMemory(...).remember()` succeeded from a separate interpreter at the same moment the in-session tools were failing.

## Root cause

`initialize()` stores the exception permanently:

```python
except Exception as e:
    logger.warning("Mnemosyne init failed: %s", e)
    self._beam = None
    self._init_error = e          # never cleared, never retried
```

Every subsequent tool call short-circuits before touching SQLite and replays that stale exception:

```python
def handle_tool_call(self, tool_name, args, **kwargs):
    ...
    if not self._beam:
        reason = self._init_error_reason()
        return json.dumps({"status": "memory_unavailable", ..., "error": f"Mnemosyne unavailable: {reason}"})
```

So a sub-second lock becomes a permanent, session-lifetime outage. Because the read path (`mode=ro`) keeps working and the DB file is fine, this is easy to misdiagnose as corruption.

## Why this copy specifically

`mnemosyne_hermes` already has the fix (`_is_transient_init_error`, `_retry_init_args`, `_maybe_retry_init`), shipped in #496. `hermes_memory_provider` never received it, despite `hermes_memory_provider/README.md` documenting it as the deployment target:

```
ln -s /path/to/mnemosyne/hermes_memory_provider ~/.hermes/plugins/mnemosyne
```

Installs created from that README (mine dates to 2026-05-29) run the copy without the fix. See #655 for the underlying duplication.

## Reproduction

```python
import sqlite3, hermes_memory_provider
from hermes_memory_provider import MnemosyneMemoryProvider

class _LockedBeam:
    def __init__(self, *a, **k):
        raise sqlite3.OperationalError("database is locked")

hermes_memory_provider._get_beam_class = lambda: _LockedBeam
p = MnemosyneMemoryProvider()
p.initialize("sess")

# Lock is now gone; point at a healthy Beam.
hermes_memory_provider._get_beam_class = lambda: WorkingBeam
print(p.handle_tool_call("mnemosyne_stats", {}))
# -> {"status": "memory_unavailable", ...} forever, despite a healthy DB
```

## Expected

A transient failure (`database is locked`, `disk i/o error`) should be re-attempted from the per-turn surfaces, exactly as `mnemosyne_hermes` does. Non-transient failures (corrupt DB, missing extras, permissions, schema mismatch) should keep the current fail-once behavior so retries cannot spam the log.

## Environment

- mnemosyne-memory 3.15.1, Python 3.11.15, Linux aarch64
- Deployed via the symlink documented in `hermes_memory_provider/README.md`
- Contributing factor: `MNEMOSYNE_BUSY_TIMEOUT_MS` defaults to 5000 ms while `auto_sleep_enabled: true` + `sleep_batch: 5000` opens long write transactions, so the lock window is reachable in normal operation. Raising it to 30000 reduced frequency but cannot fix the permanent-latch behavior.

I have a PR ready that ports the fix with tests (including one covering `handle_tool_call` recovery, which the existing `mnemosyne_hermes` suite does not assert).

aldoeliacim:
Fixes #654.

## Problem

`hermes_memory_provider` disables memory for an entire session after a single **transient** SQLite lock, while the database is healthy again seconds later.

`initialize()` stores the exception permanently and nothing ever retries or clears it:

```python
except Exception as e:
    logger.warning("Mnemosyne init failed: %s", e)
    self._beam = None
    self._init_error = e          # never cleared
```

Every later tool call short-circuits before touching SQLite and replays that stale exception as `status: memory_unavailable`. A sub-second lock therefore becomes a session-lifetime outage, recoverable only by restarting.

The fix for this already exists in the sibling `mnemosyne_hermes` copy (#496) — it just never reached this one. See #655 for the underlying duplication.

## Observed live

2026-08-07, v3.15.1. A session initialized while another writer held the write lock, then returned `memory_unavailable` for **every** `mnemosyne_*` call for hours. Sibling sessions started minutes later initialized fine against the same database, and a direct `BeamMemory(...).remember()` from a separate interpreter succeeded at the same moment the in-session tools were failing — 4,088 rows intact, nothing corrupt.

Because the read path keeps working and the file is healthy, this reads as corruption and invites the wrong fix (repair/restore).

## Change

Ports the #496 mechanism verbatim from `mnemosyne_hermes`:

- `_is_transient_init_error()` — `"database is locked"`, `"disk i/o error"`
- `_retry_init_args` / `_retry_init_at`, spaced by `_INIT_RETRY_INTERVAL_S = 60.0`
- `_maybe_retry_init()` invoked from the four per-turn surfaces: `system_prompt_block`, `prefetch`, `sync_turn`, `handle_tool_call`
- retry-pending wording in the UNAVAILABLE prompt block, so the banner stops telling users to restart when a retry is already scheduled
- a fresh `initialize()` clears any pending retry (the except path re-stashes if that attempt also fails)

Non-transient failures (corrupt DB, missing extras, permissions, schema mismatch) keep the fail-once behavior, so retries cannot spam the log.

## Tests

`tests/test_hermes_provider_init_retry.py` mirrors `integrations/hermes/tests/test_init_retry.py` and adds one case that suite does not have:

- `test_handle_tool_call_recovers_after_transient_failure` — asserts recovery on the surface users actually hit, rather than only on `system_prompt_block`

**Verified by rejection**: with the fix reverted, 5 of the 6 fail. Both copies' suites pass together (11 passed).

Full suite run in an isolated `[all,dev]` venv, A/B against the same tree with the fix stashed: no regressions, `+6` passed (exactly the new tests).

Note for reviewers: running `integrations/hermes/tests/` locally also requires `pip install -e ./integrations/hermes`, which CONTRIBUTING's documented setup omits — filed separately as #657.

Version bumped to 3.15.3 and CHANGELOG updated per CONTRIBUTING. CLA agreed.

## Summary

- Adds retry handling for transient `hermes_memory_provider` initialization failures.
- Retries SQLite `"database is locked"` and `"disk i/o error"` failures after 60 seconds.
- Supports retries from prompt, prefetch, sync, and `handle_tool_call` paths.
- Preserves fail-once behavior for permanent failures.
- Updates unavailable-status messages while a retry is pending.
- Clears pending retry state after successful initialization.
- Adds recovery and interval-gating tests.
- Updates the package version to `3.15.3`.

## Architectural impact

- Improves Hermes memory availability without changing the core memory architecture.
- Does not change retrieval, consolidation, veracity, sync, or benchmark methodology.
- Strengthens the local-first design because temporary local SQLite contention no longer disables memory for the full session.
- Keeps failures local. The change adds no external service, telemetry, or data transfer.
- Extends recovery across Hermes agent integration surfaces. MCP and CLI behavior are unchanged.
- Centralized transient-error classification and retry-state handling improve long-term maintainability.
- Retaining fail-once behavior for non-transient errors avoids masking corruption, permission errors, missing dependencies, and schema mismatches.

## Assessment

This is the right call for transient local database failures. It improves session resilience while preserving deterministic handling for permanent initialization errors. No change in this PR erodes local-first guarantees or introduces a privacy concern.


dplush:
Thank you for the careful reproduction and narrow retry implementation. The underlying incident is real.

We are not going to merge this patch because it extends `hermes_memory_provider`, the legacy path we are actively retiring. The supported destination is the standalone `mnemosyne-hermes` provider, and #651 tracks the migration away from legacy source-checkout installs. Keeping retries in both copies would continue the drift documented in #655.

Please do not spend more time addressing the current CI or review findings on this branch. #657 is independently useful and can be handled as a small contributor-setup fix.

dplush:
Closing this because it extends the legacy `hermes_memory_provider` path that we are retiring. The supported destination is the standalone `mnemosyne-hermes` provider; keeping retry behavior in both paths would continue the drift tracked in #655. Thanks again for the careful reproduction and focused patch.

PR #677, opened 2026-08-09
fix: record fallback_rate on the polyphonic recall path

cgray2222:
## Summary

`mnemosyne_recall_diagnostics` reports `fallback_rate` as `0` on every recall under `MNEMOSYNE_POLYPHONIC_RECALL=1` — even when the vector voice degrades from the sqlite-vec fast path to a numpy full-scan.

### Root cause

The polyphonic diagnostics block (added in #668) records tier hits and call counts, but never calls `record_fallback_used()`. The linear path records `_wm_fallback_used = not wm_ids` and `_em_fallback_used = not episodic_rowids`; the polyphonic branch had no equivalent signal, so `wm_fallback_rate`/`em_fallback_rate` were dead zeros under the flag production deployments use.

### Fix

The polyphonic engine now exposes a per-call degraded-path flag (`last_call_fallback`) and the polyphonic diagnostics block records it:

- **`mnemosyne/core/polyphonic_recall.py`** — the vector voice sets `last_call_fallback["em"] = not em_consumed_via_vec_episodes`, i.e. True when the EM tier ran the numpy full-scan because sqlite-vec was unavailable, errored, or its top-K ANN hits all dropped in the superseded/valid_until JOIN.
- **`mnemosyne/core/beam.py`** — the polyphonic branch calls `record_fallback_used(em=...)` from that signal before `record_call()`. On engine failure the flag resets to defaults so a broken call isn't attributed to fallback.

`wm` stays False by design: the polyphonic engine has no substring-scoring tier for working memory (the linear path's wm_fallback only fires on the substring path).

Recording is read-only signal — it never alters recall behavior.

### Verification

New tests in `tests/test_c4_recall_diagnostics.py` (5 cases):

- Engine flag defaults `{"em": False, "wm": False}`
- sqlite-vec unavailable → `em=True` flag (numpy degradation surfaced)
- sqlite-vec healthy fast path → flag stays False
- E2E degraded path: `em_fallback_rate == 1.0`, `wm_fallback_rate == 0.0`
- E2E healthy path: `em_fallback_rate == 0.0`

Full suite: 2460 passed, failure set identical to clean upstream main (111 pre-existing environment failures in sync/import/LLM suites — no regressions).

### Docs

- `docs/benchmarking.md` — polyphonic caveat updated: `em_fallback_rate` is real (sqlite-vec → numpy degradation), `wm_fallback_rate` stays 0 by design.
- `CHANGELOG.md` — entry under `[Unreleased] → Fixed`.

## Summary

- Fixes `fallback_rate` reporting for polyphonic recall.
- Records per-call EM fallback when sqlite-vec retrieval uses a NumPy full scan.
- Keeps WM fallback at `False` because polyphonic recall has no WM substring tier.
- Resets fallback state on each call and after engine failure.
- Adds regression coverage for healthy, degraded, unavailable, reset, and end-to-end paths.
- Updates documentation and the changelog.

## Architectural impact

- Improves diagnostics for the episodic tier without changing recall behavior.
- Preserves sqlite-vec ANN and NumPy full-scan retrieval strategies.
- Does not change working-memory retrieval, BEAM behavior, consolidation, veracity, or sync.
- Makes degraded retrieval visible and testable. This is the right call for maintainability.

## Local-first and integration impact

- Preserves local-first operation.
- Keeps NumPy full-scan fallback available when sqlite-vec is unavailable or fails.
- Introduces no network dependency and does not change the privacy posture.
- Does not erode local-first guarantees.

## Benchmark impact

- Adds diagnostic regression tests.
- Documents `em_fallback_rate` and the zero `wm_fallback_rate`.
- The reported suite contains 2,460 passing tests and 111 pre-existing environment failures.


AxDSan:
Thanks for the fallback_rate change on the polyphonic recall path. The branch is conflicting with main at the moment. Could you rebase onto current main and push the refreshed head? I'll review right after it's clean.

cgray2222:
Rebased onto current main (`f6ba5c8`) and pushed the refreshed head — conflict resolved in `CHANGELOG.md` (kept the merged entries + the new fallback_rate entry, dropped the stale duplicate).

Also addressed the outstanding CodeRabbit findings while I was in here:
- Reset `last_call_fallback` at the top of `_vector_voice` so an early return (voice disabled / missing embedding / empty vector / zero norm) can't leak a prior call's degraded state into the diagnostics — with a two-call regression test.
- Healthy-path fallback contract is now covered deterministically (delegating fake connection serving the ANN result) instead of relying solely on a sqlite-vec-dependent test that skips without the extension.
- Both E2E tests now seed an embedding-backed episodic memory, assert the expected content is returned, and pin `last_call_fallback['wm'] is False`.
- `test_tier_attribution_no_double_count` now snapshots diagnostics from the same recall call it asserts on.

CI re-running on `6f8bbbb`. Ready for review. 👍

dplush (changes requested):
Thanks — this is a useful diagnostics-only fix, and the healthy/degraded coverage is much better after the earlier review round. I found two blockers on the current head.

1. `last_call_fallback` is mutable state on the cached `PolyphonicRecallEngine`, then copied through mutable `BeamMemory._last_polyphonic_fallback`. Concurrent recalls can overwrite one call’s flag before that call reaches `record_fallback_used()`, so the global gauge can attribute a degraded call as healthy (or vice versa). Please make fallback attribution call-local, for example return it with the engine result rather than storing it on the cached engine/beam, and add a deterministic concurrent degraded-vs-healthy regression.

2. The six new vector diagnostics tests import NumPy unconditionally, but the declared `.[test]` extra contains only pytest/build. In a clean test-only environment, this file is 28 passed / 6 failed with `ModuleNotFoundError: numpy`; it passes with `.[embeddings,test]`. Please either declare the dependency deliberately or skip these optional-feature tests when NumPy is unavailable.

I verified the focused file passes with the embeddings extra, and current CI is green. These are diagnostics correctness and declared-test-environment blockers; no need to expand this PR into #700 or alter recall ranking.

Issue #660, opened 2026-08-08
[BUG] mnemosyne_invalidate reports success but silently no-ops on cross-session, session-scoped memories


PaoloBarts:
## Description

`mnemosyne_invalidate` reports `"status": "invalidated"` for memories it cannot actually invalidate, with no error surfaced to the caller — a silent data-integrity trap.

**Root cause:** `Beam.invalidate()` only updates rows matching the current session or global scope:

```sql
UPDATE working_memory
SET valid_until = ?, superseded_by = ?
WHERE id = ? AND (session_id = ? OR scope = 'global')
```

Rows that are `scope='session'` and belong to a *different* session (e.g. `sleep_model_refresh_proposal` artifacts written by the sleep/consolidation session, or conversation captures from another session) match nothing → `rowcount = 0` → the method returns `False`. The Hermes provider adapter (`_handle_invalidate`) **ignores the return value** and always responds `{"status": "invalidated"}`.

**Observed:** 425 `sleep_model_refresh_proposal` rows were each reported `"status": "invalidated"`; afterwards all 425 still had `valid_until IS NULL` and were still returned by `mnemosyne_recall`. No error anywhere.

The same rows cannot be resolved by any other path either:

- `mnemosyne_batch` invalidate → `BatchOperationError: memory_not_found`
- CLI `mnemosyne delete <id>` → `Error: Memory not found` (exit code 0)
- `mnemosyne_forget` → `{"status": "not_found"}`

The only working path found is direct SQL deletion (with cascade to `memory_embeddings` / `gists`) behind a backup.

## Environment

- mnemosyne-memory 3.15.1, mnemosyne-hermes 0.5.0
- Hermes v0.20.0 (2026.8.3), Windows 10 (build 26100), git-bash (MSYS2)
- Store: ~1,600 working rows, including 425 `sleep_model_refresh_proposal` rows (`scope='session'`, `session_id` = the consolidation run, not the current session)

## Reproduction Steps

1. Take any memory with `scope='session'` whose `session_id` differs from the current session (sleep-consolidation artifacts qualify).
2. Call `mnemosyne_invalidate(memory_id=<id>)`.
3. Observe the response: `{"status": "invalidated", "memory_id": <id>}`.
4. Inspect the row: `valid_until` is still `NULL`, and `mnemosyne_recall` still returns the memory.

Control: a memory with `scope='global'` (or in the current session) invalidates correctly and is excluded from recall.

## Expected Behavior

Either:
[...]

PR #542, opened 2026-07-25
fix(hermes): report memory_not_found when invalidate matches no row

Sanjays2402:
## Description

`_handle_invalidate` in `mnemosyne_hermes` discarded the boolean returned by `BeamMemory.invalidate()` and always responded `status=invalidated`, so a wrong, inaccessible, or already-removed `memory_id` looked like a successful mutation. The return value is now checked: `False` yields `status=memory_not_found` and is recorded in the audit metadata. This matches the batch path in `mnemosyne/batch_tool.py`, which already raises `memory_not_found` on the same signal.

## Related Issue

Closes #524

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [x] Existing tests still pass (`pytest integrations/hermes/tests/ -q` — 84 passed; the one failure, `test_package_version_matches_distribution_metadata`, is a pre-existing source-tree artifact requiring installed distribution metadata and fails identically without this change)
- [x] New tests added for any new functionality
- [x] Manual verification (describe what you tested)

New `test_invalidate_reports_not_found_when_no_row_matched` asserts `beam.invalidate()` returns `False` for an unknown id and that the tool then reports `memory_not_found`. Stash-verified: it FAILS without the source fix (`- memory_not_found` / `+ invalidated`) and PASSES with it.

## Checklist

- [ ] Version bumped in `mnemosyne/__init__.py` — N/A, left to maintainers' release process
- [ ] `CHANGELOG.md` updated with a brief entry — happy to add one if you'd like it in this PR
- [ ] README updated if user-facing behavior changed — N/A, tool response field only
- [x] Code follows the project's principles:
  - No new external dependencies without good reason
  - No cloud / API key requirement for core functionality
  - SQLite is the only database dependency

## Summary

- Fixes Hermes `mnemosyne_invalidate` responses by honoring `BeamMemory.invalidate()`’s boolean result.
- Returns `memory_not_found` for unknown, inaccessible, or already-removed memory IDs.
- Records the invalidation result in audit metadata.
- Adds regression coverage for unmatched IDs.

## Architectural impact

This is a targeted correctness fix in the BEAM-backed memory mutation path. It does not change tiered memory, retrieval, consolidation, veracity, synchronization, or benchmarking behavior.

[...]

PR #679, opened 2026-08-09
fix(mcp): mnemosyne_invalidate must not record a supersession link that resolves nowhere

### LINKED ISSUE #676: mnemosyne_invalidate records an unresolved replacement_id, hiding a real memory from recall
## Summary

`mnemosyne_invalidate` returns `{"status": "invalidated"}` for memory IDs that do not exist, and for IDs that exist in `working_memory` but which it cannot act on. The caller is told the invalidation succeeded when nothing was changed.

This matters more than a normal wrong-return-value, because invalidation is the mechanism a caller relies on to retire a fact that has become false. A silent no-op means the stale fact stays recallable while the caller believes it has been superseded.

## Reproduction

Version 3.15.1, fresh install, `pip install mnemosyne-memory`.

```python
from mnemosyne.mcp_tools import handle_tool_call as h

# 1. A completely fabricated ID
h('mnemosyne_invalidate', {'memory_id': 'deadbeefdeadbeef'})
# -> {'status': 'invalidated', 'memory_id': 'deadbeefdeadbeef'}

# 2. A fabricated ID plus a fabricated replacement
h('mnemosyne_invalidate', {'memory_id': 'notarealid00000', 'replacement_id': 'alsofake0000000'})
# -> {'status': 'invalidated', 'memory_id': 'notarealid00000'}
```

Both report success. Nothing exists to invalidate.

## The case that led me here

```bash
mnemosyne store "Use backend A for this job. Verified 5 Aug." src 0.8
# Stored: caa1867cdecd035f
mnemosyne store "CORRECTION: do NOT use backend A, it produces bad output. Use backend B." src 0.95
# Stored: c0390457e9e9434b
```

```python
h('mnemosyne_invalidate', {'memory_id': 'caa1867cdecd035f',
                           'replacement_id': 'c0390457e9e9434b'})
# -> {'status': 'invalidated', 'memory_id': 'caa1867cdecd035f'}

h('mnemosyne_get', {'memory_id': 'caa1867cdecd035f'})
# -> {'status': 'not_found', 'memory_id': 'caa1867cdecd035f'}
```

The row is untouched:

```
sqlite> select id, importance from working_memory where id='caa1867cdecd035f';
caa1867cdecd035f|0.8
```

And CLI recall still returns the superseded memory, ranked above its own correction:

```
$ mnemosyne recall "which backend should I use" 5
  ID: caa1867cdecd035f   Score: 0.578    <- the stale one, first
  ID: c0390457e9e9434b   Score: 0.361    <- the correction, second
```

## Probable cause

Freshly stored memories land in `working_memory`. The supersession columns (`superseded_by`, `valid_until`, `corrected_by`) are on `episodic_memory`, and `mnemosyne_get` / `mnemosyne_invalidate` appear to resolve IDs against `episodic_memory` only. Memories reach `episodic_memory` via consolidation, and `mnemosyne sleep` is age-gated:

```
$ mnemosyne sleep
Consolidation complete: {'status': 'no_op', 'message': 'No old working memories to consolidate'}
```

So a correction written on the same day as the claim it corrects cannot supersede it, and the API reports that it did.

## Suggested behaviour

1. `mnemosyne_invalidate` should resolve the ID before acting, and return `{"status": "not_found", "memory_id": ...}` when it cannot. Never report `invalidated` without a row change.
2. Same for a `replacement_id` that does not resolve: either error, or return a status making clear the replacement link was not recorded.
3. Ideally `invalidate` should work on `working_memory` rows too, since that is where a memory lives for its first hours. If that is out of scope, the error should say so explicitly rather than failing silently, e.g. `{"status": "not_invalidatable", "reason": "memory is in working tier; consolidate first"}`.

## Suggested test cases

```python
def test_invalidate_unknown_id_reports_not_found():
    assert h('mnemosyne_invalidate', {'memory_id': 'deadbeefdeadbeef'})['status'] != 'invalidated'

def test_invalidate_unknown_id_changes_nothing():
    before = row_count_with_valid_until_set()
    h('mnemosyne_invalidate', {'memory_id': 'deadbeefdeadbeef'})
    assert row_count_with_valid_until_set() == before

def test_invalidate_unresolvable_replacement_does_not_claim_success():
    mid = store_and_consolidate("fact A")
    r = h('mnemosyne_invalidate', {'memory_id': mid, 'replacement_id': 'notarealid00000'})
    assert r['status'] != 'invalidated' or r.get('replacement_recorded') is False

def test_invalidate_working_tier_memory_is_honest():
    mid = store_without_consolidating("fact B")           # lands in working_memory
    r = h('mnemosyne_invalidate', {'memory_id': mid})
    # either it works...
    if r['status'] == 'invalidated':
        assert recall_does_not_return(mid)
    # ...or it says clearly that it did not
    else:
        assert r['status'] in ('not_found', 'not_invalidatable')
```

## Environment

- mnemosyne-memory 3.15.1 (pip), Python 3.13.14, macOS (Darwin 25.5.0)
- DB at `~/.hermes/mnemosyne/data/mnemosyne.db`, default bank, mode `beam`

## Unrelated minor note

`MNEMOSYNE_DB` was set to a path in a scratch directory and appears to have been ignored; the database was created at `~/.hermes/mnemosyne/data/mnemosyne.db`. Happy to open that separately if it is not already known.

## Aside, and thank you

The canonical store is excellent and does exactly what I wanted here. `mnemosyne_remember_canonical` on a `(category, name)` slot superseded the old value immediately, kept version 1 in history with `valid_until` stamped, and `recall_canonical` returns only the current value with no ranking involved. That is the right answer for governed facts. This issue is only about `invalidate` reporting success when it has not acted.

actionjack:
Fixes #676

## What the report contains, and what is actually still broken

The issue reports two behaviours of `mnemosyne_invalidate`, verified there on `mnemosyne-memory` 3.15.1. I checked both against `main` (97289a3) before writing any code, on Python 3.13 / macOS, with `HOME` and `HERMES_HOME` pointed at a throwaway directory:

1. **Unknown `memory_id` returning `invalidated`.** Already fixed on `main` by #542, #660 and #675. On `main` it returns `memory_not_found`. This PR only adds regression coverage so it stays fixed.
2. **An unresolvable `replacement_id` is accepted and written into `superseded_by`.** Still live on `main`. This is what this PR fixes.

The issue's suspected cause (IDs resolved against `episodic_memory` only, working-tier rows unreachable until consolidation) does not hold on `main`. `BeamMemory.invalidate` tries `working_memory` first and falls back to `episodic_memory`, so a freshly stored, unconsolidated memory does invalidate for real.

## Root cause of the remaining bug

`mnemosyne/mcp_tools.py:510-520` passes `replacement_id` straight through and only inspects the outcome for the target row:

```python
if not mem.invalidate(memory_id, replacement_id=replacement_id):
    return {"status": "memory_not_found", "memory_id": memory_id}
return {"status": "invalidated", "memory_id": memory_id}
```

`mnemosyne/core/beam.py:4212-4240` writes whatever it is given:

```python
cursor.execute("""
    UPDATE working_memory
    SET valid_until = ?, superseded_by = ?
    WHERE id = ? AND (session_id = ? OR scope = 'global')
""", (now, replacement_id, memory_id, self.session_id))
if cursor.rowcount > 0:
    ...
    return True
```

Success is decided by `cursor.rowcount`, which is correct for the target. The replacement is never resolved, so any string becomes a supersession pointer. Recall filters on `superseded_by IS NULL` (`beam.py:3023`, `4111`, `6175`, `6297`), so a bogus replacement hides the memory from recall and leaves a link that leads nowhere, while the tool answers `invalidated`.

## Before and after

Both runs against an isolated `HOME`, on `main` and on this branch.

Before:

```
1. unknown id: {'status': 'memory_not_found', 'memory_id': 'deadbeefdeadbeef'}
2. unknown id + unknown replacement: {'status': 'memory_not_found', 'memory_id': 'notarealid00000'}
3. stored working id: f5fda0b9848f38ff
   before: (None, None)
4. invalidate real working id with FAKE replacement: {'status': 'invalidated', 'memory_id': 'f5fda0b9848f38ff'}
   after : ('2026-08-09T14:59:49.763444', 'totallyfake0000')
```

After:

```
1. unknown id: {'status': 'memory_not_found', 'memory_id': 'deadbeefdeadbeef'}
2. unknown id + unknown replacement: {'status': 'replacement_not_found', 'memory_id': 'notarealid00000', 'replacement_id': 'alsofake0000000'}
3. stored working id: c5d63a2ce2ad88aa
   before: (None, None)
4. invalidate real working id with FAKE replacement: {'status': 'replacement_not_found', 'memory_id': 'c5d63a2ce2ad88aa', 'replacement_id': 'totallyfake0000'}
   after : (None, None)
5. invalidate with REAL replacement: {'status': 'invalidated', 'memory_id': 'e9f7d4bcefbd461d'}
   after : ('2026-08-09T15:03:57.473745', 'e94c13990b8a3768')
```

## The change

- `mnemosyne/mcp_tools.py`: new `_resolves_in_scope()` helper, and a guard in `_handle_invalidate` that returns `{"status": "replacement_not_found", "memory_id": ..., "replacement_id": ...}` without touching the database when the replacement resolves nowhere. The helper applies the same scope predicate `BeamMemory.invalidate` uses for the target row, across both tiers, so a link can only point at a memory the caller can already see. Whether the invalidation itself succeeded is still decided by the affected-row count inside `BeamMemory.invalidate`, not inferred.
- `mnemosyne/tool_schemas.py`: the tool description documents the new status.
- `docs/api/tool-schema.mdx`: regenerated with `python3 scripts/generate-docs.py`.
- `CHANGELOG.md`: entry under Unreleased / Fixed.

Two deliberate choices worth a reviewer's attention:

- **`memory_not_found`, not `not_found`.** The issue suggests `not_found`, but `memory_not_found` is the existing convention for this tool (#542, #660), is stated in the tool description, and is asserted by existing tests. Renaming it would be a gratuitous break for MCP clients.
- **Refuse rather than record partially.** An unresolvable replacement could instead invalidate the memory and report that the link was dropped. Refusing is the safer default: the caller asked for a supersession, and a half-completed supersession is what produced the misleading state in the first place. Happy to switch to the partial form if maintainers prefer it.

## Tests added

`tests/test_mcp_server.py`, alongside the existing `test_handle_invalidate_preserves_session_scope_and_reports_not_found`:

- `test_handle_invalidate_unknown_id_changes_nothing`: an unknown ID returns `memory_not_found`, emits no wrapper event, and leaves every `working_memory` row byte-identical.
- `test_handle_invalidate_rejects_unresolvable_replacement`: an unresolvable replacement returns `replacement_not_found`, emits no wrapper event, and leaves `valid_until` and `superseded_by` NULL. Also covers both IDs unknown.
- `test_handle_invalidate_working_tier_records_supersession`: a working-tier memory invalidates for real, with `valid_until` set and `superseded_by` pointing at the replacement, without any consolidation step.
- `test_invalidate_schema_documents_scope_safe_failure`: extended to assert the new status is documented.

## Test command and result

The CI command from `.github/workflows/ci.yml`, run in a Python 3.13.11 venv with `pip install -e ".[embeddings,mcp,sync,test]"` and `pip install -e ./integrations/hermes`:

```
python -m pytest tests/ integrations/hermes/tests/test_wrapper_bootstrap.py integrations/hermes/tests/test_canonical_tools.py -v --tb=short
```

Result on this branch:

```
= 111 failed, 2468 passed, 20 skipped, 7 subtests passed in 629.05s (0:10:29) ==
```

Same command on unmodified `main` (97289a3), same machine, same venv:

```
= 111 failed, 2465 passed, 20 skipped, 7 subtests passed in 649.24s (0:10:49) ==
```

The 111 failures are identical between the two runs (I diffed the `FAILED` lines: the sets match exactly). They are pre-existing and unrelated to this change. Most of them come from test-order pollution on this machine: `tests/test_embedding_prefixes.py` leaves `embeddinggemma-300m-q4` as the resolved embedding model for the rest of the session, so every later test that goes through the wrapper `remember()` path dies with `RuntimeError: Failed to load embedding model embeddinggemma-300m-q4`. The rest are release-gate and sync tests that fail for local-environment reasons. `tests/test_mcp_server.py` on its own is `1 failed, 72 passed`, the single failure being `test_handle_batch_failure_rolls_back`, which I confirmed fails identically on unmodified `main`. I have not tried to fix any of that here.

Lint gates from `ci.yml` also pass on the changed files:

```
ruff check --select F,RUF022 -- mnemosyne/mcp_tools.py mnemosyne/tool_schemas.py tests/test_mcp_server.py   ->  All checks passed!
ruff check --select E9 .                                                                                   ->  All checks passed!
```

## Out of scope, deliberately

- `mnemosyne/batch_tool.py:204`, `hermes_memory_provider/__init__.py:2971` and `integrations/hermes/src/mnemosyne_hermes/__init__.py:1988` call `BeamMemory.invalidate` directly and share the same unchecked-replacement gap. Fixing it centrally in `BeamMemory.invalidate` would flatten "target missing" and "replacement missing" into one boolean, which callers could not tell apart, so I kept the fix at the MCP handler named in the issue rather than reshaping the storage API. Say the word if you would rather have the check pushed down, with a richer return type.
- No version bump. #675 and #660 landed without one, and the maintainers batch bumps.
- The pre-existing test-isolation defect described above.

I have read CONTRIBUTING.md and agree to the CLA.


dplush:
Thanks for the careful correction and for isolating the remaining replacement-link bug. This is now covered by merged #678, which validates replacements in the storage layer for MCP and direct callers, keeps the scope-safe `memory_not_found` contract, and serializes validation with the target-link commit. Closing this as superseded to avoid two competing fixes for the same #676 contract.

PR #689, opened 2026-08-10
fix(hermes-provider): include plugin.yaml in the built wheel's package data

### LINKED ISSUE #656: Wheel does not ship plugin.yaml — documented pip install + symlink produces a plugin Hermes cannot load
## Summary

`pip install mnemosyne-memory` does not ship `plugin.yaml`, but Hermes requires that manifest to load a directory plugin. The install flow documented in `hermes_memory_provider/README.md` therefore produces a plugin Hermes silently refuses to load.

## Detail

Hermes' plugin loader (`hermes_cli/plugins.py:19`):

> Each directory plugin must contain a ``plugin.yaml`` manifest **and** an …

and it looks for exactly `child / "plugin.yaml"` (`plugins.py:1550`).

Both provider copies have a `plugin.yaml` in-tree:

```
hermes_memory_provider/plugin.yaml
integrations/hermes/src/mnemosyne_hermes/plugin.yaml
```

But packaging never includes them:

- `pyproject.toml` has `[tool.setuptools.packages.find]` only — no `package-data`, no `include-package-data = true`
- `MANIFEST.in` covers `README.md`, `CHANGELOG.md`, `LICENSE`, `UPDATING.md`, `CONTRIBUTING.md`, `docs/*.md`, `assets/*` — not `plugin.yaml`

Result from a real install (3.15.3, non-editable, into a Hermes venv):

```
$ ls site-packages/hermes_memory_provider/
__init__.py  audit.py  cli.py  hermes_llm_adapter.py
persona_adapter.py  persona_tools.py  sync_adapter.py
# plugin.yaml is absent
```

Following `hermes_memory_provider/README.md`:

```bash
ln -s .../site-packages/hermes_memory_provider ~/.hermes/plugins/mnemosyne
```

…yields a plugin directory with no manifest.

## Why it is easy to miss

Editable installs (`pip install -e .`) point at the source tree, where `plugin.yaml` **is** present — so this reproduces only with a normal wheel/sdist install. It is also masked for anyone who installed before the packaging config changed, or who copied the file by hand.

## Suggested fix

Either

```toml
[tool.setuptools.package-data]
hermes_memory_provider = ["plugin.yaml"]
mnemosyne_hermes = ["plugin.yaml"]
```

or add to `MANIFEST.in` plus `include-package-data = true`.

Worth a regression test that asserts the built wheel contains both manifests — this class of bug is invisible in development and only appears for end users.

## Environment

mnemosyne-memory 3.15.3 (also present in 3.15.1), Python 3.11.15, Linux aarch64, setuptools backend.

AmirF194:
## Root cause

`pyproject.toml` had `[tool.setuptools.packages.find]` but no `package-data` (and `MANIFEST.in` covers only docs and top-level metadata files), so a normal, non-editable `pip install mnemosyne-memory` omitted `hermes_memory_provider/plugin.yaml` from the wheel. Hermes' plugin loader requires that manifest for a directory plugin (`hermes_cli/plugins.py`, looks for `child / "plugin.yaml"`). An editable install works because it points at the source tree, which is why this only shows up on a real install, exactly as reported.

`integrations/hermes/` is a separate distribution with its own `pyproject.toml`, which already declares `package-data` for `mnemosyne_hermes`, so that wheel is unaffected; this is specific to the root package.

## Fix

Add the missing `package-data` entry for `hermes_memory_provider`.

## Verification

- New test `test_core_wheel_ships_the_hermes_memory_provider_plugin_manifest` builds the core wheel and asserts `hermes_memory_provider/plugin.yaml` is present: fails on unmodified `main`, passes on this branch.
- `tests/` (2632 passed) and `integrations/hermes/tests/` (200 passed) both green in a clean `python:3.11-slim` container; 17 pre-existing failures in `test_repair.py`/`test_release_gates.py` are unrelated to this change and reproduce identically on unmodified `main` in the same container (missing `git` binary and a restricted hardlink syscall in the sandbox, not present in GitHub's runner).
- `ruff check` (the repo's fatal + PR-scoped rulesets), `scripts/generate-docs.py --check`, `scripts/verify-docs.py`, and `python -m build` + `twine check` all pass.
- Did not check the actual Hermes plugin-load path end to end (no Hermes install in this environment); the wheel-contents assertion is the direct check for what the issue describes.

Fixes #656

## Summary

- Includes `hermes_memory_provider/plugin.yaml` in built wheels.
- Fixes Hermes directory-plugin loading for non-editable installations.
- Adds a regression test for the core wheel contents.
- Preserves local-first operation. No new data flow, remote service, or privacy risk is introduced.
- Does not change tiered memory, retrieval, consolidation, veracity, sync, or benchmark methodology.
- Improves the Hermes integration surface. MCP and CLI integration remain unchanged.
- This is the right packaging call because it validates required integration data without changing core memory architecture.
- Improves maintainability by testing package contents and preventing future manifest omissions.


AmirF194:
Checking in, no rush. Signing the CLA needs a browser OAuth click-through I can't complete from this account, so it's still pending, but happy to answer anything about the fix in the meantime.

dplush (approved):
Approved on the current head for the narrow package-data fix. Independent artifact checks built and inspected the wheel and sdist, confirmed `hermes_memory_provider/plugin.yaml` is present and structurally valid, and verified a wheel rebuilt from the sdist retains it. The open CodeRabbit suggestion is useful hardening but non-blocking for the reported wheel regression because existing manifest/tool-surface coverage validates the source contract and the built artifacts were verified directly.

This PR is behind main. Please rebase and rerun CI/artifact verification before merge; current main also carries the separate `examples*` wheel-exclusion fix that this old base does not contain.

AmirF194:
Thanks for checking the wheel and sdist directly. Rebased onto current main (now includes the examples wheel-exclusion fix from #729) and reran the artifact and test verification: the built wheel still ships hermes_memory_provider/plugin.yaml, and tests/test_core_wheel_contents.py passes all three cases including the new top-level allowlist check. Pushed as 1409eeb.

dplush (approved):
Re-reviewed on 1409eeb after the rebase onto current main.

Independent artifact gate: core wheel and sdist both ship a valid hermes_memory_provider/plugin.yaml, byte-identical to source. The only wheel payload delta vs the PR base is that manifest. A wheel rebuilt from the sdist keeps it. Fresh --no-deps install lands the file. Top-level surface stays mnemosyne + hermes_memory_provider; examples/integrations/tests stay out.

CI on this head is green. Remaining CodeRabbit note (assert YAML contents / sdist in the unit test) is follow-up hardening, not a merge blocker for this packaging fix.

AmirF194:
Thanks for double checking the wheel and sdist independently after the rebase. Glad the packaging fix landed cleanly.

PR #694, opened 2026-08-11
fix(beam): preserve degradation savepoint during vec refresh

### LINKED ISSUE #691: [BUG] sqlite3.OperationalError: no such savepoint: degrade_row in degrade_episodic() when embedding refresh fails
# Bug: `sqlite3.OperationalError: no such savepoint: degrade_row` in `degrade_episodic()`

## Description

When `degrade_episodic()` processes episodic memories for tier degradation, it wraps each row's UPDATE + embedding refresh inside a SQLite SAVEPOINT named `degrade_row`. However, when `_refresh_episodic_embedding()` raises an exception (e.g., embedding API failure), the subsequent `ROLLBACK TO degrade_row` fails with:

```
sqlite3.OperationalError: no such savepoint: degrade_row
```

The inner except block catches this and logs a misleading message: `"Regex extraction failed, skipping"`. The error repeats for every row being degraded, producing excessive noise in the logs.

This appears to be a SQLite savepoint lifecycle bug — the savepoint may have already been released or never properly created when the rollback is attempted, possibly due to transaction state changes during exception handling.

## Environment

- **OS:** Windows 11
- **Python versions tested:** 3.12, 3.13 (both affected)
- **Mnemosyne version:** 3.15.1 (latest on PyPI)
- **Installation method:** `pip install mnemosyne-memory`
- **Hermes version:** 0.5.0 (mnemosyne-hermes)

## Reproduction Steps

1. Run a workload that triggers episodic memory degradation (e.g., many files indexed over time, causing old memories to exceed `TIER2_DAYS`)
2. When `_refresh_episodic_embedding()` fails for any row (embedding API unavailable or error), the savepoint rollback fails
3. Error repeats for each degraded row

Minimal reproduction:

```python
from mnemosyne.core.beam import BeamMemory
import sqlite3

# Create a BeamMemory instance with old tier-1 data
beam = BeamMemory(session_id="test", db_path="/tmp/test_degrade.db")

# Insert an old episodic memory (tier 1, backdated)
conn = beam.conn
old_ts = "2020-01-01T00:00:00"  # Very old timestamp
conn.execute(
    "INSERT INTO episodic_memory (id, content, created_at, tier) VALUES (?, ?, ?, 1)",
    ("test-old", "This is a test memory that should be degraded to tier 2 or 3.", old_ts)
)
conn.commit()

# Trigger degradation — if embedding refresh fails, the savepoint bug triggers
result = beam.degrade_episodic(dry_run=False)
print(result)
```

## Expected Behavior

When `_refresh_episodic_embedding()` raises an exception:
1. The SAVEPOINT should correctly roll back the content UPDATE
2. No `sqlite3.OperationalError` should be raised
3. If rollback fails, a clear error message should be logged (not "Regex extraction failed")

## Actual Behavior

```
Traceback (most recent call last):
  File ".../mnemosyne/core/beam.py", line 7787, in degrade_episodic
    cursor.execute("RELEASE degrade_row")
sqlite3.OperationalError: no such savepoint: degrade_row

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../mnemosyne/core/beam.py", line 7791, in degrade_episodic
    cursor.execute("ROLLBACK TO degrade_row")
sqlite3.OperationalError: no such savepoint: degrade_row
```

The log also prints a misleading `"Regex extraction failed, skipping"` message which has nothing to do with regex.

## Mnemosyne Status

```bash
hermes mnemosyne stats
```

(Requires Hermes to be running — output depends on user's environment)

## Optional: Database Info

Database location (default): `~/.hermes/mnemosyne/data/mnemosyne.db`

```bash
sqlite3 ~/.hermes/mnemosyne/data/mnemosyne.db "SELECT COUNT(*) FROM working_memory; SELECT COUNT(*) FROM episodic_memory;"
```

## Root Cause Analysis

In `beam.py` around line 7780-7810, the code structure is:

```python
for row in tier1_rows:
    cursor.execute("SAVEPOINT degrade_row")
    try:
        # ... UPDATE content ...
        self._refresh_episodic_embedding(row["id"], row["rowid"], final_content)
        cursor.execute("RELEASE degrade_row")  # Line ~7787
        results["tier1_to_tier2"] += 1
    except Exception:
        try:
            cursor.execute("ROLLBACK TO degrade_row")  # Line ~7791
            cursor.execute("RELEASE degrade_row")
        except Exception:
            logger.info("Regex extraction failed, skipping", exc_info=True)
```

The issue occurs when:
1. `SAVEPOINT degrade_row` is created successfully
2. The UPDATE succeeds but `_refresh_episodic_embedding()` raises an exception
3. The savepoint state becomes inconsistent (possibly due to implicit transaction commit/rollback during the exception)
4. `ROLLBACK TO degrade_row` fails because the savepoint no longer exists in the expected state

## Checklist

- [x] I can reproduce this bug consistently (every degradation with embedding failure triggers it)

dplush:
## Summary

Fixes #691. sqlite-vec episodic embedding refreshes no longer commit the transaction that owns `SAVEPOINT degrade_row`.

`_vec_insert()` retains its existing default commit behavior. Only the refresh performed from `degrade_episodic()` passes `commit=False`, leaving the existing outer per-row savepoint and final commit as the transaction owner.

## Why this path

The premature commit occurs in `_vec_table_insert()` after a successful sqlite-vec insert. Passing `commit=False` only at the degradation call site preserves all other existing callers and makes the content, tier, and vector mutation atomic again.

## Non-goals

- No schema or public API changes.
- No change to other vector-write commit behavior.
- No migration of existing data.

## Verification

- `python3 scripts/verify-docs.py` → passed
- `.venv/bin/python -m pytest -q tests/test_degrade_vector.py tests/test_plugin_manifest_versions.py::test_all_plugin_manifests_have_an_explicit_version_contract` → 8 passed
- `.venv/bin/ruff check mnemosyne/core/beam.py tests/test_degrade_vector.py` → passed
- `git diff --check` → clean

New regression coverage uses the real sqlite-vec extension and verifies both a successful degradation refresh and rollback of content, tier, and vector payload after an induced post-insert failure.

## Release note

Added a concise Unreleased / Fixed entry for #691. The release version remains owned by the release process.


dplush:
@AxDSan, this is ready for maintainer review.  
  
 The narrow fix keeps _vec_insert()’s existing default behavior and opts out of the inner commit only for degradation refreshes, so the existing per-row savepoint remains the transaction owner. Full CI, docs generation, CLA, and the sqlite-vec regression coverage are green on the current head.  
  
 CodeRabbit suggested precomputing embeddings before the savepoint to shorten the write-lock window. That is a worthwhile performance follow-up, but it changes the degradation transaction and failure boundary, so I propose keeping it separate from this correctness fix.  
  
 Since this touches BEAM transaction behavior, I wanted to flag it before merging.

PR #704, opened 2026-08-11
fix: warn before uncached local model download

### LINKED ISSUE #703: [BUG] First consolidation downloads a 656 MB model with no notice
## Description

Follow-up to #617, which #686 closed. That PR fixed the documentation side: the `MNEMOSYNE_LLM_ENABLED` default now reads `true`, the fallback order is spelled out, and the download is disclosed with its size and cache location. The code side is unchanged.

The issue is that `_download_model()` (`mnemosyne/core/local_llm.py:103`) calls `hf_hub_download` with no message. A user who has not read `docs/configuration.md` gets an unannounced 656 MB transfer on their first consolidation, and because `sleep()` is synchronous, the call simply blocks until it completes with nothing indicating why.

When triaging #617, the first fix was scoped as three things:

1. decide and document the default contract,
2. document the fallback order and cache location, and
3. "emit a clear pre-download notice naming the model and approximate size".

#686 delivered the first two. This issue covers the third.

When the download fails, `_get_llm()` catches every exception from `_download_model()` and sets `_llm_available = False` (`local_llm.py:185-188`), so a user on a metered or offline connection sees consolidation silently degrade to AAAK with no indication that a download was attempted at all.

## Reproduction Steps

1. Start from a machine with no cached model (`~/.hermes/mnemosyne/models` absent or empty).
2. Leave `MNEMOSYNE_LLM_ENABLED` unset, so it defaults to `true`.
3. Trigger a consolidation, e.g. `mnemosyne sleep` or let auto-sleep fire.

## Expected Behavior

Before the fetch begins, a message naming the model (`MiniCPM5-1B-Q4_K_M.gguf` from `openbmb/MiniCPM5-1B-GGUF`), its approximate size, the destination directory, and how to avoid it (`MNEMOSYNE_LLM_ENABLED=false` for AAAK-only consolidation, or pre-caching the GGUF).

## Actual Behavior

No output. The process blocks for the length of the transfer, then continues. If the transfer fails, consolidation quietly falls back to AAAK.

## Notes

Decisions:

- Where should the notice go? `local_llm.py` is library code reached from the MCP server, the CLI, and the Hermes provider, so `print()` may not be right for every caller and a logger may not be visible to the one that matters.
- Should a first download be opt-in rather than announced? That is a larger behavior change than a notice, and I'll consider it outside the scope of this issue.

I am happy to open a PR for the notice given direction the first point.

## Checklist

dplush:
## Summary

Fixes #703 with a single warning before an uncached local GGUF download starts.

- Logs the effective model filename, HuggingFace repository and destination cache path.
- Includes the approximate 656 MB size only for the built-in default artifact.
- Explains the AAAK-only opt-out (`MNEMOSYNE_LLM_ENABLED=false`) and pre-cache option.
- Keeps all default, cache, download, retry and fallback behavior unchanged.
- Does not write to CLI/MCP stdout.

## Verification

- `91 passed` — focused local-LLM and extraction suite.
- Immutable archive run at `05e7c70c50ade4bda1988aaa7cb7d4dfacfc9732`: `91 passed`, integrity pass.
- Covers uncached warning-before-download, cached no-warning/no-download, failed-download fallback, effective override wording, stdout silence, and recorded download kwargs.
- `git diff --check` passes.

## Scope

No opt-in/default change, cache relocation, backup-policy change, fallback-routing change, docs, version bump, or new configuration surface.

## Summary

- Adds a warning before an uncached local GGUF model download.
- Reports the model filename, HuggingFace repository, cache path, approximate size, and blocking behavior.
- Preserves `MNEMOSYNE_LLM_ENABLED=false` and pre-cache guidance.
- Preserves cache behavior, retries, fallback to AAAK, stdout behavior, and defaults.
- Adds focused tests with 91 passing tests, integrity checks, and `git diff --check`.

## Architectural impact

- **Sync layer:** No changes.
- **Agent integration:** No changes to Hermes, MCP, or CLI interfaces.
- **Privacy and local-first design:** Improves transparency for the existing local model download. It adds no telemetry or remote inference. The local-model opt-out remains available.
- **Maintainability:** Keeps download messaging in `_download_model()` and adds focused behavioral tests.

This is the right call for the issue. It improves observability without changing model selection, fallback routing, opt-in behavior, or local-first guarantees. It does not alter memory architecture or agent integration surfaces.

### DIFF
--- a/mnemosyne/core/local_llm.py
+++ b/mnemosyne/core/local_llm.py
@@ -124,6 +124,7 @@ def _download_model() -> Path:
     size_notice = " (approximately 656 MB)" if default_artifact else ""
     logger.warning(
         "Downloading local LLM model %s from %s%s to %s. "
+        "The current operation will block until the download completes. "
         "Set MNEMOSYNE_LLM_ENABLED=false for AAAK-only consolidation, "
         "or pre-cache the GGUF to avoid this download.",
         DEFAULT_MODEL_FILE,
--- a/mnemosyne/core/local_llm.py
+++ b/mnemosyne/core/local_llm.py
@@ -9,6 +9,7 @@ Model cache: ~/.hermes/mnemosyne/models/
 Default model: openbmb/MiniCPM5-1B-GGUF (Q4_K_M, ~656MB)
 """
 
+import logging
 import os
 import sys
 import re
@@ -19,6 +20,8 @@ from typing import List, Optional
 DEFAULT_MODEL_REPO = "openbmb/MiniCPM5-1B-GGUF"
 DEFAULT_MODEL_FILE = "MiniCPM5-1B-Q4_K_M.gguf"
 MODEL_CACHE_DIR = Path.home() / ".hermes" / "mnemosyne" / "models"
+
+logger = logging.getLogger(__name__)
 
 LLM_ENABLED = os.environ.get("MNEMOSYNE_LLM_ENABLED", "true").lower() in ("1", "true", "yes")
 LLM_MAX_TOKENS=int(os.environ.get("MNEMOSYNE_LLM_MAX_TOKENS", "2048") or "2048")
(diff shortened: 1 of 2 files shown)


dplush:
Scope clarification: this PR adds observability only. It logs before an uncached local-model download starts; it does not pause, prompt, or require user consent, and it leaves the current enabled-default and fallback behavior unchanged.

A possible follow-up can evaluate an explicit opt-in / download gate, including the non-interactive CLI, MCP, and service compatibility trade-offs. That is intentionally separate from this narrow notice fix.

tonythomson:
One suggestion: the message says what is being fetched and where it lands, but not that the caller is stopped while it happens. Users experience the resulting hang. Something like "the current operation blocks until the download finishes" after the destination would cover it.

dplush:
Thanks for the helpful suggestion, Tony. Added the blocking behavior to the pre-download warning and covered it in the focused notice regression: fc2d58e.

dplush:
@coderabbitai Thanks for the additional coverage suggestions. I verified the current boundaries: this PR only adds the pre-download warning; it does not change `_load_llm()` failure handling or the Beam/Sleep AAAK fallback path.

The focused test proves that an uncached custom-model download failure still leaves the local LLM unavailable, while existing Beam/Sleep coverage exercises the caller-level AAAK fallback when `llm_available()` is false.

The two partial override combinations would be useful broader test-matrix coverage for the pre-existing default-artifact size condition, but are not required to establish the new warning behavior. I’d keep this PR limited to the notice and track that expanded coverage separately.

PR #705, opened 2026-08-11
fix(beam): exclude dialog sources from working-memory vec pool (#696)

### LINKED ISSUE #696: Working-memory vec pool flooded by raw dialog rows — real facts starved out of recall candidates (dense_score=0 despite valid embedding)
## Problem

Working-memory vector recall can silently **starve real facts out of the candidate pool**: an operational instruction with `importance=0.7` and a valid embedding (cosine ≈ 0.70 to the query) gets `dense_score=0.0` and a hybrid score of ~0.10 — below any injection threshold — because the top-k of the vec search is entirely occupied by raw dialog rows.

Concrete case (mnemosyne-memory 3.14.0, Hermes provider, single-user DB with auto-captured conversation rows, `source='conversation'`):

- Query (RU, short, conversational): *"В Kuma пока нет, добавь. Он перегружается очень быстро, посмотри как настроить чтобы не пропустить."* (add a Uptime-Kuma monitor)
- The relevant memory: *"Creating Uptime Kuma monitors is CLI-only: ~/workspace/kuma-tools/kuma.py (own venv with python-socketio)…"* — `importance=0.7`, valid `vec_working` int8[384] embedding, float cosine to query ≈ 0.703.
- **Ranking of that memory in the full vec search: #204 of 4096.** Of the 204 rows ranked closer: 175 were `source='conversation'` (raw `[USER]`/`[ASSISTANT]` dialog rows about the same topic), 18 `honcho:*` imports, 6 orphans, ~5 actual facts.
- Result: `_wm_vec_search_sqlite(..., k=50)` never returns the row → `dense_score=0.0` → score 0.10 → the Hermes prefetch filter drops it. Meanwhile a *stale* record («python-socketio is NOT installed», true at write time, false 1h later) scored 0.38 and got injected, steering the agent to the wrong tool (browser instead of CLI).

## Root cause

Three compounding factors in `_wm_vec_search_sqlite` (beam.py):

1. **The vec pool is selected with no source/importance awareness.** `k` for working-memory recall is small (`max(top_k*3, 50)` with the Hermes prefetch `top_k=16` → k=50) and the nearest-N is dominated by raw dialog rows, which are numerous, recent and topically identical to the query.
2. **sqlite-vec applies extra `WHERE` clauses as a post-filter on the k nearest rows.** So a naive `AND wm.source <> 'conversation'` exclusion *after* `MATCH ... AND k=50` leaves ~1 row (only 1 of the 50 nearest was a fact). The candidate pool size has to account for the post-filter.
3. **Working-memory dense is only a 20% blend into a keyword-first score** (`base_score = base*0.80 + vec_sim*0.20`), and the row must still clear a lexical relevance gate. A row that only missed the vec pool gets `dense_score=0.0` and is scored on FTS/keyword alone.

## Proposed fix (2 lines, verified)

In `_wm_vec_search_sqlite`, exclude dialog rows from the vec candidate pool **and** search a larger pool so the post-filter semantics of sqlite-vec still leave enough facts:

```python
k_search = max(k, 1500)   # WHERE is a post-filter on the k nearest; pool must be large
# in each MATCH branch's WHERE:
AND wm.source <> 'conversation' AND wm.source NOT LIKE 'honcho:%'
```

The FTS path is untouched, so dialog rows remain retrievable lexically.

Measured effect on the failing query (prefetch pipeline, plugin filters identical):

| | before | after |
|---|---|---|
| core recall results | 3 | 8 |
| CLI instruction dense_score | 0.0 | 0.88 |
| CLI instruction in injected top-5 | no | yes |
| stale «not installed» record | injected | gone |

Vec search latency at k=1500 on 5.6k int8[384] rows: ~60 ms (same as k=50) — negligible.

## Related

- #615 (closed): weak-lexical-overlap **injection** filter — our case is upstream of it: the facts never reach the injection filter.
- #159, #224 (closed): store-time embedding bugs (`dense_score always 0.0`) — different layer (write path), same observable symptom.

## Bonus observation

The default embedding model `BAAI/bge-small-en-v1.5` has very weak topic discrimination on non-English content: same-topic vs cross-topic cosine gap measured **0.044** on Russian (query about Uptime Kuma vs unrelated facts about Xray/TLS certs both ~0.68–0.73). `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` (same dim 384, works with the existing `int8[384]` schema) gives a gap of **0.383** — 9× better separation. After switching + `reindex`, the injected top-5 for the failing query became 5/5 topically relevant. Worth considering a multilingual default for non-English deployments.

Happy to turn the two-line fix into a PR (patch is applied and regression-tested on a production DB locally).

Goonik:
## Summary

Fixes #696. Raw dialog capture (`source='conversation'`, legacy `honcho_*` imports) is now excluded from the **working-memory dense candidate pool**, while remaining fully reachable through the FTS path.

Conversational queries are topically identical to their own dialog rows, so those rows saturate the nearest-N vector pool and starve distilled facts out of the dense voice — a fact can rank beyond the pool and surface with `dense_score=0.0` or not at all. `#608` already widened the candidate neighborhood (`scan_k`), which helps shallow floods; this change is what makes the pool capacity effective against the dialog flood itself and fixes the rank-beyond-pool cases (#608 alone cannot: if the 500 nearest rows are dialog, facts still never enter).

## Changes

`mnemosyne/core/beam.py` — `BeamMemory.recall()`:

- Builds a separate `wm_vec_where` for the vector search only: the recall `where_sql` gets `AND (source IS NULL OR (source <> 'conversation' AND source NOT LIKE 'honcho%'))` appended.
- The final row fetch keeps the original `wm_where`, so FTS hits on dialog rows are unaffected — `recall("what did we discuss")` still returns them.
- The exclusion is skipped when the caller passes an explicit `source=`/`topic=` filter (e.g. `recall(source='conversation')` still works).

The same `where_sql` flows into both the sqlite-vec path and the `memory_embeddings` compatibility scan, so both backends behave identically.

## Tests

`tests/test_beam_wm_vec_dialog_exclusion.py` — four tests:

1. `test_dialog_flood_does_not_starve_fact_from_vec_pool` — reproduces #696: 560 `conversation` rows share the query's only real token (FTS returns the flood and ranks the fact beyond its top-k), and every dialog embedding is closer to the query than the fact's. Asserts the fact reaches recall via the vector voice. **Fails on unpatched code** (fact absent), passes with the fix.
2. `test_dialog_flood_pure_vector_case_recall_first_mode` — same flood under `MNEMOSYNE_LEXICAL_GATE_MIN=0` (recall-first): a lexically invisible fact is admitted via the dense voice alone, and no dialog row leaks into results. **Fails on unpatched code**, passes with the fix.
3. `test_wm_vec_search_excludes_dialog_sources` — pins the exact WHERE semantics on the vec search: `conversation` and `honcho_*` rows are dropped, `NULL`-source and distilled rows kept.
4. `test_explicit_conversation_source_filter_still_returns_dialog` — `recall(..., source='conversation')` still surfaces dialog rows (exemption contract).

All tests use synthetic embeddings — no model download, deterministic, fast.

## Verification

- New tests: 4 passed with the fix; tests 1 and 2 fail against the unpatched `recall` (verified via `git stash`).
- Related suites: `tests/test_beam.py`, `tests/test_e5a_vector_voice_dense_rewire.py`, `tests/test_prefetch_profiles.py` — 122 passed, 14 skipped (sqlite-vec-extension tests skipped: not installed in the scratch venv). No regressions.

**Scope note**: consolidated-row exclusion (plus its polyphonic parity) is not part of #696 itself — it implements the existing `get_context` invariant (issue #427: consolidated rows must not compete with hot working-memory rows) for the dense candidate pool, and is tested independently (`test_default_vec_predicate_excludes_consolidated_rows`, `test_polyphonic_path_applies_default_dense_source_filter`).

## Review follow-up (CodeRabbit)

- **Polyphonic path (Major)**: `PolyphonicRecallEngine._vector_voice` now applies the same default dense-source predicate (dialog/honcho exclusion + consolidated rows) to its working-memory tier **before** top-K selection. `BeamMemory._recall_polyphonic` passes `default_dense_source_filter=not (source or topic)` so explicit source=/topic= requests are never pre-empted. Regression: `test_polyphonic_path_applies_default_dense_source_filter`.
- **Consolidated rows (Major)**: `consolidated_at IS NULL` added to the default linear vec predicate (mirrors `get_context`, invariant #427), bypassed for explicit source=/topic=. Regression: `test_default_vec_predicate_excludes_consolidated_rows`.
- **Coverage (Minor)**: added recall-level tests for explicit `topic=` bypass, FTS-only default recall returning conversation rows, and consolidated-row exclusion; existing sqlite-vec-path tests skip gracefully when the extension is unavailable.

Full suite: 8 new tests pass (2 fail pre-fix); related suites 122 passed / 14 skipped, no regressions.

## Summary

- Working-memory dense retrieval excludes raw `conversation` and `honcho_message` rows by default.
- Durable `honcho_summary` and `honcho_import` rows remain eligible.
- Consolidated rows remain excluded from dense candidate selection.
- FTS retrieval remains unchanged.
- Explicit `source` and `topic` filters apply exact predicates before top-K selection.
- SQLite-vector and polyphonic retrieval use consistent filtering.
- Enhanced-recall cache keys now use `v3:` semantics.
- Cache entries invalidate after consolidation, orphan reclamation, degradation, and sleep mutations.

## Architectural impact

This change protects working-memory dense recall from raw dialog flooding. It preserves access to factual memories and lexical dialog recall.

The change improves retrieval within the working-memory tier. It does not change episodic memory, BEAM behavior, the veracity system, or the sync layer.

Cache invalidation now follows consolidation-related mutations. Failed enrichment paths also invalidate affected cache entries.

This is the right call for predictable retrieval and cache correctness.

Retrieval remains local. The PR adds no external services, telemetry, data movement, or retention changes.

The PR does not erode local-first guarantees. It reduces raw-dialog dominance in dense retrieval without removing lexical access to dialog content.

## Integration surfaces

`PolyphonicRecallEngine.recall()` now accepts `default_dense_source_filter`, `source`, and `topic`.

No Hermes, MCP, or CLI behavior changes are reported.

Shared filtering rules align SQLite-vector and polyphonic retrieval. Exact predicates before top-K selection make retrieval behavior predictable.

The `v3:` cache schema separates current entries from legacy `v2:` entries. Updated test doubles support the expanded recall contract.

Regression tests cover dialog flooding, FTS retrieval, explicit filters, durable Honcho rows, consolidated-row exclusion, polyphonic predicates, topic equality, cache isolation, and cache invalidation.

Reported validation includes green CI checks and 121 passes across touched suites. Full-suite failures were attributed to missing dependencies and filesystem limitations. No benchmark methodology changes are reported.


Goonik:
Addressed CodeRabbit's review in commit `7dcdbdf` — all three items are handled:

1. **Major — polyphonic path**: `PolyphonicRecallEngine._vector_voice` now applies the default dense-source predicate (dialog/honcho exclusion + consolidated rows) to its working-memory tier before top-K selection. `BeamMemory._recall_polyphonic` passes `default_dense_source_filter=not (source or topic)` so explicit `source=`/`topic=` requests are not pre-empted. Regression test: `test_polyphonic_path_applies_default_dense_source_filter`.
2. **Major — consolidated rows**: `consolidated_at IS NULL` added to the default linear vec predicate (mirrors `get_context`, #427), bypassed for explicit filters. Regression test: `test_default_vec_predicate_excludes_consolidated_rows`.
3. **Minor — coverage**: added recall-level tests for explicit `topic=` bypass and FTS-only default recall returning conversation rows.

Results: 8/8 new tests pass (tests 1–2 fail on the unpatched linear path), related suites 122 passed / 14 skipped, no regressions.

Goonik:
Round 2 follow-up (commit `94ba6e3`) — both new findings addressed, plus the CI failure:

## CI failure fixed
`test (3.10)` failed: 6 tests in `test_e3a3_cross_tier_dedup.py` + `test_pre_experiment_fidelity.py` mock the polyphonic engine with fakes whose `recall()` signature didn't accept the new `default_dense_source_filter` kwarg → `TypeError` was swallowed by the engine-failure guard → empty results. The fakes now mirror the updated engine contract (`default_dense_source_filter`, `source`, `topic`). I had missed those two suites in my earlier adjacent-suite selection — this time the full suite was run locally before pushing.

## Inline finding (valid — fixed)
Explicit `source=`/`topic=` now propagates into the engine as a real predicate, not just a disable-switch: `PolyphonicRecallEngine.recall`/`_vector_voice` accept `source`/`topic` and apply `wm.source = ?` (same equality semantics as the linear `wm_where`, including topic-in-source-column) **before** top-K selection. Without this, 30 closer non-conversation rows would occupy the vector voice's top-20 slice and starve the requested dialog row out of the pool before the post-engine filter. Regressions: `test_polyphonic_explicit_source_predicate_applied_before_topk`, `test_polyphonic_explicit_topic_predicate_applied_before_topk` (both fail without the predicate).

## Outside-diff finding (valid — fixed)
`recall_enhanced()` opaque cache schema bumped `v2:` → `v3:` in `_enhanced_recall_cache_key` — the dense candidate predicate changed, so pre-change entries can still contain dialog/honcho/consolidated candidates. `QueryCache._OPAQUE_V2_KEY_RE` now accepts both prefixes so v3 keys keep the exact-only opaque path and stale v2 keys never degrade to fuzzy lookups. Regression: `test_legacy_v2_opaque_entry_not_reused_after_dense_predicate_schema_bump` plants a v2 entry before any v3 entry exists and proves it's never reused; the pinned-schema test was renamed/updated to v3.

## Results
- All touched files: **117 passed** (10 dialog-exclusion incl. 4 new polyphonic, enhanced-cache suite incl. new regression, e3a3, pre_experiment_fidelity, engine suite).
- Full local suite: **2592 passed, 42 skipped, 33 failed** — every failure is an environment gap of the scratch venv (missing `mcp` and `SyncEncryption` deps → `test_mcp_server`/`test_sync*`; `linkat(AT_EMPTY_PATH)` unsupported on this filesystem → `test_repair`), confirmed identical on the base commit; zero failures in changed code.

Goonik:
Round 3 (commit `70d522f`) — the Minor finding is addressed:

- `_FakeEngine.recall()` now records `default_dense_source_filter` alongside `top_k`/`source`/`topic`.
- The overscan test asserts `True` for default recall (plus `source=None`/`topic=None`).
- New test `test_explicit_source_topic_filters_propagate_to_engine_with_filter_disabled` asserts `False` + the propagated `source=`/`topic=` values for explicit recalls.

All 7 CI checks green on the latest commit (test 3.10–3.13, lint, docs-check, build); e3a3 suite 27 passed, related files 40 passed.

dplush (changes requested):
Thanks for the thorough work on this important recall-path fix. The linear/polyphonic pre-top-K handling and cache invalidation direction are strong, but I found two correctness blockers before this can merge:

1. **Keep durable Honcho content in the dense pool.** The default `source NOT LIKE 'honcho%'` predicate excludes `honcho_summary` and `honcho_import` along with raw `honcho_message` rows. The importer treats `honcho_summary` as a deliberate session summary with higher importance, and `honcho_import` is a generic import default, so neither is reliably raw dialog. This reverses the goal of #696 for those durable records: they can no longer receive a dense score in default recall. Please narrow the exclusion to the actual raw-dialog sources, at minimum `conversation` and `honcho_message`, in both linear and polyphonic paths. Add positive default-dense regressions showing that `honcho_summary` and `honcho_import` remain eligible.

2. **Make the polyphonic `topic=` post-filter exact.** The linear path implements `topic` as `source = ?`, and the new polyphonic vector predicate does the same before top-K. But `_polyphonic_row_passes_filters()` still accepts a substring, so a non-vector voice can return `source='conversation_archive'` for `topic='conversation'`. Please switch this final check to equality and add that regression, so all voices preserve the documented linear semantics.

Once these two points are addressed, I’m happy to re-review the updated head.

Goonik:
Both blockers addressed in commit `02bab94`:

## 1. Durable honcho rows stay in the dense pool
The default exclusion is narrowed from `source NOT LIKE 'honcho%'` to exactly the raw dialog sources — `source <> 'conversation' AND source <> 'honcho_message'` — in **both** the linear `wm_vec_where` and the polyphonic vector voice's working-memory tier. `honcho_summary` / `honcho_import` remain eligible for a dense score in default recall.

Regressions (all fail on the pre-fix predicate):
- `test_wm_vec_search_excludes_dialog_sources` — pinned WHERE now admits honcho_summary/honcho_import, drops conversation/honcho_message;
- `test_default_dense_recall_keeps_durable_honcho_rows_eligible` (linear, recall-first mode);
- `test_polyphonic_default_dense_keeps_durable_honcho_rows_eligible` (polyphonic; QUERY has no temporal keywords so the temporal voice is inert and the vector voice alone decides).

## 2. Polyphonic `topic=` post-filter is exact
`_polyphonic_row_passes_filters` now matches `row_dict["source"] != topic` (equality), identical to the linear path's `source = ?` semantics. Regression: `test_polyphonic_topic_filter_exact_equality` — a fake engine returns a row with source `conversation_archive` for `topic='conversation'` and recall excludes it (pre-fix substring match admitted it).

Also folded in the CodeRabbit Minor from the previous round while touching the cache tests: the opaque-cache regression now proves both the legacy v2 entry and the v3 hit survive full SQLite reloads, and that v3 keys route to `get_opaque()` without touching the fuzzy tiers.

Results: all touched suites **121 passed**; the three new behavioral regressions fail on the pre-fix core and pass with the fix. CI is re-running on the new head.

Goonik:
Hi @dplush — gentle ping on your CHANGES_REQUESTED review (Aug 12, 06:22Z). Both blockers were addressed in `02bab94` and have since been verified:

1. **Durable honcho rows stay in the dense pool** — default exclusion narrowed to `source <> 'conversation' AND source <> 'honcho_message'` in both the linear `wm_vec_where` and the polyphonic vector voice's working-memory tier. `honcho_summary` / `honcho_import` remain eligible for a dense score in default recall. Regressions: `test_wm_vec_search_excludes_dialog_sources`, `test_default_dense_recall_keeps_durable_honcho_rows_eligible`, `test_polyphonic_default_dense_keeps_durable_honcho_rows_eligible`.
2. **Exact `topic=` post-filter** — `_polyphonic_row_passes_filters` now uses equality (`row_dict.get("source") != topic`), mirroring the linear `source = ?` semantics; regression `test_polyphonic_topic_filter_exact_equality` covers the `conversation_archive` / `conversation` case.

Branch is rebased onto current main (`6005bae`, CI 7/7 green, CodeRabbit re-review of `02bab94` clean). Happy to adjust anything else you spot on the updated head — thanks again for the careful review!

Goonik:
recheck

(CLA was signed and confirmed on Aug 11; this branch has since been merged with upstream main, which introduced new committers into the PR history — asking for a re-evaluation on the current head `6005bae`.)

Goonik:
Addressed the new CodeRabbit finding on the current head (`c51cab4`):

**Cache invalidation after consolidation mutations** — `consolidate_to_episodic()`, `reclaim_orphans()`, and `sleep()` (after the claim commit and after the final summary/degradation commit) now call a best-effort `_invalidate_query_cache_after_commit()` so warmed enhanced-recall v3 entries are never served stale once dense-pool eligibility changes.

Regressions (all 3 fail on the pre-fix code):
- `test_consolidate_to_episodic_invalidates_warmed_v3_cache` — warms an entry, consolidates, asserts the cache version bumps and the next request recomputes (proven via a recall-call counter, not just version).
- `test_reclaim_orphans_invalidates_warmed_v3_cache` — plants a stale claim, reclaims, asserts invalidation.
- `test_sleep_invalidates_warmed_v3_cache` — consolidates an old row via `sleep()`, asserts invalidation.

Local: 216 passed / 10 skipped across the adjacent suites. CI is re-running on the new head.

dplush (approved):
Reviewed the current head `0567fe7` against #696 and the earlier findings.

The default dense filter now excludes only raw dialog, raw Honcho messages, and consolidated working-memory rows before top-k selection, while FTS and explicit `source`/`topic` recall remain available. Linear and polyphonic paths apply the same pre-limit contract.

The v3 enhanced-recall cache boundary and post-mutation invalidation are covered, including consolidation enrichment failure paths. The latest test-only follow-up is green across CI and CodeRabbit generated no actionable comments. Earlier request-changes findings are addressed on this head.

Approved.

dplush:
@AxDSan Could you please take a final review and merge decision on this one?

I reviewed and approved the current head `0567fe7`; CI, CLA, and CodeRabbit are green. The #696 fix now has linear/polyphonic pre-top-k parity, keeps FTS and explicit `source`/`topic` recall intact, and includes the cache-schema/invalidation regressions.

Because it also changes enhanced-recall cache invalidation around consolidation, I’d prefer your final sign-off before merging. Thanks!

dplush:
Please rebase.

Goonik:
Thanks for the careful review — here's how each finding is handled on the new head (`a6ef3a2`):

## 1. Invalidate the cache inside `degrade_episodic()` — ✅ fixed

Moved `_invalidate_query_cache_after_commit("degrade_episodic")` into `degrade_episodic()` right after its commit, so direct calls and the `sleep_all_sessions()` degradation pass invalidate enhanced-recall entries too. `sleep()` keeps its own invalidation for its other mutations (claim + summaries). Regression `test_degrade_episodic_invalidates_warmed_v3_cache` warms an entry, degrades directly, and proves version bump + recompute via an instrumented `memory.recall` counter (fails on the pre-fix code).

## 2. Complete polyphonic filter contract (date/scope/identity) — pre-existing, out of this PR's diff

This finding is outside the diff range (as the review notes). On the base commit `f6ba5c8` the polyphonic call was `engine.recall(query=query, query_embedding=query_embedding, top_k=top_k*2)` — **no** filters were forwarded at all. This PR deliberately narrows its scope to the #696 dialog/consolidation exclusion and, at your earlier request (round 2), added the explicit `source`/`topic` predicates so those filters apply before top-K selection. Completing the contract for `from_date`/`to_date`/`scope`/identity is a distinct, valuable change (to_date boundary semantics included) — I'd rather track it as a separate issue than expand this already-reviewed PR. Happy to file it.

## 3. Episodic entity/fact supplements bypass `em_where` — pre-existing, out of this PR's diff

Verified against base `f6ba5c8`: the entity/fact supplement queries at the base already query `FROM episodic_memory em` without joining `em_where` (lines ~8545/8586 on base). This PR does not touch those paths and does not regress them. Same suggestion — worth its own issue/PR so it gets a dedicated regression suite.

Local: 255 passed / 14 skipped across the adjacent suites, lint clean. CI re-running on `a6ef3a2`.

dplush:
Thanks for the detailed breakdown. I agree that the two broader filter-contract gaps should not be folded into this already large #696 PR.

For this PR, please keep the contract explicit and narrow:

- default raw-dialog/consolidated-row exclusion applies to dense working-memory candidates;
- explicit `source` / `topic` predicates reach the linear and polyphonic vector selection before their candidate limits;
- FTS behavior remains unchanged.

Please do not present this as full linear/polyphonic filter parity or complete caller-filter enforcement while date, identity, scope, veracity/type, and the entity/fact episodic supplements still have the pre-existing gaps.

Two small follow-ups would be the right place:
1. full polyphonic eligibility propagation before top-k, including inclusive `to_date`;
2. reuse the complete episodic eligibility predicate for entity/fact supplements.

For the cache fix, please strengthen the direct-degradation regression so it actually seeds an old tier-2 episodic row and asserts the tier-2 → tier-3 transition before checking cache invalidation. The current fixture only creates working memory, so it does not prove the mutation occurred.

Finally, the branch is now behind current `main` after #750. Please rebase, keep the scope clarification/follow-up links in the PR, and rerun CI. Then we can do the fresh final review and request @AxDSan signoff.

Issue #708, opened 2026-08-12
[FEATURE] Make the model cache location configurable


tonythomson:
## Problem

Follow-up to #617, which #686 closed. Two concerns in that report were deferred as follow-ups but weren't filed before it closed as completed. Both originate in the fact that the 656 MB GGUF lives somewhere the user cannot change, inside a tree that gets archived.

**The cache location is hardcoded.** `MODEL_CACHE_DIR` is `Path.home() / ".hermes" / "mnemosyne" / "models"` (`mnemosyne/core/local_llm.py:21`), with no environment or config override. `MNEMOSYNE_LLM_MODEL_REPO` and `MNEMOSYNE_LLM_MODEL_FILE` already let you change *which* model is fetched (`local_llm.py:32-33`); there is no equivalent for where it lands. On a small home partition, or with `$HOME` on a network mount, the only remedy is a symlink.

**It is included in backups.** I checked the exclusion list in Hermes (`hermes_cli/backup.py:51`): `_EXCLUDED_DIRS` covers `venv`, `site-packages`, `.cache`, `__pycache__` and similar, on the stated rationale that they are "regenerated by reinstalling; never irreplaceable state". There is no `models` entry and no `.gguf` suffix exclusion. The `.cache/` subdirectory inside the models directory *is* excluded, by name, but the 656 MB file beside it is not. On this machine:

```
656M  ~/.hermes/mnemosyne/models/
      .cache/                          <- excluded
      MiniCPM5-1B-Q4_K_M.gguf  656 MB  <- archived
```

At the default retention of 5 for pre-update and pre-migration archives, that is roughly 3.3 GB of backups of a file that is re-downloadable from Hugging Face and holds no user data. It fits the existing exclusion rationale exactly.

## Proposed Solution

Honour a cache directory override, e.g. `MNEMOSYNE_MODEL_CACHE_DIR`, falling back to the current path. That follows the precedent already set by the repo and filename overrides, and it is the part that belongs in this repo.

The backup exclusion does not belong: that list is Hermes', so a `models` entry would be an upstream change. Worth noting that a configurable cache pointed outside `HERMES_HOME` solves the backup cost without touching Hermes at all, which is an argument for doing this one first and seeing whether the other is still needed.

## Additional Context

Related: #703 covers the missing pre-download notice, the third item from the same triage.
You scoped the cache-path and backup work out of that slice explicitly, which is why this is separate.

## Checklist

[...]

PR #710, opened 2026-08-12
fix(llm): announce the model download before it starts

### LINKED ISSUE #703: [BUG] First consolidation downloads a 656 MB model with no notice

tonythomson:
## Description

The first uncached use fetched the ~656 MB `MiniCPM5-1B-Q4_K_M.gguf` from Hugging Face with no output, blocking the caller for the length of the transfer. #686 documented the download; nothing announced it at the point it happens, so a user who had not read `docs/configuration.md` saw only a hang.

`_download_model()` now logs one warning immediately before an uncached fetch, naming the model, repo, approximate size, destination and the `MNEMOSYNE_LLM_ENABLED=false` AAAK-only opt-out.

Following your spec on the issue, with three details worth calling out:

**A logger, not a print.** This module is reached from the CLI and the MCP server, which own stdout. With no logging configured at all, Python's `lastResort` handler still puts a WARNING on stderr, so the notice stays visible without a logging setup step. Verified rather than assumed.

**The notice does not name its caller.** It is reached from `sleep()` via `_call_local_llm`, but also from `llm_available()` (`local_llm.py:486` calls `_load_llm()`), which callers do not expect to download anything. An earlier draft said "this runs during sleep()", which would have been wrong for that second path.

**The size is quoted only for the built-in model.** `MNEMOSYNE_LLM_REPO` / `MNEMOSYNE_LLM_FILE` can name a file whose size we cannot know without asking the Hub, so an override reads "size unknown for an overridden model" instead. The check compares the current repo and filename against the built-in pair at call time rather than reading a flag set at import, so the notice cannot describe values it is not about to print.

Nothing else changes: a cached model neither warns nor downloads, and a failed download still falls back to AAAK encoding.

Out of scope, per your comment: no default change, no opt-in download, no cache-path or backup work (#708), no fallback-policy change.

## Related Issue

Closes #703

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

Four tests in `TestModelDownloadNotice`, covering the three cases you asked for plus the override:

| test | on this branch | on `main` |
|---|---|---|
| uncached download is announced before it starts | pass | **fail** |
| overridden model does not claim the default size | pass | **fail** |
| cached model neither downloads nor warns | pass | pass |
| failed download still falls back to AAAK | pass | pass |

The last two are guards over pre-existing behaviour, so they pass either way by design; the first two are the actual regressions.

One trap worth flagging for anyone touching this area: `tests/conftest.py:152` has an autouse fixture that replaces `local_llm._load_llm` with `lambda: None`. The AAAK-fallback test was silently asserting against that stub and passing without exercising anything, so it now captures the real function at import time. The conftest documents the escape hatch; it is easy to miss.

Manual verification of the no-logging-configured path:

```
$ python -c "import logging; logging.getLogger('mnemosyne.core.local_llm').warning('...')" >/dev/null
...        # stderr, stdout clean
```

Full suite with `MNEMOSYNE_NO_EMBEDDINGS=1`: 2625 passed, 26 skipped. Both ruff gates clean over the changed files, and `scripts/generate-docs.py --check` reports no drift.

## Checklist

No README change: `docs/configuration.md` already documents the download, its size and cache location as of #686, and this adds no new configuration surface.

No version bump, matching #686 and #620. Say the word if you would rather it carry one.

Note on the first commit: `a7fc0b8` clears nine pre-existing F violations in `tests/test_local_llm.py` (an unused `pytest` import, seven placeholder-less f-strings, one unused `as mock_local` binding). None are mine, but the PR-scoped ruff gate lints whole files, so they became blocking as soon as this branch touched that file. Kept separate so the functional diff stays reviewable.

## Summary

- Adds a pre-download warning for uncached local LLM models.
- Reports the model, repository, approximate size, destination, blocking behavior, and `MNEMOSYNE_LLM_ENABLED=false` opt-out.
- Reports unknown size for overridden models.
- Preserves silent cached-model use and AAAK fallback after download failure.
- Adds focused tests, changelog coverage, and Ruff cleanup.

## Architecture impact

- **Consolidation:** Improves visibility into the local LLM step without changing download behavior or AAAK fallback.
- **Retrieval and veracity:** No changes.
- **Sync layer:** No changes.
- **Agent integrations:** No changes to Hermes, MCP, or CLI interfaces.
- **Privacy and local-first guarantees:** The change strengthens operational transparency while keeping processing local. It adds an explicit AAAK-only opt-out and does not make downloads opt-in or send data to a remote service.
- **Maintainability:** Centralized model metadata, standard Python logging, and regression tests improve maintainability.

This is the right call for issue `#703`. It removes an opaque blocking operation without expanding scope or weakening local-first behavior.


tonythomson:
Closing as a duplicate of #704, which you opened first and which lands the same fix. I misread "I'll take a focused PR" on #703 as an invitation rather than you taking it on, and didn't check the open PR list before starting. Apologies for the noise.

One small suggestion left on #704.

PR #711, opened 2026-08-12
fix(hermes): use the shared environment check in the provider diagnostic

### LINKED ISSUE #709: [BUG] Two venvs sharing a base interpreter compare as the same runtime
## Description

Follow-up to #623. CodeRabbit raised this during that review; it was scoped out as a narrow follow-up.

`register_memory_provider()` compares `_hp.resolve()` against `Path(sys.executable).resolve()` to decide whether Hermes runs from a different interpreter than Mnemosyne. A venv's `bin/python` is a symlink to the base interpreter it was created from, so `resolve()` collapses both sides onto that base. Two distinct environments built from one Python therefore compare equal, and the version diagnostics behind that branch are skipped for the case they exist to report.

Still present on current `main` at `integrations/hermes/src/mnemosyne_hermes/__init__.py:3273`.

This is the same defect #623 fixed inside `_find_hermes_python()`, but in a different file. The reasoning there applies unchanged: running a venv's path activates its site-packages, running the resolved base path does not, so the resolved form does not identify the runtime.

## Reproduction Steps

1. Create two venvs from the same base interpreter:

   ```bash
   python3 -m venv /tmp/a
   python3 -m venv /tmp/b
   ```

2. Confirm they collapse to the same path:

   ```bash
   python3 -c "from pathlib import Path; print(Path('/tmp/a/bin/python').resolve()); print(Path('/tmp/b/bin/python').resolve())"
   ```

3. With Hermes running from one and `mnemosyne-hermes` installed in the other, the
   comparison treats them as one runtime.

## Expected Behavior

Two different virtual environments are two different runtimes, and the version diagnostics
report accordingly.

## Actual Behavior

They compare equal and the diagnostics are skipped.

## Notes

The fix is presumably to compare the paths as selected rather than resolved, matching what `run_install()` now does. Worth confirming that nothing downstream depends on the resolving behaviour to paper over a legitimately equivalent case, e.g. a launcher symlink where both sides genuinely name one environment.

Related: `mnemosyne_hermes/install.py` also reports "Python version MISMATCH" in `status` while actually comparing paths, which is a separate wording problem in the same area. Happy to fold it in here or file it apart, whichever you prefer.

## Checklist

- [ ] I can reproduce this bug consistently

tonythomson:
Closes #709.

Rebased onto `main` after #751 landed, and reduced to what that PR did not cover.

## What is left of #709

#751 fixed the two `status` call sites by introducing `_hermes_python_mismatch()`, which compares environment roots rather than resolved interpreter paths. The third site reported in #709, the provider's failure diagnostic in `register_memory_provider()`, was not part of that change and still read:

```python
if _hp and _hp.resolve() != Path(_sys.executable).resolve():
```

A venv's `bin/python` is a symlink to the interpreter it was created from, so resolving collapses two distinct environments onto that one binary and skips the diagnostic in exactly the case it exists to report. On macOS it also rewrites `/tmp` to `/private/tmp`, so one environment can fail to match itself across spellings.

That site now calls `_hermes_python_mismatch()`, so the provider diagnostic and `status` answer the question the same way instead of drifting apart again.

## Hardening the shared helper

Deriving the environment root with `.parent.parent` before normalising means a path spelled `<venv>/bin/../bin/python` yields `<venv>/bin/..`, which names `<venv>` but does not compare equal to it, so one environment is reported as two. Both sides are now passed through `os.path.normpath` first. The normalisation is lexical and does not follow symlinks, so venv identity is preserved; following the symlink is the original defect.

Reachable through a `VIRTUAL_ENV` containing `..`. Narrow, but it is the same defect class as #709 and the helper now governs three call sites.

## Remediation quoting

The provider diagnostic prints a `FIX: Run:` command containing the interpreter path. #751 shell-quoted the equivalent `status` line; this one was still raw, so a Hermes venv under a path with spaces produced an unrunnable command. Now quoted with `shlex.quote`, matching `status`. Raised by CodeRabbit.

The `uv pip install --python {hermes_python}` hints in `run_install` have the same problem and are deliberately left alone here, since #752 covers those.

## Verification

Both new tests were confirmed red with the source change stashed:

- `test_provider_diagnostic_reports_two_venvs_over_one_base` fails without the fix. It patches `sys.executable` as well as `sys.prefix`, both to the other venv. Without that, the superseded check compares against the real test interpreter and prints for an unrelated reason, and the test passes against the buggy code.
- `test_hermes_python_mismatch_normalises_a_detour_spelling` fails without the `normpath` change.
- `test_provider_diagnostic_stays_quiet_for_one_environment` passes either way and is labelled in its docstring as a guard rather than evidence.

403 tests in `integrations/hermes/tests/` pass, and both lint gates are clean.

## Summary

- Fixes Hermes interpreter mismatch diagnostics by using `_hermes_python_mismatch()`.
- Preserves distinct virtual environments that share a base interpreter.
- Normalizes equivalent path spellings without resolving symlinks.
- Shell-quotes remediation commands for paths that contain spaces.
- Adds regression tests for normalized paths, distinct environments, and matching environments.

## Architectural impact

- No changes to tiered memory, retrieval, consolidation, veracity, sync, or benchmark systems.
- No changes to Mnemosyne’s core memory architecture or privacy posture.
- Improves the Hermes integration surface. MCP and CLI behavior remain unchanged.
- Preserves local-first behavior. The change performs local path comparison and adds no external data flow.
- Improves maintainability through shared mismatch logic and focused regression coverage.

This is the right call for Hermes environment detection. It preserves virtual-environment identity while accepting equivalent path spellings. Nothing in this change erodes the local-first design.


dplush:
reviewed and merge-ready

dplush (approved):
Reviewed current head 4bbaf8c. The shared lexical environment check fixes the resolved-symlink diagnostic gap without losing venv identity; the provider path now reuses it and quotes the remediation command. Focused integration tests pass. LGTM.

PR #713, opened 2026-08-12
feat(llm): allow relocating the local model cache

### LINKED ISSUE #708: [FEATURE] Make the model cache location configurable

tonythomson:
## Description

`MODEL_CACHE_DIR` was pinned to `~/.hermes/mnemosyne/models`, so the only way to move a 656 MB GGUF off a small home partition was a symlink, while the repo and filename were already overridable.

Built to your spec: `MNEMOSYNE_MODEL_CACHE_DIR`, environment-only and read at import beside the existing local-LLM overrides, unset or blank preserving the historical path, `~` expanded, and no moving, copying or deleting of existing models. No change to the enabled default, download timing, model selection, fallback behaviour or backup policy.

I swept for every site that touches the cache directory rather than only the definition, so lookup, creation and the download all read the same value and cannot disagree about where the model lives:

| site | role |
|---|---|
| `_model_path()` | cached-file lookup |
| `_ensure_model_cache_dir()` | directory creation (new) |
| `_download_model()` | `hf_hub_download(local_dir=...)` |
| the #704 pre-download notice | reports the destination |

Two decisions worth surfacing rather than leaving to review.

**The raw value becomes the path.** `strip()` only decides whether anything was named. A POSIX directory name may begin or end with whitespace, so stripping the value we actually use would select a different directory. Same reasoning as `--python` in #623, though the empty-value contract is deliberately opposite here: you specified blank as "unset", where `--python` treats blank as an error.

**The error is logged as well as raised.** This is the one thing I would flag hardest. `_load_llm()` catches every exception from the download path and degrades to AAAK, so a raised message alone would never reach the user, and "fail clearly" would have failed invisibly. Logging at the raise site is what makes it visible without changing the caller's control flow.

An explicitly set path is authoritative, so an unusable one fails naming the variable and the path. With no override set, the same failure offers the variable rather than blaming it:

```
MNEMOSYNE_MODEL_CACHE_DIR is set to /mnt/models, which is not writable.
Point it at a writable directory, or unset it to use the default.

The model cache directory /home/u/.hermes/mnemosyne/models, which is not writable.
Set MNEMOSYNE_MODEL_CACHE_DIR to relocate it.
```

Writability is checked separately from creation, since `mkdir(exist_ok=True)` succeeding on a pre-existing read-only directory proves nothing.

One judgement call for you: whitespace-only (`MNEMOSYNE_MODEL_CACHE_DIR="   "`) is treated as blank, so it keeps the default. A directory literally named three spaces is legal but far less likely than an accidental value, and the alternative silently creates one. Say if you would rather it be honoured.

## Related Issue

Closes #708

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

`TestModelCacheDirOverride`, 11 tests. Against `main`, 1 passes and 10 fail. Being precise about what that proves, since most of the failures are simply the new API being absent:

| group | on `main` | what it shows |
|---|---|---|
| unusable directory fails naming the variable and path (3 tests) | **fail** | behavioural: `main` raises a bare `OSError` with no variable, no path, no logging, which `_load_llm` then swallows |
| import-time resolution: unset, blank ×3, `~` expansion, whitespace preserved (6 tests) | fail | the setting does not exist there |
| lookup, creation and `hf_hub_download` all use the override | **pass** | a guard: the existing wiring already routes through `MODEL_CACHE_DIR`, and this pins it |
| unset default failure suggests the variable | fail | new message |

The import-time tests reload the module under a patched environment rather than asserting against a stub, so they exercise the real resolution rule.

Suite with `MNEMOSYNE_NO_EMBEDDINGS=1`: 2635 passed, 26 skipped. Both ruff gates clean over the changed files. `scripts/generate-docs.py --check` and `scripts/verify-docs.py` both clean after regenerating.

## Checklist

Docs: the variable is registered in `ENV_ONLY_DESCRIPTIONS` in `scripts/generate-docs.py`, which is what puts it in the generated `docs/api/configuration.mdx`; added to the environment table in `docs/configuration.md`; and the prose there that stated the cache location as fixed now mentions the override. No `integrations/hermes/README.md` change, since this is core rather than the Hermes plugin surface.

No version bump, matching #704 and #686.

## Summary

- Adds `MNEMOSYNE_MODEL_CACHE_DIR` for the local GGUF model cache.
- Preserves the historical path when the variable is unset or blank.
- Expands `~` for nonblank values and preserves surrounding whitespace.
- Uses the configured path for lookup, creation, validation, downloads, and notices.
- Reports explicit path errors without silently falling back.
- Does not move, copy, or delete existing models.

## Architectural impact

- **Retrieval and consolidation:** No changes to retrieval strategies or the consolidation pipeline.
- **Veracity and sync:** No changes to the veracity system or sync layer.
- **Local-first design:** Strengthens local-first behavior by allowing the model cache to use suitable local storage. No remote service or telemetry path is added.
- **Privacy:** Preserves local model storage and does not change model data handling.
- **Agent integration:** No changes to Hermes, MCP, or CLI interfaces. The variable is read at import time.
- **Maintainability:** Centralizes cache-path behavior, tracks explicit configuration, and validates directory permissions. Logged errors name the environment variable and selected path.
- **Benchmark methodology:** No benchmark changes.

## Documentation and validation

- Updates configuration documentation, generated references, and the changelog.
- Adds tests for defaults, overrides, tilde expansion, whitespace, path consistency, permissions, import restoration, and error handling.
- Existing tests, documentation checks, and lint checks pass.

## Architectural assessment

This is the right call for Mnemosyne-scoped cache portability. It resolves storage constraints without changing memory architecture or introducing non-local dependencies. No change in this PR erodes the local-first design.


dplush (approved):
Approved on c357780. Verified the environment-only import-time contract, consistent lookup/create/download path use, explicit-path failure behavior, generated documentation, and focused tests. CI and CLA are green; no current-head CodeRabbit findings remain.

PR #714, opened 2026-08-12
Feature/Cross-session contradiction resolution for global-scope facts

### LINKED ISSUE #712: [FEATURE] Cross-session contradiction resolution for global-scope facts
# Cross-session contradiction resolution for global-scope facts — feature request

## Problem / Motivation

Contradiction detection in Mnemosyne runs **per consolidation group = one session × one source**. Facts stated in **different threads** are therefore never compared: both are stored, both survive consolidation, and `recall()`/prefetch can return both to the agent (e.g. "favorite color is blue" from Thread A and "…is green now" from Thread B).

With `default_scope: global` — or any cross-session-shared memory — the outdated and current versions surface together. This is additive contradiction, not a genuine supersession: the stale version keeps coming back alongside the correction. It affects **both** explicit recall **and** the automatic prefetch that runs before every turn, because both read the same storage and neither reasons about which statement is current. The canonical store resolves this only for facts that deliberately/derivatively land in it; model-refresh auto-canonicalization is also per-group.

## Proposed Solution

Add an opt-in **cross-session resolution pass** that reuses Mnemosyne's existing, proven primitives, with **no change to recall ordering/scoring/dedup/top_k**:

- New `BeamMemory.resolve_cross_session_conflicts(dry_run=False, min_gap_hours=None)`:
  1. selects `scope='global'`, not-superseded rows from `working_memory` **and** `episodic_memory` across all sessions;
  2. groups by `source`, sorts ascending, and runs conflict detection per group;
  3. generalizes `_detect_conflicts` with a `min_gap_hours` parameter (default 1.0, backward compatible) plus episodic-aware embeddings (`vec_episodes` + `memory_embeddings` fallback);
  4. confirms via `validate_conflict_pair` when `MNEMOSYNE_LLM_CONFLICT_DETECTION` is on, else heuristic-only (deterministic, no-LLM);
  5. `invalidate(older, replacement=newer)` — already cross-session-safe for `scope='global'` rows, and recall already filters `superseded_by IS NULL`, so the loser is hidden from every surface immediately while its history is preserved.
- Runs from `sleep_all_sessions()` (after the per-session loop) and is exposed as an on-demand tool for explicit invocation.
- **Decoupled canonical write:** supersession (hiding the stale copy) happens for every confirmed contradiction; recording the winner as a canonical "current truth" is a separate step gated by the *existing* canonical-worthiness rules — never fabricated.
- **Retirement guard:** once a working row is superseded it must not later be consolidated into a fresh summary (add `superseded_by IS NULL` to sleep's selection).
- **Safety flag:** `MNEMOSYNE_CROSS_SESSION_CONFLICT_RESOLUTION`, default `0` (off), consistent with the project's opt-in pattern for new mutating behavior.

## Alternatives Considered

- **Canonical-preferential / suppression at recall time** (the "validation-aware recall" idea, cf. #449). Considered, but rejected as the primary fix: it keeps the stale copy live in storage, only suppresses it at one surface, risks inconsistency across surfaces, and touches the recall path. The proposed fix resolves in storage so recall and prefetch automatically agree.
- **Widening only the LLM validation gate.** Not sufficient by itself — the bottleneck is that cross-session pairs are never presented for validation at all; the pass is what brings them together.
- **Sub-fact / atomic-fact granularity** (split blended messages so a single stale clause can be pruned). The most complete, but a large architectural change; kept for a separate PR.
- **Do nothing / rely on canonical.** Only resolves facts that reach the canonical layer; raw and episodic copies of contradictory facts still co-present.
- The chosen approach is preferred because it is additive and minimal, reuses trusted primitives, fixes the underlying storage (so every surface agrees), respects scope isolation, and ships behind a default-off switch.

## Impact

- **Who benefits:** anyone using cross-session/global shared memory — e.g. multi-thread assistants, `default_scope: global` setups, and self-hosted agents that want corrupting/stale facts to genuinely supersede instead of accumulating.
- **Breaking change?** No. Gated behind a default-off flag; existing behavior is unchanged until enabled. No change to recall semantics or ordering; session-private memories are never touched; no new dependencies, no cloud/API-key requirement (the LLM step remains opt-in; the default path is deterministic and local).

## Willing to Contribute

- [x] I am willing to submit a PR for this feature
- [ ] I need guidance on where to start
- [ ] Just sharing the idea

sich97:
## Description

What does this change do and why? Keep it focused.

Fixes additive contradiction for cross-session global facts (#712): when `default_scope: global`
(or equivalent) is used, two versions of the same fact stated in **different threads** (e.g.
"favorite color is blue" in Thread A and "…is green now" in Thread B) both survive, and general
recall / automatic prefetch can present the outdated one alongside the correction — the exact
"Hindsight" pain of corrections accumulating instead of superseding.

Contradiction detection in Mnemosyne ran **per consolidation group = one session × one source**,
so conflicting global rows from different sessions were never compared. This PR adds an opt-in
**cross-session resolution pass** that reuses existing, proven primitives and makes **no change to
recall ordering, scoring, dedup, or `top_k`**:

- `BeamMemory.resolve_cross_session_conflicts(dry_run=False, min_gap_hours=None)` selects
  not-superseded `scope='global'` rows from **both** banks (`working_memory` and `episodic_memory`)
  across **all** sessions, groups them by `source`, and runs conflict detection per group.
- `_detect_conflicts` is generalized with a `min_gap_hours` parameter (default 1.0 — backward
  compatible) and episodic-aware embedding lookup (`vec_episodes` + `memory_embeddings` fallback).
- When `MNEMOSYNE_LLM_CONFLICT_DETECTION` is on, candidate pairs are confirmed via the existing
  `validate_conflict_pair`; otherwise the deterministic heuristic is used (no-LLM, local).
- The older row of each confirmed pair is superseded via the existing, cross-session-safe
  `invalidate(older, replacement=newer)`. Recall and prefetch already filter
  `superseded_by IS NULL`, so the loser disappears from every surface immediately while its history
  is preserved.
- A **retirement guard** (`superseded_by IS NULL`) was added to sleep's selection so a superseded
  working row can never be re-minted into a fresh episodic summary.
- Exposed on-demand as the provider tool `mnemosyne_resolve_conflicts` (`dry_run=True` reports
  candidate pairs without mutating), gated by the new opt-in
  `MNEMOSYNE_CROSS_SESSION_CONFLICT_RESOLUTION` flag (default `0`/off).

The deterministic (no-LLM) path is **candidate-reporting-first**: supersession happens only on an
explicit apply (invoking the tool without `dry_run`), and is LLM-confirmed when
`MNEMOSYNE_LLM_CONFLICT_DETECTION` is on. No new dependencies; core remains SQLite-only.

**Phasing:** per maintainer guidance on #712, an automatic invocation from `sleep_all_sessions()`
is deliberately **deferred to a follow-up PR** — in this PR, resolution only ever happens on an
explicit operator action (candidate reporting, then apply). The retirement guard ships here so a
row superseded by the tool can never be re-minted into a fresh episodic summary meanwhile.

## Related Issue

Closes #712

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Refactor / performance
- [x] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

Manual verification: on a real store, two `scope='global'` contradictory rows in different
sessions both surfaced from a third thread; after `resolve_cross_session_conflicts(dry_run=False)`
only the winner remains and the loser is `superseded_by`-set. `dry_run=True` reports the same pair
without mutating anything. Session-private (non-global) memories are untouched. New tests cover
the resolution mechanism, opt-in-default no-op, session isolation, LLM confirm/decline branches,
`min_gap_hours` parametrization, the retirement guard, and provider tool registration/parity.

## Checklist

- [x] Version bumped in `mnemosyne/__init__.py` (3.16.0 → 3.17.0)

## Summary

- Adds opt-in contradiction resolution for global facts across sessions and both memory banks.
- Preserves private-session isolation.
- Uses existing contradiction detection and optional LLM validation.
- Supersedes older confirmed conflicts while preserving history.
- Prevents superseded working-memory rows from re-entering consolidation.
- Supports episodic embedding decoding, configurable time gaps, candidate limits, and LLM validation limits.
- Preserves recall ordering, scoring, deduplication, and `top_k`.
- Defers automatic execution from `sleep_all_sessions()`.
- Exposes `mnemosyne_resolve_conflicts` for deterministic dry runs and explicit apply operations.
- Keeps mutation behind `MNEMOSYNE_CROSS_SESSION_CONFLICT_RESOLUTION`, disabled by default.

## Architecture and privacy

This is the right call for the veracity system. It resolves stale global facts at the storage layer instead of adding recall-only filtering. Recall and prefetch consumers therefore see consistent data.

The tiered memory model remains intact. Working memory excludes superseded rows during consolidation. Episodic memory supplies conflict candidates and embeddings. BEAM retrieval behavior remains unchanged.

The default-off flag, private-row exclusion, dry-run mode, and explicit apply mode protect privacy and operator control. Bank-specific invalidation prevents duplicate IDs across memory banks from affecting the wrong row. Ambiguous cross-bank pairs remain unresolved.

The change adds no sync-layer or remote-service changes. It does not erode local-first guarantees. Optional LLM validation can use an external service, but the default heuristic path remains local.

## Integration and maintainability

Hermes exposes `mnemosyne_resolve_conflicts` with dry-run and apply modes. No MCP or CLI integration is added. The provider reports unavailable resolver capability when the loaded Beam does not support it.

Candidate bounds, validation caps, negative-cap normalization, deterministic dry runs, scan-failure reporting, chained-supersession guards, and shared embedding loading improve operational safety and maintainability.

Tests cover feature gating, cross-session invalidation, privacy boundaries, LLM confirmation, embeddings, time gaps, source grouping, chained conflicts, dry-run immutability, failed scans, candidate limits, duplicate-ID safety, bank-specific invalidation, and consolidation protection. Documentation, changelog entries, and version `3.17.0` updates are included.

The sync layer and benchmark methodology remain unchanged. Automatic `sleep_all_sessions()` integration is intentionally deferred.


hermes-blip:
All actionable CodeRabbit findings have been addressed and verified with regression tests, covering:

- correct decoding of episodic embeddings per their stored vector type (float32/int8/bit);
- a bounded candidate scan (`MNEMOSYNE_CROSS_SESSION_MAX_CANDIDATES`, default 10000) and bounded LLM validation per run (`MNEMOSYNE_CROSS_SESSION_MAX_LLM_VALIDATIONS`, default 20), both surfaced in the result (`candidates_truncated`, `llm_cap_reached`);
- fail-fast on a bank query error (explicit `failed` status, no mutation);
- dry-run / apply parity (dry-run is deterministic and simulates the chain guard);
- atomic, bank-scoped supersession that never overwrites an existing successor and never cross-supersedes a duplicate id across banks; ambiguous pairs are skipped.

The cross-session suite (20 tests, incl. the new edge cases), provider registration/parity, and the surrounding sleep/consolidation/invalidate suites all pass. Ready for another review pass.

PR #720, opened 2026-08-12
fix(beam): warn on skipped single-item embeddings

### LINKED ISSUE #718: [BUG] remember() silently skips vector storage on None or wrong-count embeddings
## Description

`BeamMemory.remember()` silently skips vector storage when embeddings are
available but a per-call embedding attempt returns `None` or the wrong number
of vectors. `remember_batch()` already emits warnings for both conditions, but
the single-item path only logs exceptions.

This causes ordinary conversational writes to lose vector-recall coverage
without an operator signal.

## Environment

- OS: Linux
- Python: 3.11+
- Mnemosyne version/commit: `85a4b88` (upstream `main`)
- Installation: source checkout

## Reproduction Steps

1. Make `_embeddings.available()` return `True`.
2. Make `_embeddings.embed([content])` return `None`, then call:

```python
memory_id = beam.remember("example content")
```

3. Repeat with a returned vector list whose length is not `1`.

## Expected Behavior

The memory row remains stored because embedding is best-effort, but each
failed vector operation emits a content-free warning explaining that vector
storage was skipped.

## Actual Behavior

The memory row is stored and the vector write is skipped, but neither `None`
nor a wrong vector count produces a warning. `remember_batch()` reports these
same conditions explicitly.

This is distinct from #159/#224 (the missing single-item embedding block),
#474 (availability/table honesty), #680 (vector query type), #491 (parent
race), and #600 (reindex failure).

## Suggested Fix

Mirror the two explicit warning branches already present in
`remember_batch()` and add regression tests that assert the memory remains
stored, the vector row is absent, and the warnings contain no memory content.

## Checklist

- [x] I searched existing issues and pull requests for duplicates.
- [x] I can reproduce the failure on upstream `main`.
- [x] The report contains no credentials or private operational data.

bayhnf:
## Description

Fixes the single-item `BeamMemory.remember()` embedding observability gap.
When embeddings are available but the per-call result is `None` or has the
wrong count, the memory write remains best-effort while vector storage is
skipped with a content-free warning.

The patch mirrors the explicit failure handling already present in
`remember_batch()` and adds regression tests for both failure modes.

## Related Issue

Closes #718

## How Has This Been Tested?

- `python -m pytest tests/test_beam.py tests/test_e2_remember_batch_enrichment.py tests/test_beam_e4_remember_batch_veracity.py tests/test_optional_embeddings.py -q`
- `ruff check mnemosyne/core/beam.py tests/test_beam.py`
- `git diff --check`

The focused suites passed with 133 tests and 10 documented optional-dependency
skips.

## Compatibility

Memory insertion remains non-fatal when embedding preparation fails. The
change only makes previously silent `None` and wrong-count results observable;
successful single-item vector writes are unchanged.

## Scope

This PR intentionally covers `remember()` only. The separate
`update_working()` and episodic consolidation paths are unchanged and can be
handled in a follow-up if maintainers want full warning parity.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] No new dependencies
- [x] No credentials or private operational data

## Summary

- `BeamMemory.remember()` now warns when embedding generation returns `None` or an incorrect vector count.
- The working-memory row still persists.
- Vector storage remains skipped for invalid results.
- Warnings exclude memory content and identifiers.
- Successful single-item vector storage is unchanged.
- Regression tests cover both failure modes.

## Architectural impact

- **Tiered memory:** Changes BEAM insertion only. Working-memory persistence remains fail-soft. Episodic-memory behavior is unchanged.
- **Retrieval:** Invalid embeddings reduce vector-recall coverage, but operators now receive a warning. Other retrieval strategies are unchanged.
- **Consolidation and veracity:** No changes.
- **Sync layer:** No changes.
- **Agent integration:** Hermes, MCP, and CLI interfaces are unchanged.
- **Privacy:** Content-free warnings prevent memory disclosure. This is the right call.
- **Local-first guarantees:** Preserved. The change adds no dependency, external service, or remote data flow.
- **Maintainability:** Single-item behavior now matches `remember_batch()`. Regression tests protect warning and persistence behavior.

Focused checks passed: 133 tests with 10 documented optional-dependency skips.


bayhnf:
recheck

dplush:
Thanks for the focused handling of the `None` and wrong-count embedding results. The branch is now behind current `main` and its previous CI failed before the matrix completed.

Could you please rebase onto current `main`, resolve the docs-check / Python 3.11 failures, and force-push the refreshed head? Once the updated CI is complete, I will review the current diff and the fail-soft logging behavior.

bayhnf:
Thanks for the review! Addressed on the refreshed head (`2ce3593`):

- Rebased onto current `main` (`66790b8`) and force-pushed; the unrelated `3.17.0` version bump is dropped, which resolves the docs-check and manifest-version failures (they were caused by the bump without regenerated docs).
- Both regression tests now also assert the memory row persists in `working_memory` (exactly one row for the returned id), covering the fail-soft persistence path directly.
- Verified locally with a clean environment (no `MNEMOSYNE_*` leakage) and TZ=UTC: the focused suites pass. One local note: running the suite with leftover `MNEMOSYNE_EMBEDDING_DIM`/`MNEMOSYNE_EMBEDDING_API_URL` env vars reproduces embedding-dim failures — that's environment leakage on my side, not the patch; CI's hermetic runners don't have those vars. The only unrelated failure (`test_stored_offset_bearing_valid_until_chronologically_filtered`) reproduces identically on clean upstream `main`, so it is pre-existing.

Fresh CI should now be green since the incomplete version bump is gone.

bayhnf:
Correction to my earlier local-test note: I re-ran on clean upstream `main` (`66790b8`) with a fully clean environment and the previously mentioned `test_stored_offset_bearing_valid_until_chronologically_filtered` failure **does not** reproduce there (full file 114/114). It was environment leakage on my side (leftover `MNEMOSYNE_*` vars), which I initially misattributed to upstream. With a clean env and TZ=UTC, the suites pass on `main` and on this PR head. No upstream bug to report; retracting that part of my earlier comment.

PR #730, opened 2026-08-13
fix(packaging): exclude Hermes sources from Core wheel

### LINKED ISSUE #729: [BUG] Core wheel ships repository-only Hermes provider source and tests
## Summary

The root `mnemosyne-memory` wheel includes a repository-only `integrations/hermes/` source tree, including standalone-provider source and tests. This payload is not part of the intended installable Core package and duplicates the separately distributed `mnemosyne-hermes` surface.

## Evidence

Audited on current `main` `85a4b88b506252ac522285f35ff86792731c65f3`:

- Root package discovery excludes `tests*` but not `integrations*`: `pyproject.toml:77-78`.
- A clean PEP 517 root-wheel build produced a 688-KB wheel containing 24 `integrations/hermes/...` members, including:
  - `integrations/hermes/src/mnemosyne_hermes/*.py`
  - `integrations/hermes/tests/test_*.py`

Reproduction:

```bash
python -m build --wheel --outdir /tmp/mnemosyne-wheel-audit
zipinfo -1 /tmp/mnemosyne-wheel-audit/mnemosyne_memory-*.whl | grep '^integrations/hermes/'
```

## Expected

The Core wheel contains only its installable Core package payload. The standalone Hermes provider is delivered by the separate `mnemosyne-hermes` distribution.

## Actual

Core consumers receive a non-installable Hermes source/test subtree in their production artifact.

## Suggested fix

Exclude `integrations*` from the root package finder and add a regression that builds the root wheel and asserts it contains no `integrations/hermes/` members.

## Related but distinct

- Historical PR #469 addressed repository-only wheel payloads but is not in current `main`.
- #656 / PR #689 concern `plugin.yaml` in the separate provider wheel, not this Core-wheel payload. Keep the fixes separate.

ddddddddwp:
## Summary

- exclude the repository-only `integrations/` tree from the Core setuptools package finder
- add a regression test that builds the real Core wheel and inspects its members
- preserve the independently distributed `mnemosyne-hermes` wheel

## Root cause

The root `[tool.setuptools.packages.find]` configuration excluded tests, benchmarks, scripts, docs, and assets, but did not exclude `integrations*`. Setuptools therefore included the nested Hermes provider source and tests in the `mnemosyne-memory` wheel even though Hermes is published separately.

## Impact

Core consumers no longer receive the duplicate, repository-only Hermes source and test payload. There are no public API, runtime dependency, or database schema changes.

## Validation

- TDD regression: failed before the fix with 24 leaked `integrations/hermes/` members, then passed after the package-discovery change
- `MNEMOSYNE_NO_EMBEDDINGS=1 .venv/bin/python -m pytest -q tests/test_core_wheel_contents.py tests/test_plugin_manifest_versions.py` — 5 passed
- `.venv/bin/ruff check tests/test_core_wheel_contents.py` — passed
- built both Core and standalone Hermes wheels
- confirmed the Core wheel contains `mnemosyne/__init__.py` and zero `integrations/hermes/` members
- confirmed the standalone Hermes wheel still contains `mnemosyne_hermes/plugin.yaml`

Fixes #729

## Summary

- Excludes the repository-only `integrations/` tree from the Core setuptools wheel.
- Keeps Hermes available through the separate `mnemosyne-hermes` distribution.
- Adds regression tests for Core and standalone Hermes wheel contents.
- Adds `build` to the `test` extra for portable wheel-build validation.

## Architectural impact

- Core memory architecture is unchanged. Working, episodic, and BEAM tiers are unaffected.
- Retrieval, consolidation, veracity, and sync behavior are unchanged.
- The Core wheel no longer duplicates Hermes source and test files.
- The change preserves local-first packaging. It adds no runtime dependency, network requirement, or privacy change.
- Hermes remains an explicit agent integration surface through `mnemosyne-hermes`. MCP and CLI behavior are unchanged.
- The package boundary is clearer and easier to maintain. Excluding `integrations*` is the right call for separating repository integration code from the installable Core payload.
- No benchmark methodology changes are included.

## Validation

- Regression tests build isolated Core and standalone Hermes wheels.
- Core assertions confirm that legitimate in-package integrations and console entry points remain packaged.
- The Core wheel contains no `integrations/hermes/` members.
- The standalone Hermes wheel retains `plugin.yaml`.
- Hermes manifest and runtime versions remain aligned.
- Expected plugin registrations remain present.


dplush:
Thanks for the focused fix and the real-wheel regression coverage. I reviewed the current head independently.

Before merge, could you please:

1. Rebase onto current `main` (`a8d93a7`), then refresh CI.
2. Add a positive Core-wheel guard for `mnemosyne/integrations/memory_browser.py` and `mnemosyne/integrations/auto_save_openwebui.py` (ideally also their `mnemosyne-browser` / `mnemosyne-auto-save` entry points). This makes sure a future broadening of the exclude cannot remove the legitimate in-package integrations while preventing the repository-root `integrations/` leak.

The standalone Hermes wheel coverage already looks appropriately scoped. I would keep offline/slow-test handling, richer build-failure output, and possible sdist assertions as separate follow-ups unless you are already touching that helper.

ddddddddwp:
Addressed both requested items:

- Rebased the branch onto the latest `main` (`a5a4f0f`, which includes `a8d93a7`) and refreshed CI.
- Added positive Core-wheel guards for `mnemosyne/integrations/memory_browser.py` and `mnemosyne/integrations/auto_save_openwebui.py`, including the `mnemosyne-browser` and `mnemosyne-auto-save` console entry points.

I also addressed the follow-up standalone wheel metadata version check. All checks are green across Python 3.10–3.13, build, lint, docs, CodeRabbit, and CLA.

tests/test_core_wheel_contents.py
```diff
@@ -0,0 +1,157 @@
...
+
+    assert "mnemosyne_hermes/plugin.yaml" in members
+    assert manifest_version == runtime_version
+    assert manifest_version == metadata_version
+    assert entry_points["hermes_agent.plugins"]["mnemosyne"] == (
+        "mnemosyne_hermes:register"
+    )
+    assert (
+        entry_points["hermes_agent.memory_providers"]["mnemosyne"] == "mnemosyne_hermes"
+    )
```
dplush, line 166: The standalone distribution also declares the public installer entry point `mnemosyne-hermes = "mnemosyne_hermes.install:main"` under `[project.scripts]`, but this new artifact regression only checks the plugin and memory-provider groups. The freshly built wheel contains the console script today, yet a future packaging regression could drop it without failing this suite.

Please add the matching assertion for `entry_points["console_scripts"]["mnemosyne-hermes"] == "mnemosyne_hermes.install:main"`, then refresh the focused artifact test and CI.

dplush (changes requested):
Current-head review of `942fe7e` (merge-base `a5a4f0f`).

The root wheel exclusion and the standalone artifact checks are otherwise well scoped. I independently built the wheel/sdist, checked the Core payload and fresh no-deps installs. The Core wheel no longer contains the repository-root `integrations/` tree while the in-package integrations and their console scripts remain present.

One release-contract assertion is still missing below.

tests/test_core_wheel_contents.py
ddddddddwp: Thanks for pointing this out. I rebased the branch onto current `main` (`b775e1d`) and added the requested assertion for `entry_points["console_scripts"]["mnemosyne-hermes"] == "mnemosyne_hermes.install:main"`. The focused artifact tests pass, and CI has been refreshed on `0c37401`.

dplush:
Thanks for the narrow packaging fix and the archive-level regression coverage. The branch is now behind current `main`, and the previous changes-requested decision is not current-head evidence.

Could you please rebase onto current `main` and force-push the refreshed head? Once the full updated CI is complete, I will re-review the actual wheel/sdist contents and the separate `mnemosyne-hermes` package checks on the new head.

ddddddddwp:
Thanks for the careful review and for the clear follow-up. I've rebased the branch onto current `main` (`1aa2454`) and force-pushed the refreshed head (`fd5660f`), so the earlier changes-requested was based on a stale snapshot. CI is re-running across the full matrix (3.10-3.13, build, lint, docs); locally the wheel-contents and Hermes manifest checks all pass clean. I'd really appreciate another look at the wheel/sdist contents and the `mnemosyne-hermes` package checks on the new head once CI finishes. Thanks again for your time.

dplush:
The packaging behavior itself looks good: I rebuilt the core and standalone provider artifacts, confirmed the core wheel excludes the repository `integrations/` tree while retaining `mnemosyne.integrations`, and confirmed the standalone wheel keeps `plugin.yaml` and its entry points.

One small regression-test portability blocker remains before approval. `tests/test_core_wheel_contents.py` builds its copied projects with:

```python
[sys.executable, "-m", "pip", "wheel", "--no-deps", ...]
```

but the declared `test` extra only installs `pytest`; it does not guarantee `pip` (or `build`). In a normal unseeded `uv` test venv, `python -m pip` is absent, so both new artifact tests fail before exercising their assertions. CI currently masks this because it explicitly upgrades/installs pip first.

Could you make the test runner dependency explicit or use a build path guaranteed by the declared test environment? Please keep the archive-level assertions unchanged. After that small fix and fresh CI, I can re-review for approval.

ddddddddwp:
Thanks for the careful check and for catching this. You're right, the tests were quietly leaning on CI's `pip install --upgrade pip` step. I've made the build frontend explicit: `build` is now declared in the `test` extra, and `_build_wheel` uses `python -m build --wheel --outdir` instead of `python -m pip wheel`.

I also verified it in a fresh unseeded `uv` venv with only `.[test]` installed (no `pip` present) and both artifact tests now pass, with the archive-level assertions unchanged. Pushed as `46f430c`, CI is re-running.

dplush:
One tiny release-note request before I approve: could you add a short factual `CHANGELOG.md` entry under Unreleased for #729? Something like: the Core wheel no longer ships the repository-only Hermes provider source/test tree, while the standalone `mnemosyne-hermes` package remains separate.

The implementation and artifact checks otherwise look good. Thanks again for also tightening the test-environment portability.

ddddddddwp:
Done. Added a short CHANGELOG.md entry under Unreleased for #729: the Core wheel no longer ships the repository-only Hermes provider source/test tree, while the standalone `mnemosyne-hermes` package remains separate. Pushed as `fd8763b`.

dplush (approved):
Approved on the current head after fresh artifact verification: the Core wheel excludes the repository-only Hermes tree while retaining in-package integrations; the standalone wheel keeps its manifest and entry points; the unseeded `.[test]` artifact checks pass without pip. The Unreleased changelog entry now covers #729.

Issue #724, opened 2026-08-13
Feature request: optional fact/entity extraction during sleep consolidation (bank-aware)


tiangao88:
**Is your feature request related to a problem?**
Temporal triples (knowledge graph) and canonical facts are only ever created through `remember(extract=True)` or explicit `mnemosyne_triple_add` calls. The nightly `sleep` consolidation — the one operation that runs unattended on every deployment — never extracts anything. Verified against 3.15.1 source: `_extract_and_store_entities` / `_extract_and_store_facts` are called only from the `remember()` path; `sleep()` performs summarization/consolidation only. Result: unattended deployments (e.g. a central MCP server with a nightly consolidation cron) accumulate entity annotations but **0 triples**.

**Describe the solution you'd like**
An opt-in extraction pass during consolidation, e.g. `mnemosyne sleep --extract` (or config `sleep_extract: true`), that:
1. selects the day's unconsolidated working rows (optionally filtered by `importance >= N`, default e.g. 0.7);
2. runs the existing entity and/or fact-extraction pipeline on them (respecting `MNEMOSYNE_LLM_ENABLED`; regex entity fallback when the LLM is off);
3. then proceeds with normal consolidation.

**Bank awareness (dependency on #548)**
Extraction during sleep must respect the bank context in which sleep runs (`MNEMOSYNE_BANK` / bank parameter), so triples land in the bank being consolidated (private or shared), not implicitly in the default store. This depends on #548 — TripleStore default DB is currently disconnected from `mnemosyne_triple_add`/`mnemosyne_triple_query` and the triple tools accept no bank parameter. Without bank-aware triple writes, this feature can only populate the default bank.

**Alternatives considered**
- Keep extraction write-time only (status quo) — KG stays empty on unattended deployments.
- Separate batch tool (`mnemosyne extract --pending`) — also fine; sleep integration is the natural scheduling point and already has the bank loop.

**Additional context**
Running 3.15.1 as a central MCP server (private + shared banks) with a nightly consolidation cron; every bank has its own triples table, all empty after months of operation.

PR #742, opened 2026-08-14
fix: reduce entity-extraction noise (stopword expansion, ALL-CAPS/junk filters, proper-noun shape rule)

### LINKED ISSUE #741: Entity extraction noise: capitalized chat words, ALL-CAPS tokens and sentence fragments flood the memory graph
### Summary

`extract_entities_regex` in `mnemosyne/core/entities.py` and the "X is Y" /
"X has Y" fact extractor in `mnemosyne/core/episodic_graph.py` treat nearly
every capitalized word and quoted phrase in stored memory content as an
entity. For conversation-derived memories this floods the entity/memory
graph with noise ("If", "One", "Nothing", "EXISTING", "A pass that does
nothing"…) and drowns out real entities.

### Affected files

- `mnemosyne/core/entities.py` — `extract_entities_regex` + `ENTITY_EXTRACTION_STOP_WORDS`
- `mnemosyne/core/episodic_graph.py` — `extract_facts` + `_is_low_quality_subject`

### Reproduction (mnemosyne 3.15.1)

Memory content: `User: If nothing works, try Docker. Assistant: One option is the EXISTING setup.`

`extract_entities_regex(...)` returns:
`['If', 'Nothing', 'Docker', 'One', 'EXISTING']` — 4 of 5 are noise.

### Root causes

1. **Stopword list too small** (~120 entries). Common capitalized
   sentence-words ("If", "One", "Nothing", "So", "Want", "Done") and
   code/UI tokens ("ADD", "EXISTING", "PINNED", "LOADED") are not filtered.
2. **Single-word ALL-CAPS tokens are accepted** — emphasis/code, not names.
3. **Multi-word entities are dropped when *any* word is a stopword.**
   "New York" is rejected because "new" is a stopword — legitimate
   multi-word entities are lost.
4. **Quoted phrases are accepted verbatim**, so sentence/template fragments
   ("reliable workflow", "stop doing X") become entities.
5. **"X is Y" fact subjects are any capitalized phrase before "is"** —
   sentence fragments ("A pass that does nothing") become `fact_*` graph
   nodes with high frequency.

### Impact

- Entity graph and entity-aware recall are polluted by noise matches.
- Real entities are diluted. On a private conversation bank with **864
  memories**, 1,078 fact nodes were extracted; after the fixes below the
  same content yields **170** — all meaningful (product names, people,
  organizations).
- Top noisy mention values observed on that bank (pre-fix): "If" ×339,
  "One" ×304, "Nothing" ×246, "So" ×213.

### Proposed fix (PR-ready, tested)

1. `entities.py`:
   - Expand `ENTITY_EXTRACTION_STOP_WORDS` to ~600 entries (common
     capitalized English sentence-words + code/UI tokens).
   - Skip single-word ALL-CAPS tokens.
   - Drop multi-word entities only when **all** words are stopwords
     (fixes the "New York" loss).
   - Require multi-word entities to be proper-noun shaped — each word
     uppercase-start, a digit, or a stopword — which keeps "New York" /
     "San Francisco" and rejects quoted sentence fragments.
   - Skip template junk (leading `;`, `/`, `\`).
2. `episodic_graph.py`:
   - Extend `_is_low_quality_subject` to reject sentence-word-led and
     non-proper-noun-shaped subjects (shares the expanded stopword list).

### Verification

- Same 864-memory bank: fact nodes 1,078 → 170 (−84%); the top-30 mention
  values switch from sentence-words/code tokens to real entities.
- Fresh extraction: `If nothing works, try Docker` → `['Docker']`.

### Related

- #724 (optional entity extraction during consolidation) touches the same
  extraction path.

tiangao88:
Fixes #741.

### What & why

`extract_entities_regex` (entities.py) and the "X is Y"/"X has Y" fact
extractor (episodic_graph.py) treat nearly every capitalized word and
quoted phrase in stored memory content as an entity. On conversation
memories this floods the graph with noise ("If", "Nothing", "EXISTING",
"A pass that does nothing") and dilutes real entities.

### Changes

**`mnemosyne/core/entities.py`**
- Expand `ENTITY_EXTRACTION_STOP_WORDS` from ~120 to ~620 entries: common
  capitalized English sentence-words ("If", "One", "Nothing", "So",
  "Want", "Done") + code/UI tokens ("add", "existing", "pinned", "db"…).
- Skip single-word ALL-CAPS tokens (emphasis/code, not names).
- Drop multi-word entities only when **all** words are stopwords —
  upstream dropped them when *any* word was a stopword, losing legitimate
  entities like "New York" ("new" is a stopword).
- Require multi-word entities to be proper-noun shaped: every word must
  start uppercase, be a digit, or be a stopword. Keeps "New York" /
  "PMO City"; rejects quoted sentence fragments ("reliable workflow",
  "stop doing X").
- Skip template junk (leading `;`, `/`, `\`).

**`mnemosyne/core/episodic_graph.py`**
- Extend `_is_low_quality_subject` to also reject sentence-word-led
  subjects and non-proper-noun-shaped subjects (shares the expanded
  stopword list).

### Verification

On a private conversation bank with 864 memories:
- Fact nodes: 1,078 → 170 (−84%); the remaining subjects are all
  meaningful names (products, people, organizations).
- Top noisy mention values pre-fix: "If" ×339, "One" ×304, "Nothing"
  ×246, "So" ×213 — all gone post-fix.
- Fresh extraction: `If nothing works, try Docker` → `['Docker']`.

Also fixes the latent multi-word loss: `New York` now extracts correctly.

### Notes

- The display-side common-English wordlist used by our viewer is
  deliberately **not** part of this PR (display-layer choice; could be a
  follow-up issue).
- No behavioral change for @mentions/hashtags/quoted-phrase extraction
  semantics beyond the filters above.

## Summary

Reduces entity and fact-extraction noise in the episodic graph.

- Expands stopword coverage.
- Rejects standalone ALL-CAPS tokens, template junk, quoted fragments, and code-like values.
- Rejects multi-word candidates that contain stopwords unless they are explicit `@mentions` or hashtags.
- Requires multi-word entities and fact subjects to match a proper-noun shape.
- Preserves entities such as `New York`, `Europe`, `Docker`, and `Mnemosyne`.

On 864 memories, fact nodes decreased from 1,078 to 170. Values such as `If`, `One`, `Nothing`, and `So` no longer pollute the graph. `Docker` remains extractable from `If nothing works, try Docker`. Entity tests pass.

## Architectural impact

The change improves episodic-graph quality and graph-based retrieval inputs. It does not change working memory, episodic/BEAM tier structure, scratchpad behavior, retrieval algorithms, consolidation flow, veracity handling, or sync protocols.

The proper-noun validation and separate fact-subject filtering are the right call. They reduce false associations while preserving meaningful entities. The larger stopword set increases maintenance cost and requires careful review to avoid rejecting valid entities.

## Privacy and local-first impact

The change remains inside Mnemosyne’s local core extraction pipeline. It adds no network calls, external services, telemetry, or data-sharing behavior. It preserves local-first guarantees.

## Integration impact

Hermes, MCP, and CLI interfaces remain unchanged. Agent integrations should receive cleaner graph results without integration changes.

## Maintainability

The centralized stopword configuration and defensive import behavior support maintainability. The static stopword list increases update and review cost. Explicit handling for mentions and hashtags preserves clear entity signals.

## Verification

The 864-memory check validates extraction precision and graph-size reduction. It is not a BEAM, LongMemEval, retrieval, veracity, or tiered-memory benchmark. Current-tree benchmark claims require separate reruns.

### DIFF
--- a/mnemosyne/core/entities.py
+++ b/mnemosyne/core/entities.py
@@ -45,7 +45,7 @@ ENTITY_EXTRACTION_STOP_WORDS: Set[str] = {
     "having", "however", "if", "into", "just", "keep", "know", "last",
     "like", "look", "looking", "made", "make", "makes", "many", "maybe",
     "might", "more", "most", "much", "must", "need", "needs", "negative",
-    "never", "new", "next", "nothing", "off", "once", "one", "only",
+    "never", "next", "nothing", "off", "once", "one", "only",
     "onto", "other", "others", "otherwise", "out", "over", "part",
     "please", "possible", "rather", "really", "right", "same", "say",
     "see", "seen", "several", "should", "show", "shows", "since", "so",
@@ -239,10 +239,15 @@ def extract_entities_regex(text: str) -> List[str]:
             # Filter out stop words (single word only); case-insensitive
             words = entity.split()
             if len(words) == 1 and entity.lower() in _STOP_WORDS:
-                continue
-            # Multi-word entities: drop only when EVERY word is a stopword
-            # (upstream dropped when ANY word was, killing "New York").
-            if len(words) > 1 and all(w.lower() in _STOP_WORDS for w in words):
+                # @mentions / hashtags are explicit entity signals — bypass
+                # the stopword filter (mirrors the lowercase check below)
+                match_start = match.start(1)
+                prefix_char = text[match_start - 1] if match_start > 0 else ''
+                if prefix_char not in ('@', '#'):
+                    continue
+            # Multi-word entities: drop when ANY word is a stopword
+            # (upstream contract — "The Quick Brown Fox" drops on "the").
+            if len(words) > 1 and any(w.lower() in _STOP_WORDS for w in words):
                 continue
             # Multi-word entities must be proper-noun shaped: every word starts
             # uppercase, is a digit, or is a stopword. Keeps "PMO City",
--- a/mnemosyne/core/entities.py
+++ b/mnemosyne/core/entities.py
@@ -17,7 +17,7 @@ from typing import List, Set, Tuple
 # =============================================================================
 
 ENTITY_EXTRACTION_STOP_WORDS: Set[str] = {
-    # Standard stop words
+    # --- upstream standard stop words ---
     "the", "a", "an", "and", "or", "but", "in", "on", "at", "to", "for",
     "of", "with", "by", "from", "as", "is", "was", "are", "were", "be",
     "been", "being", "have", "has", "had", "do", "does", "did", "will",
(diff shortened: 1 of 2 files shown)


dplush:
Thanks for the detailed report and the focused PR. The noise problem is real, and tightening this path before #724 expands its use makes sense.

A few things need another pass before this is ready:

- CI currently fails in `tests/test_entities.py`: `The Quick Brown Fox` and `The Mnemosyne` are now extracted. Changing the multi-word check from “any stopword” to “all stopwords” is too permissive.
- `europe` should not be in the generic stopword set. `Europe` is a valid entity and can also occur as a fact subject.
- Please add focused regression coverage for both the rejected noise cases and preserved entities, including `New York` and `Europe`.

The goal seems right. I would keep the filtering rule narrower rather than relying on a large broad stopword list. Once the existing tests and these boundary cases are covered, happy to take another look.

dplush:
One implementation direction that may keep this maintainable: I would avoid growing a single global stopword list with ordinary nouns and proper names.

It may be cleaner to distinguish hard single-token rejects, such as pronouns, chat roles, and obvious sentence starters, from phrase-leading connectors such as `the`. A full phrase beginning with a connector can be rejected while its meaningful fallback candidate remains available. That keeps `The Mnemosyne` out as a phrase without losing `Mnemosyne`, and still allows `New York` and `Europe`.

I would also keep quoted fragments and ALL-CAPS tokens as separate candidate classes rather than handling them through the same word list. They have different false-positive patterns.

A small table-driven regression set would make the boundary explicit: noise examples such as `If` and `EXISTING`, preserved names such as `Docker`, `New York`, and `Europe`, plus rejected full phrases such as `The Mnemosyne`.

e.g.
```
noise
"If nothing works, try Docker." -> ["Docker"]
"One option is the EXISTING setup." -> []
'"reliable workflow"' -> []

real names
"New York" -> ["New York"]
"Europe" -> ["Europe"]
"OpenAI API" -> ["OpenAI API"]
"NASA launched Artemis" -> ["NASA", "Artemis"]  # maybe low confidence for NASA

phrase edges
"The Mnemosyne project" -> ["Mnemosyne"]
"The Quick Brown Fox" -> no "The Quick Brown Fox"
```

For fact subjects, I would prefer a separate validation rule instead of importing the entity stopword set directly. The graph-write contract is stricter and should not change whenever the entity filter changes.

dplush:
Thanks for the focused noise-reduction work. I think this can move forward as an entity-extraction PR, with a few concrete fixes before approval:

1. Keep geographic proper names out of the generic stopword set. The current `"europe"` entry rejects `Europe` as an entity and as a fact subject; please remove it and add entity plus fact-subject regression coverage.
2. Preserve explicit `@`/`#` signals through the ALL-CAPS and substring-removal checks. `@NASA`, `#API`, etc. should not lose their provenance when the regex capture strips the prefix.
3. Extend the narrow template/code predicate to reject placeholder delimiters such as `{{user}}`, without broadening it into a general text filter.
4. Please rebase onto current `main` (`9d34b6d`) and let the full matrix rerun.

Scope note: this PR should remain a focused entity-noise reduction. #837 is a separate fact subject/object-quality contract and should not be folded in here.

The broad stopword/proper-noun policy remains a maintainer decision after the technical fixes are in place; we should not silently treat this PR as the final policy for all graph/recall extraction.

AxDSan (changes requested):
The problem is real and the ALL-CAPS, junk-character and proper-noun-shape rules are good. The stopword list is where I have to push back, because it drops entities that are unambiguously entities.

## Measured regression

Same inputs, `main` vs this branch:

| Input | main | this PR |
|---|---|---|
| `We are expanding to Europe next year` | `['Europe']` | `[]` |
| `The report is in English` | `['English']` | `[]` |
| `August revenue was strong` | `['August']` | `[]` |
| `Meeting on Monday with Sarah` | `['Meeting', 'Monday', 'Sarah']` | `['Meeting', 'Sarah']` |

A continent, a language, a month and a weekday. Those are not noise in any corpus, they are the category entity extraction exists to capture. `europe`, `english`, `august`, `monday`, `mondays`, `router`, `webhook`, `websocket`, `hub`, `net`, `core`, `domain` and `identity` are all now permanent stopwords, and the list is global, so there is no way for a user whose memories are about European deployments or English-language documents to get them back.

The existing suites do not catch this: `tests/test_entities.py` + `tests/test_entity_integration.py` are **40 passed** on this branch. That is the coverage gap, not a green light.

## Why I think the approach needs to change rather than the list be trimmed

The second and third blocks are labelled "observed high-frequency single-cap words in dev-chat". That is a sample of one corpus, this project's own, and roughly 500 words of it are being written into a global constant that every Mnemosyne user inherits. A word being frequent in our chat is not evidence it is not an entity in someone else's memories.

Three directions, any of which I would take:

1. **Frequency-based, not list-based.** A word appearing in a large fraction of a store's memories is noise *in that store*. That adapts per user instead of encoding ours.
2. **Keep the list but make it configurable**, with the shipped default limited to words that are never entities in any corpus. `entity_stopwords` alongside the existing config keys.
3. **Keep the shape rules, drop the corpus-derived blocks.** The ALL-CAPS filter, the `;/\` junk filter and the proper-noun-shape rule are principled and corpus-independent. I would take those on their own today.

Option 3 is the smallest useful PR and I would merge it more or less as-is.

## Smaller things

**Dead clause in `entities.py`.** In the new multi-word block:

```python
if len(words) > 1 and any(w.lower() in _STOP_WORDS for w in words):
    continue
if len(words) > 1 and not all(
    w[0].isupper() or w[0].isdigit() or w.lower() in _STOP_WORDS
    for w in words
):
```

By the time the second check runs, the first has already dropped anything containing a stopword, so `or w.lower() in _STOP_WORDS` can never be true. It is live in `episodic_graph._is_low_quality_subject`, where there is no preceding `any()`, which is probably where it was copied from. Harmless, but it reads as if it does something.

**Duplicates.** `be`, `this`, `there`, `them`, `thing`, `time`, `was`, `way`, `were`, `what`, `whether`, `which` already exist in the upstream block. The set literal dedupes them silently, so no bug, but it suggests the list was appended without a pass over what was already there.

**No CHANGELOG entry and no tests.** 129 lines changing extraction behavior with neither. Whichever direction this goes, the Europe/English/August cases belong in `test_entities.py` as assertions so the next tuning pass cannot quietly re-break them.

**One thing I checked and was wrong about**, noting it so it does not come up again: I suspected the new `@`/`#` lookback duplicated a broken `entity.startswith('@')`. It does not. The lowercase branch already does the same lookback, `startswith` is vestigial there too, and your comment saying it mirrors the check below is accurate. `@denis` extracts correctly on both branches.

AxDSan (changes requested):
Decision: the structural filters stay, the word list has to shrink.

Keep: the ALL-CAPS single-token skip, the all-words-stopword rule for multi-word entities (the New York fix is right), the proper-noun shape rule, and the template-junk skip.

Change: the expanded list is matched case-insensitively, so it drops real entities that happen to be common words. From the list as it stands, at least August, Monday, Mondays, Europe, English, Password, Router, Bridge, Hub, Dashboard and Watchdog are legitimate proper nouns or product names in the memories this extractor runs on. Please cut every entry that can be a name, a month, a weekday, a place, a language or a product, and keep only words that are never an entity on their own.

Also: #876 landed today and touches _is_low_quality_subject, and #894 removed the quoted-span patterns, so this needs a rebase over current main. The three article entries #876 added to _LOW_QUALITY_SUBJECT_LEADERS become redundant with your shape rule, which is fine.

Happy to merge once the list is trimmed and CI is green on the rebased branch.

Milgauss:
Cross-link, so two stop lists do not get written in parallel: #909 covers the same ground from the other direction, and @dplush proposed splitting it into exactly the two pieces this PR straddles.

**The split proposed on #909 (2026-09-03):** (1) a narrow core-quality fix carrying a conservative, maintainer-approved global stop list, with coverage on direct extraction and on `remember(..., extract_entities=True)`; (2) a separate deployment-config feature, `MNEMOSYNE_ENTITY_STOP_WORDS`, additive union with the built-ins, case-insensitive comma-separated, process-start semantics.

This PR is piece (1), and it is the natural home for it. Nothing in #909 needs a second PR against `entities.py` if the list lands here.

**One thing worth carrying over, because it is what the review feedback here is circling.** The extractor applies the stop-word filter to every word of a multi-word candidate, so a global entry does not only drop a shouted interjection, it can also eat a real name that contains the word. That is the same failure as the `"europe"` entry called out above: `Europe` stops being an entity because a generic list claimed the word.

The policy we proposed on #909 is a measurable version of that objection: **a word goes on the global list only if it has zero measured collisions; any word that costs a real entity somewhere stays off it and reaches a deployment through the env hook instead.** Measured against 5,100 distinct stored entity values from a production store, that admits

```
wait, sorry, please, really, know, again, whole, course, important
```

and rejects `god`, `jesus`, `christ`, which cost `Mad God` and `Jesus Christ`. Same treatment as `one` and `more`: real noise, but deployment-dependent, so not global. Geographic names fail the test for the same reason `europe` does here.

Happy to hand the measurement over rather than open a competing PR. If it is useful, I can run any candidate list against the same corpus and report which entries collide with a stored entity and which do not, so the global list is decided on counts instead of intuition. Say the word and I will post the results here.

PR #743, opened 2026-08-14
fix(mcp): handle non-ASCII bearer tokens safely

### LINKED ISSUE #739: [BUG] MCP server 500s on every request to `/mcp` due to `TypeError` in bearer token comparison
### Description
The MCP server's auth middleware crashes with a `TypeError` on every request to the `/mcp` (streamable-HTTP) endpoint. The `/sse` transport's equivalent auth check is unaffected.

### Traceback
```
File ".../mnemosyne/mcp_server.py", line 229, in __call__
    if not hmac.compare_digest(presented, expected):
TypeError: comparing strings with non-ASCII characters is not supported
```

### Root cause
`hmac.compare_digest()` requires both arguments to be either `bytes`, or `str` containing only ASCII characters — it raises `TypeError` on any other combination. The `presented` token value derived from the incoming `Authorization` header is not guaranteed to satisfy this, and neither value is being encoded to `bytes` before comparison.

### Fix (confirmed working via manual patch + restart)
```python
# before
if not hmac.compare_digest(presented, expected):

# after
if not hmac.compare_digest(presented.encode(), expected.encode()):
```

After applying this and restarting the server, requests to `/mcp` no longer 500 — they resolve to normal routing behavior (e.g. 404 on paths/methods that don't exist, rather than crashing on the auth check itself).

### Environment
- `mnemosyne-memory[mcp]` version 3.15.1
- Run via `mnemosyne mcp --transport sse --host 0.0.0.0 --port 8000` inside a `python:3.11-slim` container
- Reproduced consistently regardless of token validity — even a correct, valid bearer token triggers the crash on `/mcp`

### Steps to reproduce
1. Start the MCP server with `MNEMOSYNE_MCP_TOKEN` set.
2. Send any request to `/mcp` or `/mcp/` with an `Authorization: Bearer <token>` header (any token, valid or invalid).
3. Server returns `500 Internal Server Error` instead of `401`/expected routing behavior.
4. Compare with the same request against `/sse` — this endpoint's auth check works correctly.

### Impact
Any MCP client defaulting to the streamable-HTTP transport (`/mcp`) rather than SSE (`/sse`) cannot connect to a token-protected Mnemosyne server at all.

ddddddddwp:
## Summary

- keep MCP SSE bearer tokens as bytes through parsing and constant-time comparison
- return `401 Unauthorized` for malformed non-ASCII bearer bytes instead of raising `TypeError`
- cover both authenticated routes, `GET /sse` and `POST /messages/`
- preserve matching and mismatching ASCII-token behavior

## Root cause

The pure-ASGI auth middleware decoded the raw `Authorization` header with Latin-1 and passed the resulting strings to `hmac.compare_digest()`. Python rejects non-ASCII string operands there, so malformed bearer material could raise `TypeError` and produce a server error.

The fix compares raw presented bytes with the UTF-8 bytes of the configured token. This keeps constant-time comparison and avoids an unsupported string-comparison path.

This is scoped to the current SSE transport routes; it does not add or change a streamable `/mcp` endpoint.

## Validation

- TDD regression: both non-ASCII route cases failed with the reported `TypeError` before the fix
- MCP SSE auth and MCP server tests: 110 passed
- Core test suite: 2680 passed, 34 skipped, 7 subtests passed
- Ruff on changed files: passed
- Python compileall: passed

Closes #739

## Summary

- Fixed non-ASCII bearer-token handling for `GET /sse` and `POST /messages/`.
- Compared raw authorization bytes with UTF-8 bytes from the configured token.
- Malformed non-ASCII tokens now return `401 Unauthorized` instead of raising `TypeError`.
- Preserved valid and invalid ASCII-token behavior.
- Added integration coverage for authentication and routing responses.

## Architectural impact

- No changes to Hermes or CLI integration.
- The existing MCP SSE authentication surface now handles malformed input safely.
- No changes to the streamable `/mcp` endpoint.
- The fix preserves local-first behavior. It adds no external service, data flow, or privacy exposure.
- Byte-based constant-time comparison is the right call for authentication correctness and maintainability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
+- **MCP SSE authentication rejects malformed non-ASCII bearer tokens with `401` instead of returning a server error (#739).**
 - **API embedding failures now leave a redacted diagnostic trace (#735).** Final HTTP, network, and invalid-response failures still degrade to keyword-only retrieval, but now log the endpoint and safe error class or status without request content, API keys, URL userinfo, query strings, or fragments.
 - **Hermes tool discovery now honors `memory.mnemosyne.tools` (#725).** Tools outside the configured allowlist are no longer advertised through Hermes provider schemas before provider initialization.
 - **Truncated LLM reasoning traces no longer reach memory persistence (#734).** Malformed or unbalanced `<think>` output is rejected before fact extraction, model-refresh parsing, or sleep consolidation; sleep falls back to AAAK rather than persisting a partial LLM summary.
--- a/mnemosyne/mcp_server.py
+++ b/mnemosyne/mcp_server.py
@@ -189,7 +189,8 @@ def _build_sse_app(host: str = "127.0.0.1"):
 
     middleware = []
     if require_auth:
-        expected = token
+        assert token is not None
+        expected = token.encode("utf-8")
 
         class _BearerTokenMiddleware:
             """Pure-ASGI bearer auth middleware.
@@ -212,12 +213,12 @@ def _build_sse_app(host: str = "127.0.0.1"):
                 if scope.get("type") != "http":
                     await self.app(scope, receive, send)
                     return
-                header = ""
+                header = b""
                 for k, v in scope.get("headers", []):
                     if k == b"authorization":
-                        header = v.decode("latin-1")
+                        header = v
                         break
-                if not header.startswith("Bearer "):
+                if not header.startswith(b"Bearer "):
                     resp = JSONResponse(
                         {"error": "missing bearer token"},
                         status_code=401,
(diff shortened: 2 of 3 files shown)


dplush (approved):
Current-head security review approved for f1a3482.

Verified against merge-base a5a4f0f: the byte-based constant-time comparison removes the non-ASCII TypeError without changing normal authentication semantics. The middleware covers the actual protected SSE routes, GET /sse and POST /messages/. Focused ASGI checks verified valid ASCII credentials pass through, malformed non-ASCII credentials return 401 rather than 500, and no auth bypass was found.

The original /mcp wording in #739 is stale: the current SSE server does not register that streamable-HTTP route. CI, CLA, and CodeRabbit are green on this head.

dplush:
Thanks again for the focused fix. Could you please rebase or update the branch against the current `main` (`b775e1d`) and refresh CI?

The security review is otherwise approved, so once the current head is green we can complete the merge gate. Thanks!

ddddddddwp:
Rebased the branch onto current main at `b775e1d` and force-pushed the updated head (`5714fce`). CI has been refreshed and is now running against the rebased commit.

dplush:
Thanks for the focused fix. One merge-gate item remains: this is a user-facing MCP authentication robustness fix, but the PR does not yet include an `Unreleased` → `Fixed` entry in `CHANGELOG.md`.

Could you please add this minimal note and refresh CI?

```md
- **MCP SSE authentication rejects malformed non-ASCII bearer tokens with `401` instead of returning a server error (#739).**
```

No version bump is needed in this PR; release versions are handled separately. Once that is in, the security review is otherwise approved. Thanks!

Issue #707, opened 2026-08-12
[BUG] WAL/SHM sidecars unlinked under live gateway connection shortly after boot (occurs on 3.15.1, post-#432/#520)


GusBot69:
## Summary

On a single-user home server running the Hermes gateway with mnemosyne-memory **3.15.1** + mnemosyne-hermes **0.5.0** (dedicated venv, Python 3.11, Ubuntu 24.04, SQLite WAL mode + sqlite-vec), the `-wal`/`-shm` sidecar files of `mnemosyne.db` are **unlinked from disk while the long-lived gateway process still holds them open**, shortly after a machine reboot. The gateway then keeps writing through the orphaned inode: writes fail with `disk I/O error` / `E6: failed to initialize annotations schema: disk I/O error`, vector reads fail with `database disk image is malformed` — while any **fresh** connection (CLI, health tooling) sees a perfectly healthy DB (`PRAGMA integrity_check` = ok). Only a gateway restart recovers. **Occurred twice (2026-08-08 and 2026-08-11).**

## Forensic signature

- Gateway holds: `/proc/<pid>/fd/N → .../mnemosyne.db-wal (deleted)` and `/proc/<pid>/fd/M → .../mnemosyne.db-shm (deleted)`
- On-disk sidecars gone; the gateway kept writing to the orphaned WAL inode (grew to ~6.4 MB after deletion) — data went nowhere recoverable
- Timeline: reboot 18:41 → gateway opens DB 18:43 with fresh sidecars → sidecars unlinked by **18:50** (within ~7 min of boot)
- Copy-replay of the orphaned WAL into a DB copy reproduces `malformed` at a page-level mismatch (Tree 17, page 28672) → the orphaned WAL is a **foreign generation** and must NOT be merged
- Fresh connections pass `integrity_check` the whole time — the classic "two views of the same DB" trap

## Why this is distinct from #382 / #432 / #520

Those are WAL *lock/checkpoint* problems (frozen WAL, blocked checkpoint, concurrent Beam access). This is **file unlinking** — the sidecars are removed from the directory while a live connection holds them. Notably, **3.15.1 already contains the #432 and #520 fixes** (lock hygiene, Beam access serialization), so this failure occurred *despite* them.

## Suspects ruled out on our side

- Hermes backup job: explicitly excludes `-wal`/`-shm`
- Dashboard service: read-only connections
- disk-cleanup plugin: scoped to tracked ephemeral files, never touches `*.db-wal/shm`
[...]

PR #747, opened 2026-08-14
fix(core): bound SQLite FD retention under connection churn

israellot:
## Issue

`mnemosyne.core.memory` and `mnemosyne.core.beam` cache SQLite connections in thread-local storage. When short-lived threads exit, or a live thread replaces its cached connection while switching database paths, the displaced SQLite objects can remain in cyclic garbage and continue holding database/WAL file descriptors until Python performs a full collection. Sustained connection churn can therefore exhaust the process file-descriptor limit.

## Fix

Add a shared connection-creation counter for both caches and run a full cyclic-garbage collection after every eight new connections. This bounds descriptor retention while preserving the existing connection lifecycle. The call is process-wide: it collects all unreachable cycles, not only SQLite objects. Connections still referenced by live Mnemosyne or Beam objects remain usable.

Regression tests cover both caches under short-lived thread churn, live-thread database switching, and connections that outlive their creating thread.

## Relation to prior work

This is a narrow follow-up to #382 and the lifecycle concerns recorded in #477. Unlike the cleanup approach reverted in #477, this change does not add object destructors, explicit WAL checkpoints, instance-wide connection shutdown, or creator-thread-based connection closing. It leaves multi-database ownership semantics unchanged. The periodic full collection scans process-wide cyclic garbage, so occasional tail latency scales with the process heap rather than only the SQLite objects that motivated the change.

## Summary

- Adds process-wide cyclic garbage collection after every eight new thread-local SQLite connections in `memory` and `beam`.
- Reclaims unreachable SQLite connection cycles from short-lived threads and database-path changes.
- Preserves connections that live objects still reference.
- Adds regression tests for thread churn, live-thread database switching, and connections that outlive their creating threads.
- Documents that `gc.collect()` can collect all unreachable cyclic garbage and that tail latency can depend on process heap size.

- Improves file-descriptor retention for working memory and BEAM caches.
- Does not change tiered memory behavior, retrieval strategies, consolidation, veracity, or sync.
- Preserves the local-first design. The change adds no remote service, data export, or new privacy boundary.
- Preserves local SQLite durability and connection usability.
- Does not add destructors, WAL checkpoints, instance-wide shutdown, or creator-thread-based connection closing.
- Centralizes connection-cycle cleanup in `collect_connection_cycles()`. This is the right call for shared behavior and long-term maintainability.
- Uses a lock-protected counter to limit collection frequency.
- Process-wide collection can add heap-size-dependent tail latency. The documentation states this trade-off.
- Tests disable cyclic garbage collection during retention checks. This provides direct coverage for descriptor retention under connection churn.


dplush:
Thanks for the focused FD-retention fix. I independently reproduced the intended improvement under thread-local connection churn, and also verified that active Core/BEAM connections on the same DB stay usable across the collection cadence; committed WAL writes survived reopen and `integrity_check` stayed clean.

The remaining blocker is only the current conflict with `main`. Could you please rebase, resolve it, and rerun CI plus CodeRabbit on the refreshed head?

One small documentation/rationale correction while touching the branch: `gc.collect()` is process-wide. It collects all unreachable cyclic garbage, not only SQLite connection cycles, and can therefore introduce occasional heap-size-dependent tail latency. That trade-off looks acceptable for this bounded FD-retention fix, but it should be stated accurately rather than framed as SQLite-only cleanup.

This is independent of #707: this PR should not add a lifecycle/lease contract or attempt to solve WAL-sidecar ownership. After a narrow rebase and fresh green current-head checks, I expect it to be ready for approval.

israellot:
Rebased onto current `main` and resolved the `mnemosyne/core/memory.py` conflict while preserving the new connection-owner behavior.

I updated the helper docstring, changelog, and PR description to state that `gc.collect()` is process-wide and that occasional tail latency depends on heap size. Local core, Hermes integration, lint, docs, and package checks pass.

@coderabbitai review

israellot:
@coderabbitai full review

dplush (approved):
Approved after current-head CI/CLA/CodeRabbit verification plus independent Sol and Claude/Opus SQLite lifecycle gates. The FD-retention fix remains scoped to bounded collection cadence; it does not establish a connection ownership contract or resolve #707.

PR #750, opened 2026-08-14
fix: split leading-hyphen fragments into FTS-safe recall terms (#744)

### LINKED ISSUE #744: FTS recall misses short and symbolic technical queries (rm -rf, C++)
## Summary

BeamMemory.recall() returns 0 results for queries containing short or symbolic technical tokens such as `rm -rf`, `C++`. The failure is silent: no exception reaches the caller, the method just returns an empty list. Memories phrased with those terms are therefore unrecallable by their natural wording.

## Root cause

Two layers, depending on version.

On current main (confirmed by maintainer review): the tokenizer drops any token below the three-character meaningful gate. For `rm -rf`, both `rm` and `rf` fall under that floor, so the query produces zero candidates and recall returns `[]` without raising. `C++` is filtered the same way.

On the build the report was first filed against (mnemosyne 3.15.1 core, mnemosyne-hermes 0.5.0), the symptom also surfaced as a swallowed SQLite error, `no such column: rf`. That points at the leading `-` reaching the FTS5 query builder as the NOT / column-exclusion operator (`"-rf"`), which SQLite rejects and recall() then swallows. Hyphen-component expansion on main appears to cover this, but the observation is kept here for version provenance.

## Minimal reproduction

```python
from mnemosyne.core.beam import BeamMemory
bm = BeamMemory()
bm.remember("The user does not like the use of `rm -rf`.", source="user", scope="global")

# Hyphenated / short query -> 0 results, no error
print(len(bm.recall("rm -rf", top_k=5)))   # -> 0

# Hyphen-free paraphrase -> works
print(len(bm.recall("trash retention", top_k=5)))  # -> 1
```

Direct FTS check (version-dependent; see root cause):
```sql
SELECT id, rank FROM fts_working WHERE fts_working MATCH 'rm OR -rf';
-- On the reported build: Error: no such column: rf
```

## Expected behavior

Short or symbolic technical queries should recall their stored memories, either by splitting hyphen fragments into components (`rm`, `rf`) for OR-matching, or by a bounded fallback that avoids dropping them silently.

## Impact

Any stored preference or fact whose natural recall phrase uses a short or hyphenated token (shell commands, library names, programming terms) is effectively unrecallable. Combined with the swallowed exception, this is hard to diagnose: recall just misses without explanation.

## Environment

- mnemosyne 3.15.1 (core), mnemosyne-hermes 0.5.0 (original report)
- Python 3.11, macOS (Apple Silicon)
- sqlite-vec 0.1.9 (FTS5 available)

## Status

A fix is in flight: PR #750 (fix: split leading-hyphen fragments into FTS-safe recall terms). It adds a 2-char floor and recovers hyphen-fragment components, and keeps `git-rebase` / `node_modules` behavior intact.

ddddddddwp:
## Summary

Fixes #744. `BeamMemory.recall()` silently returned 0 results for queries containing leading-hyphen fragments such as `rm -rf` (also `--force`, `-v` style flags).

Root cause on current main:

- `_recall_tokens()` cannot capture a token that starts with `-` (`_RECALL_TOKEN_RE` requires a word character to start a token), so fragments like `-rf` silently vanish from recall. With `rm` and `rf` both below the 3-char meaningful gate, the query produced zero tokens, every candidate was lexical-gated to 0, and recall returned `[]` with no error.
- Latent FTS5 crash: had a `-`-leading term ever reached the FTS query builder, `"rm" OR "-rf"` is invalid FTS5 (`-` is the NOT / column-exclusion operator -> `no such column: rf`), and the working-memory FTS search swallows the exception, returning `[]`.

## Changes (`mnemosyne/core/beam.py`)

1. New `_hyphen_fragment_tokens()` helper: recovers components of leading-hyphen fragments (`rm -rf` -> `['rf']`, `--force` -> `['force']`), with a 2-char floor and stopword/digit filtering. Embedded compounds (`git-rebase`) are untouched.
2. `_fts_query_terms()`: never emits a term starting with `-` (FTS5-safety), and appends fragment components as quoted OR terms.
3. `_lexical_relevance()`: admits fragment components on both the query and content side so the lexical gates stop silently blocking hyphen-fragment queries.

## Verification

- `recall("rm -rf")` now returns the memory (was 0 results, no error).
- Existing hyphen behavior unchanged: `git-rebase`, `node_modules`, `orion-telemetrie` tokenize exactly as before.
- New regression tests in `tests/test_multilingual_local_recall.py`:
  - `test_hyphen_fragment_tokens_extract_leading_hyphen_components`
  - `test_fts_query_terms_never_emit_hyphen_leading_terms`
  - `test_hyphen_fragments_score_lexically_without_admitting_distractors`
  - `test_leading_hyphen_fragments_recall_via_public_api`
- Suites green: `test_multilingual_local_recall.py` (22), `test_recall_precision_regressions.py` + `test_benchmark_pure_recall_gate.py` + `test_recall_explain.py` + `test_c4_recall_diagnostics.py` (71), `test_beam.py` + `test_memory_banks.py` + `test_fact_recall_ranking.py` (133), and full `tests/` run (511 passed; the only failure, `test_config_profiles.py::test_apply_writes_to_config`, reproduces identically on clean `origin/main` and is unrelated).
- `ruff check` clean on both changed files.

## Summary

Fixes recall failures for leading-hyphen fragments such as `rm -rf`, `--force`, and `-v`.

- Extracts safe fragment components.
- Prevents invalid FTS5 terms that begin with `-`.
- Uses bounded exact lexical matching for symbolic terms such as `C++`, `C#`, and `g++`.
- Preserves literal flag matches over bare-word distractors.
- Preserves embedded-hyphen behavior.
- Adds regression tests and changelog documentation.

## Architectural impact

- Improves retrieval in the BEAM path.
- Leaves working memory, episodic memory, and tiered-memory boundaries unchanged.
- Does not change consolidation, veracity, synchronization, Hermes, MCP, or CLI surfaces.
- Preserves local FTS5 and lexical retrieval.
- Does not change data movement, external services, or the privacy posture.
- Preserves local-first guarantees.
- Improves maintainability through centralized fragment handling and focused regression coverage.

This is the right call for retrieval correctness. It removes silent zero-result recall without eroding the local-first design.


dplush:
Thanks for the focused fix. The existing CI matrix is green, but the branch is now conflicting with current `main`.

Could you please rebase onto current `main` and force-push the refreshed head? Once the updated CI is complete, I will review the current diff, with particular focus on the confirmed `rm -rf` and symbolic-query behavior.

ddddddddwp:
Thanks — rebased onto current `main` and force-pushed the refreshed head. CI is now fully green (test 3.10–3.13, lint, build, docs-check, CodeRabbit, CLA).

On your review focus: the refreshed diff now covers both confirmed behaviors from #744, with regression tests for each.

- **`rm -rf` / `--force` / `-v` leading-hyphen fragments** are split into FTS-safe components. Tests cover extraction (`_hyphen_fragment_tokens`), FTS5 safety (no `-`-leading terms), lexical scoring, and end-to-end recall.
- **Symbolic queries (`C++`, `C#`, `F#`, `g++`)** are now handled as well: they were dropped by the 3-char meaningful-token gate, so `recall("C++")` silently returned 0. They are admitted as exact lexical tokens on both sides, deliberately **not** through FTS5 — unicode61 tokenizes `C++` down to a single `c` token, which would flood candidates with bare-`c` noise, so pure symbolic queries fall back to the bounded exact-only path. Added tests cover extraction (rejecting `a+b`, `git-rebase`), FTS5 non-emission, lexical scoring without admitting distractors, and end-to-end recall for both `C++` and `c#`.

dplush:
Thanks for the rebase and the careful FTS-safety work. I found one precision blocker before approval.

`--force` is currently reduced to the bare component `force` for both FTS and lexical scoring. That makes ordinary uses of “force” indistinguishable from the literal CLI flag. In a public recall reproduction with embeddings disabled:

- intended: `Use the deployment command with --force only after confirmation.` (importance 0.1)
- distractor: `The force field calibration is complete.` (importance 1.0)

`recall("--force")` returned the generic “force field” distractor first. Both rows received `keyword_score=1.0`.

Could you preserve or explicitly weight the literal leading-hyphen form so an exact `--force` match cannot be outranked by a bare `force` occurrence? Please add a persisted public `recall("--force")` regression with this distractor boundary. The existing `rm -rf` test does not expose this overlap.

The FTS parser-safety work itself looks good; this is only the precision contract for literal flags.

ddddddddwp:
Thanks for the precision callout, it was spot on. Both rows were getting `keyword_score=1.0` because the literal `--force` form was being reduced to the bare component `force`. Fixed on the branch, CI green:

- `_leading_hyphen_fragments()` keeps the literal flag form as a distinct lexical token, so an exact `--force` row scores 1.0 while a bare `force` row scores 0.5.
- A bounded `_literal_flag_bonus` is applied in the working/episodic scoring loops, so the exact flag match cannot be outranked by a bare-word occurrence on importance.
- Regression tests cover your exact distractor boundary (intended imp 0.1 vs bare `force` imp 1.0), a mixed `deploy --force` query, and the `--forceful`/`foo--force` prefix boundary.

Your reproduction now returns the literal `--force` row first.

dplush:
Thanks for the precise `--force` fix and the focused regression boundaries. The branch is now behind the green `main` after #730. Could you please rebase onto current `main` and force-push the refreshed head? Once the full CI completes on that head, I will run the final current-head review.

ddddddddwp:
Done, rebased onto current main (post-#730) and force-pushed. CI is re-running on the refreshed head now. Once it's green it should be ready for your final review.

dplush:
Thanks for the rebase and the careful literal-flag fix. I found one remaining configurable-ranking blocker before approval.

The fixed literal bonus is not enough when importance is configured to dominate. On the current head:

```python
beam.recall("--force", top_k=2, vec_weight=0, fts_weight=0, importance_weight=1)
```

with the exact `Use --force after confirmation.` row at importance `0.1` and a bare `The force field calibration is complete.` distractor at `1.0`, the bare `force` row still ranks first. The fixed `+0.3` bonus cannot override that weight configuration.

Could you make literal-flag precedence a true selection/ranking invariant, for example by disqualifying or capping bare-component-only candidates when the query contains that literal flag? Please add persisted public-recall regressions for `importance_weight=1` and the corresponding resolved config/env path, on both working and episodic recall.

The FTS parser safety, default-weight flag behavior, symbolic handling, and cache-version coverage otherwise look good. I would keep Polyphonic handling out of this PR as a separate feature-mode follow-up.

ddddddddwp:
Hi @dplush,

Thank you for pointing out the remaining configurable-ranking issue.

I’ve updated the current head (`fab565f`) so literal-flag precedence is now a selection invariant. Candidates matching only bare flag components are excluded before `top_k`, so importance-only ranking can no longer place the bare `force` distractor above an exact `--force` match.

I also added persisted public-recall regressions for `importance_weight=1`, covering explicit arguments and the resolved config/environment paths for both working and episodic recall.

The separate Polyphonic engine remains unchanged in this PR. I’ll follow up on that behavior in a separate PR as suggested.

Local affected tests and lint pass.

dplush:
@AxDSan Heads-up before merge: #750 is ready on `fab565f` after a fresh high-reasoning recall gate. It fixes FTS-safe technical queries while preserving literal flag precision: `--force` now cannot be outranked by a bare `force` hit, including under YAML/env-resolved importance-only weighting, across working and episodic recall. Current CI, CLA, and checks are green; the scope is limited to recall/tokenization, cache versioning, tests, and an Unreleased note.

dplush (approved):
Approved on the current head after an independent recall merge gate. Literal flags retain precedence over bare components under default and YAML/env-resolved importance-only weighting in working and episodic recall; FTS safety, cache v4 migration, tests, CI, and the Unreleased note are all verified.

PR #751, opened 2026-08-14
fix: status reports interpreter mismatch instead of version mismatch (#736)

### LINKED ISSUE #736: [BUG] mnemosyne-hermes status reports "Python version MISMATCH" while comparing interpreter paths, not versions
## Description

Following up from a [comment](https://github.com/mnemosyne-oss/mnemosyne/issues/709#issuecomment-5262617556) by @dplush on #709.

`mnemosyne-hermes status`  checks whether the install and Hermes use the same interpreter by comparing two file paths, then reports the result as a version mismatch. Different path and different version are not the same condition, and the usual way they diverge is pretty common: create two virtualenvs from one base Python and their bin/python paths differ while their versions are identical. `status` then warns about a version difference that is not there.

`integrations/hermes/src/mnemosyne_hermes/install.py:2255-2257` on `main`:

```python
if state.mode == "symlink" and hermes_python.resolve() != Path(sys.executable).resolve():
    print("  ⚠ Python version MISMATCH! Install and Hermes use different Python versions.")
    print(f"  → Run: {_ver.split()[1]}" if " " in _ver else "")
```

Nothing in the predicate looks at a version and there are two defects in that statement:

1. **Wrong noun.** The condition detects *different interpreters*; the message reports *different versions*. A user whose two environments are both 3.12.13 is told they are not, and sends them looking for a Python upgrade that will not help.
2. **The remediation line is not a command.** `_ver.split()[1]` is the bare version number, so the output is `→ Run: 3.12.13`. When `_ver` has no space (an empty or unexpected `--version` result) the ternary makes `print` emit a blank line instead.

## Reproduction Steps

Two venvs over one base interpreter, `status` pointed at one with the other selected as Hermes' Python:

```python
base = tmp / "base" / "bin" / "python3.12"          # echoes "Python 3.12.13"
a = make_venv(tmp / "a", python_target=base)        # bin/python -> base, + pyvenv.cfg
b = make_venv(tmp / "b", python_target=base)
assert a.resolve() == b.resolve() == base

install._find_hermes_python = lambda **kw: b
install.sys.executable = str(a)
install.main(["status"])
```

## Actual Behavior

```
  This Python: /tmp/work/a/bin/python (3.12.13)
  Hermes' Python: /tmp/work/b/bin/python (Python 3.12.13)
  ⚠ Python version MISMATCH! Install and Hermes use different Python versions.
  → Run: 3.12.13
```

Both interpreters report 3.12.13 on the two lines immediately above the warning that says they differ.

## Expected Behavior

The warning should name what was actually compared, and the follow-up line should be something the user can run. Roughly:

```
  ⚠ Different Python interpreters! The install and Hermes are not the same environment.
  → Run: /tmp/work/b/bin/python -m pip install -U 'mnemosyne-hermes[all]'
```

The `FIX:` line in `hermes_plugin_entry()`
(`integrations/hermes/src/mnemosyne_hermes/__init__.py:3284`) already emits exactly that command shape for the same condition, so this is aligning `status` with a form that exists.

## Scope

The same comparison exists at several sites; I've documented a proposed action for each:

| Site | Wording | Action |
|---|---|---|
| `install.py:2256-2257` | "Python version MISMATCH", non-command hint | fix both |
| `install.py:2262-2264` | "Hermes Python vs install Python mismatch" | already correct, leave |
| `__init__.py:3269` | comment says "version mismatch diagnostics" | comment only; output and FIX line are correct |

Note: `integrations/hermes/tests/test_find_hermes_python.py` asserts on the literal string `MISMATCH` (and lowercase `mismatch` in the negative case), so a wording change needs those two assertions updated.

## Environment

- **OS:** macOS 15 (Darwin 25.6.0)
- **Python version:** 3.12.13
- **Mnemosyne version/commit:** 3.15.1, `9e2dc16`
- **Installation method:** `pip install -e .`

Present on `main` today. #711 does not introduce it and does not fix it: that PR replaces the `.resolve()` comparison with an identity comparison, which makes this warning fire in *more* cases, all of them same-version.

ddddddddwp:
## Summary

Fixes #736. `mnemosyne-hermes status` compared interpreter **paths** but reported a **version** mismatch, and the follow-up line printed a bare version number instead of a runnable command.

## Changes

`integrations/hermes/src/mnemosyne_hermes/install.py` (status command):

- Message now names what was actually compared: `⚠ Different Python interpreters! Install and Hermes are not the same environment.` (was `Python version MISMATCH! ... different Python versions.`).
- Follow-up line is now a runnable command, aligned with the existing `FIX: Run: ...` shape in `hermes_plugin_entry()` (`__init__.py`):
  `→ Run: {hermes_python} -m pip install -U 'mnemosyne-hermes[all]'`
  (was `→ Run: 3.12.13`, a bare version number, or a blank line when `_ver` had no space).

The other sites documented in the issue (`install.py` "Hermes Python vs install Python mismatch" line and `__init__.py` FIX line) were already correct and are left untouched.

## Verification

- New regression test `test_status_mismatch_names_interpreters_and_emits_a_runnable_command` in `integrations/hermes/tests/test_install_status.py` reproduces the issue's two-venv-over-one-base scenario: asserts the new wording, that `version MISMATCH` is gone, and that the emitted line is the real pip command (not `→ Run: 3.12.13`).
- Full `integrations/hermes/tests/` suite: 399 passed, 1 skipped (with `pip install -e ./integrations/hermes` as CI does).
- `ruff check --select E9,F63,F7,F82` (CI lint selection): clean on both files.

## Summary

- Corrects `mnemosyne-hermes status` to report different Python interpreters instead of a false Python version mismatch.
- Provides a runnable, shell-quoted command to install `mnemosyne-hermes[all]` into Hermes’ Python environment.
- Adds regression coverage for separate virtual environments that use the same base Python interpreter.
- Preserves the core memory architecture. No changes affect working, episodic, or BEAM memory tiers; retrieval; consolidation; veracity; sync; or benchmark methodology.
- Preserves local-first guarantees and privacy posture. The change adds no external data flow or new service dependency.
- Improves the Hermes CLI integration surface and long-term maintainability by aligning diagnostics with the actual environment comparison.
- This is the right call because it makes remediation actionable without changing memory behavior or weakening local-first design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
+- **`mnemosyne-hermes status` now reports the real interpreter mismatch (#736).** The warning compared interpreter paths but claimed a Python version mismatch and printed a bare version number instead of a runnable fix; it now compares the Hermes and installer environments and emits a shell-quoted `→ Run: <python> -m pip install -U 'mnemosyne-hermes[all]'` command.
 - **MCP SSE authentication rejects malformed non-ASCII bearer tokens with `401` instead of returning a server error (#739).**
 - **API embedding failures now leave a redacted diagnostic trace (#735).** Final HTTP, network, and invalid-response failures still degrade to keyword-only retrieval, but now log the endpoint and safe error class or status without request content, API keys, URL userinfo, query strings, or fragments.
 - **Hermes tool discovery now honors `memory.mnemosyne.tools` (#725).** Tools outside the configured allowlist are no longer advertised through Hermes provider schemas before provider initialization.
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -1028,6 +1028,17 @@ def _iter_mnemosyne_profiles(hermes_home_path: str | Path | None = None) -> list
     return selected
 
 
+def _hermes_python_mismatch(hermes_python: Path) -> bool:
+    """Return whether Hermes runs in a different environment from the installer.
+
+    Compares the environment roots (``sys.prefix`` vs the Hermes interpreter's
+    ``parent.parent``) instead of resolved interpreter paths: separate venvs
+    created over one base interpreter resolve to the same binary while keeping
+    different site-packages, so a resolved-path comparison misses the mismatch.
+    """
+    return Path(hermes_python).parent.parent != Path(sys.prefix)
+
+
 def _link_profile(profile_home: Path, source: Path, *, force: bool = False) -> Optional[Path]:
     """Symlink ``profile_home/plugins/mnemosyne`` to source. Idempotent.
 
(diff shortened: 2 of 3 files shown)


dplush (approved):
Current-head review complete: the env-root comparison, copy-paste-safe repair command, focused status tests, and diff hygiene all checked clean on de445d9.

PR #757, opened 2026-08-15
fix: allow updating global-scope memories from any session

ghtmarco:
## Summary

`update_working()` and `Mnemosyne.update()` were session-locked, so a **global-scope** memory (e.g. a durable user fact) could never be updated from a later session: `update()` reported `not_found` even though `get()` and `invalidate()` could reach the row.

## Root cause

- `beam.update_working()`: `UPDATE working_memory ... WHERE id = ? AND session_id = ?` — the only write path that does **not** use the `(session_id = ? OR scope = 'global')` authorization pattern already used by `get()` and `invalidate()`.
- `Mnemosyne.update()`: reported success based on a `UPDATE memories ... WHERE id = ? AND session_id = ?` against the **legacy** table (which no longer holds BEAM-era rows), while the result of the real BEAM update was discarded.

## Fix

- `beam.update_working()`: same authorization pattern as `get()`/`invalidate()` — `(session_id = ? OR scope = 'global')`.
- `Mnemosyne.update()`: BEAM `working_memory` is now the primary store and its result determines the return value; the legacy `memories` table is a best-effort mirror, still scoped to the caller's session. Returns `False` when nothing matched (previously it could claim success while the BEAM row was untouched).

## Security

Session-scoped memories remain **not** updatable from other sessions — only the caller's own session rows and `scope='global'` rows are writable, matching the existing read/invalidate boundary.

## Tests

Added `TestUpdateCrossSessionAuthorization` in `tests/test_beam.py`:

- `test_update_global_memory_from_another_session` — global memory updatable from any session (was the bug)
- `test_update_session_memory_from_another_session_denied` — session-scoped memory still protected
- `test_mnemosyne_update_global_memory_from_another_session` — `Mnemosyne.update()` propagates cross-session

Verified with `MNEMOSYNE_NO_EMBEDDINGS=1 python -m pytest tests/test_beam.py -k "TestUpdateCrossSessionAuthorization or TestUpdateRefreshesDerivedState"` (existing update tests + new ones pass).

## Summary

- Global-scope BEAM memories support authorized cross-session updates.
- Session-scoped memories remain protected from other sessions.
- `Mnemosyne.update()` uses BEAM `working_memory` as the primary store.
- Legacy `memories` rows receive best-effort session-scoped mirrors.
- Failed, empty, unknown, or unauthorized updates return `False`.
- `MEMORY_UPDATED` events emit only after a successful BEAM update.
- Scope migration defaults and backfills existing rows as `session`.

## Architectural impact

This change strengthens the working-memory tier and makes BEAM the authoritative update path. It does not change episodic memory, retrieval, consolidation, veracity, synchronization, or benchmark methodology.

The authorization rule is the right call. It supports shared global memory while preserving session privacy. The local-first design remains unchanged because updates use local stores and add no remote dependency.

Hermes, MCP, and CLI integration surfaces do not change. The BEAM-first flow improves maintainability by separating the primary store from the legacy compatibility mirror and by returning a clear update result.

Regression tests cover cross-session authorization, denied updates, event suppression, update propagation, empty and unknown updates, batch behavior, and migration backfill.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -4493,7 +4493,7 @@ class BeamMemory:
             return False
         params.extend([memory_id, self.session_id])
         cursor.execute(
-            f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ? AND session_id = ?",
+            f"UPDATE working_memory SET {', '.join(updates)} WHERE id = ? AND (session_id = ? OR scope = 'global')",
             params
         )
         affected = cursor.rowcount
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1083,10 +1083,18 @@ def init_beam(db_path: Path = None):
     # --- Migration: temporal validity + scope (v2.2) ---
     _add_column_if_missing(conn, "working_memory", "valid_until", "TIMESTAMP DEFAULT NULL")
     _add_column_if_missing(conn, "working_memory", "superseded_by", "TEXT DEFAULT NULL")
-    _add_column_if_missing(conn, "working_memory", "scope", "TEXT DEFAULT 'global'")
+    if _add_column_if_missing(conn, "working_memory", "scope", "TEXT DEFAULT 'session'"):
+        # Pre-scope rows only carried a session_id -- they were session-private.
+        # The ADD COLUMN default would backfill them as 'global', which would
+        # make them readable AND (with cross-session update authorization)
+        # writable from any other session. Normalize to 'session'.
+        conn.execute("UPDATE working_memory SET scope = 'session' WHERE scope = 'global'")
+        conn.commit()
     _add_column_if_missing(conn, "episodic_memory", "valid_until", "TIMESTAMP DEFAULT NULL")
     _add_column_if_missing(conn, "episodic_memory", "superseded_by", "TEXT DEFAULT NULL")
-    _add_column_if_missing(conn, "episodic_memory", "scope", "TEXT DEFAULT 'global'")
+    if _add_column_if_missing(conn, "episodic_memory", "scope", "TEXT DEFAULT 'session'"):
+        conn.execute("UPDATE episodic_memory SET scope = 'session' WHERE scope = 'global'")
+        conn.commit()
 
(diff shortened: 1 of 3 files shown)


ghtmarco:
Thanks for the thorough review — all findings addressed in b5dbc45:

**🔒 Security (scope migration)** — Agreed, this was a real hole. `ADD COLUMN scope DEFAULT 'global'` backfills every pre-scope row as global, so once `update_working()` authorizes cross-session global updates, old session-private rows become writable from any other session. Fixed in the migration: when the `scope` column is first added, all rows are normalized to `'session'`, and the column default is now `'session'` (a bulk-import `INSERT` at beam.py:3754 omits `scope` and was silently creating global rows). Note: DBs that already ran the v2.2 migration keep their current `scope` values — a one-time backfill for those would need a heuristic to distinguish migrated rows from intentionally-global ones; happy to add a follow-up if you want one.

**💬 Mnemosyne.update() return semantics** — Agreed. The return value and `MEMORY_UPDATED` event now depend on the BEAM result only; the legacy-table mirror remains best-effort and no longer drives the outcome.

**🎯 Test coverage** — Added: `Mnemosyne.update()` cross-session denial (content preserved + no `MEMORY_UPDATED` event), empty update and unknown-ID both returning `False` without emitting, and a migration regression test proving pre-scope rows are backfilled to `'session'` and stay non-writable cross-session.

**🩺 Stability (transaction held during embedding)** — Pre-existing behavior, not introduced by this PR (the diff doesn't touch the transaction structure). Worth a separate fix/issue: commit the row before embedding, then persist the embedding in a short guarded transaction. I left it out to keep this PR minimal.

Verified locally: 14/14 checks pass (cross-session global update, denial boundaries, event suppression, migration backfill, batch path).

AxDSan:
Closing in favor of #582, which fixes the same update_working boundary with a smaller change and no migration edit. Thank you for the report and the tests; the cross-session test cases are worth carrying over and I have asked for them on #582.

The other change in this PR, flipping the scope column default from global to session and rewriting existing rows on first migration, is a real concern but a separate one: it changes what pre-scope memories mean for every existing database. If you want to pursue it, please open an issue with the reasoning so it can be decided on its own rather than inside an update() fix.

PR #759, opened 2026-08-15
fix(core): normalize valid_until writes and comparisons to aware UTC (#525)

### LINKED ISSUE #525: bug(core): valid_until mixes local wall-clock timestamps with SQLite UTC comparisons
## Bug Description

Mnemosyne mixes naïve local timestamps with SQLite UTC timestamps when handling `valid_until`. This makes activity/expiry checks disagree by the host's UTC offset.

### Write path

`BeamMemory.invalidate()` uses:

```python
now = datetime.now().isoformat()
```

and stores that naïve local wall-clock value in `working_memory.valid_until`.

### Read/repair paths

Some code instead uses SQLite UTC semantics:

```sql
CURRENT_TIMESTAMP
julianday('now')
```

For example, the repair and doctor logic tests expiry using `julianday(valid_until) > julianday('now')`.

### Reproduction

On a host configured for Europe/Copenhagen (UTC+02:00 in summer):

- local clock: `2026-07-23T22:50:47+02:00`
- UTC clock: `2026-07-23T20:50:47+00:00`
- a simple invalidation wrote: `valid_until='2026-07-23T22:40:05.389817'`

Normal recall/context paths compare against another local `datetime.now().isoformat()` and treat the record as expired immediately. SQLite-based diagnostic/repair paths compare the same naïve value with UTC `julianday('now')` and treat it as active until roughly 00:40 local time.

Thus different Mnemosyne surfaces disagree for approximately the local UTC offset, and behavior changes with daylight-saving time.

## Expected Behavior

All memory timestamps and comparisons should follow one explicit convention, ideally timezone-aware UTC instants.

## Suggested Direction

- Write timestamps as aware UTC ISO-8601 values, or consistently use SQLite UTC at the write boundary.
- Normalize legacy naïve timestamps deliberately and document the migration rule.
- Replace lexical/implicit comparisons with one centralized temporal helper or a consistently normalized SQLite expression.
- Add tests under a non-UTC `TZ`, including a DST-offset case, covering invalidation, recall/context filtering, doctor, and repair.

## Environment

- Host timezone: Europe/Copenhagen (UTC+02:00 at reproduction time)
- Mnemosyne core: 3.14.0
- mnemosyne-hermes: 0.5.0
- Hermes Agent: 0.19.0
- Python: 3.13.5
- Linux

ddddddddwp:
## Summary

Fixes #525. `BeamMemory.invalidate()` wrote a **naive local wall-clock** ISO value (`datetime.now().isoformat()`) into `valid_until`, while the SQLite-side surfaces (doctor `julianday('now')`, repair, MCP `validate` via `CURRENT_TIMESTAMP`) compare in UTC. On any host not at UTC+0, expiry checks disagreed by the host's UTC offset, and the offset shifted with DST.

## Contract (conservative)

- **Writes**: aware UTC ISO-8601 (`datetime.now(timezone.utc).isoformat()`).
- **Python read paths**: all `valid_until > ?` comparisons now pass aware UTC now, converging with SQLite's UTC semantics.
- **Legacy naive rows** (no offset, written pre-fix): interpreted as UTC, matching what SQLite `julianday` already does; both surfaces therefore agree.
- **Date-only `valid_until` inputs** (e.g. `2099-12-31`): unchanged pass-through API semantics.

## Changes

- `mnemosyne/core/beam.py`: `invalidate()` write sites + 6 recall/context comparison params (`3026`, `4102`, `5957`, `6225`, `6347`, `6455`) now use `datetime.now(timezone.utc)`.
- `mnemosyne/core/memory.py`: pattern-detection `valid_until` comparison now aware UTC.
- `mnemosyne/core/polyphonic_recall.py`: `now_iso` for EM/WM `valid_until` filters now aware UTC.
- `tests/test_beam.py`: 4 new regressions under a non-UTC host TZ (`Europe/Copenhagen`): aware-UTC stamp, cross-surface agreement (Python vs `julianday`), legacy naive interpretation, date-only pass-through.
- `tests/test_hermes_memory_provider_audit.py`: naive-now bounds updated to aware UTC to match the new stored format.
- `CHANGELOG.md`: `[Unreleased]` → Fixed entry.

Not changed (out of scope): `last_recalled`/recency bookkeeping, degradation cutoffs, and health staleness use internally-consistent timestamps and are not part of the `valid_until` contract.

## Verification

- New tests: 4 passed under `TZ=Europe/Copenhagen`
- Focused suites (`test_beam.py`, `test_memory_lifecycle.py`, `test_hermes_memory_provider_audit.py`, `test_hermes_memory_provider_validation.py`, `test_patterns.py`, `test_mcp_server.py`, `test_beam_e5_polyphonic_recall.py`, `test_enhanced_recall_cache.py`): 299 passed, 1 skipped
- Full `tests/` run: 2607 passed, 6 skipped, 4 subtests passed (2 pre-existing environment-dependent failures excluded: `test_config_profiles::test_apply_writes_to_config` and `test_handle_batch_failure_rolls_back`, both failing on clean `main` in this environment)
- `ruff check` on changed files: all passed
- `git diff --check`: clean

## Checklist

- [x] Issue linked (`Closes #525`)
- [x] Type of change: bug fix
- [x] Tests added/updated
- [x] Changelog updated
- [x] Version bump: N/A (release versions are prepared separately)

## Summary

- Standardizes `valid_until` writes and Python comparisons on aware UTC timestamps.
- Normalizes timestamps at remember, consolidation, import, and sync boundaries.
- Interprets legacy naïve timestamps as UTC.
- Normalizes lowercase-`t`, offset-bearing, and space-separated timestamps.
- Preserves exact date-only `valid_until` behavior.
- Adds regression coverage across invalidation, expiry filtering, sync, episodic writes, entity recall, `get_all_memories`, vector voice, Hindsight import, and polyphonic recall under `Europe/Copenhagen`.

## Architectural impact

- Aligns expiry behavior across working memory, episodic memory, BEAM invalidation, context, vector search, recall tracking, polyphonic recall, imports, and sync.
- Uses chronological SQLite comparisons and parsed Python datetimes.
- Prevents expired or invalidated memories from reappearing through alternate retrieval paths.
- Strengthens the consolidation pipeline and sync layer without changing their boundaries or protocols.
- Does not change tier boundaries, retrieval strategies, or the veracity system.
- Does not change Hermes, MCP, or CLI contracts. Their expiry behavior becomes more consistent.
- Preserves local-first operation. The PR adds no network, remote storage, or telemetry dependency.
- Improves privacy and predictability by removing host-local timezone and DST ambiguity from memory validity checks.
- Establishing one explicit UTC convention across memory lifecycle surfaces is the right call for maintainability.
- Other unrelated naïve timestamps remain outside this focused fix.


dplush (changes requested):
Blocking: this still leaves supported import paths able to persist noncanonical `valid_until` strings, while several read paths remain lexical SQLite comparisons against the new aware-UTC value.

Examples:
- `BeamMemory.import_from_dict()` writes raw `item.get("valid_until")` into working and episodic rows (`beam.py:9345-9363`, `9429-9441`).
- The Hindsight episodic importer does the same (`importers/hindsight.py:360-384`).
- These values are then filtered by SQL `valid_until > ?` predicates, including `get_context`, working/episodic recall, `get_all_memories`, and polyphonic vector SQL.

This is observable: against `2026-08-15T12:30:00+00:00`, SQLite lexically rejects `2026-08-15T11:00:00-02:00` (chronologically 13:00Z) and `2026-08-15 13:00:00` (legacy naive ISO with a space), even though both are still active.

Please either normalize every supported persistence/import boundary before the lexical predicate is relied on, or make every read filter chronological and add import-regression coverage for offset-bearing and legacy-naive values. Without that, imported rows can silently disappear from recall/context while still valid.

dplush:
Re-reviewed the current head `f9727de` against the earlier UTC, SQLite comparison, polyphonic, and import-path findings.

The fix now normalizes supported write and import boundaries to UTC, uses chronological SQLite comparisons across the affected recall paths, and preserves the documented date-only behavior. The earlier bot findings are addressed on the current diff.

I also ran the focused regression matrix locally: 237 passed, 14 skipped. The full GitHub CI matrix is green and the PR is clean.

The previous changes-requested review targets an older head. From my review, this is ready for a fresh maintainer approval and merge decision.

dplush:
Thanks for the thorough update. I found one remaining edge case before merge.

`datetime.fromisoformat()` accepts a lowercase `t` separator, but `_normalize_valid_until()` currently only attempts normalization when the value contains uppercase `T` or a space.

For example:

```text
2099-12-31t23:00:00-02:00
```

is a valid future instant and is accepted by the Python-side parser. It currently remains unnormalized, while SQLite returns `NULL` for `julianday(...)`. The SQL recall paths then exclude it, but the Python/polyphonic filter treats it as active.

Could you please preserve the pass-through only for an exact `YYYY-MM-DD` value, and otherwise attempt UTC normalization? A small regression covering lowercase `t` through both an SQL-backed recall path and the Python-side filter would close this gap.

ddddddddwp:
Thanks for catching this. I limited the pass-through to an exact `YYYY-MM-DD` value, and everything else now goes through UTC normalization, so lowercase-`t` forms like `2099-12-31t23:00:00-02:00` get canonicalized to aware UTC on write. The SQL recall paths and the Python-side filter now agree. Added a regression that pushes a lowercase-`t` value through `remember()` and asserts it surfaces on both the SQL-backed recall and the Python filter.

ddddddddwp:
Good catch. Updated the changelog to say "any other parseable form ... is normalized; unparseable values pass through unchanged", which matches what `_normalize_valid_until()` actually does.

dplush (approved):
Approved on current head `294c6a8`.

The lowercase-`t` regression is fixed: only exact date-only values bypass normalization; other parseable timestamp forms are canonicalized to UTC. The independent current-head review found no remaining technical blocker. Full CI is green.

Issue #726, opened 2026-08-13
[BUG] Legacy and BEAM dual writes can leave permanently divergent stores after an intermediate failure


dplush:
## Summary

The compatibility layer writes the Legacy and BEAM representations as separate committed operations. If the second operation fails, one representation can remain durable while the other is absent or stale. This creates a permanently divergent local database rather than an all-or-nothing memory mutation.

## Evidence

Audited on current `main` `85a4b88b506252ac522285f35ff86792731c65f3`:

- `Mnemosyne.remember()` persists through `BeamMemory.remember()` before the Legacy write: `mnemosyne/core/memory.py:426-455`.
- `forget()` and `update()` also cross the two representations with separate mutation boundaries: `mnemosyne/core/memory.py:605-643`.
- Existing tests cover successful compatibility writes but not an exception between the layers: `tests/test_beam.py:1206-1232`.

A controlled local fault injection at the second write produced `working_memory=1` and `memories=0`. The inverse ordering in the other paths can likewise leave one side changed after the companion mutation fails.

## Expected

A public memory mutation either updates every supported representation consistently or leaves them all unchanged and reports the failure.

## Actual

A failure between committed compatibility writes can expose contradictory state to callers that use different read paths.

## Decision needed before implementation

Please confirm the intended compatibility boundary for the Legacy `memories` table. A safe fix needs either:

1. one caller-owned SQLite transaction over BEAM and Legacy writes on the same connection, or
2. an explicit retirement/migration of the Legacy compatibility write.

Do not add compensating best-effort deletes after a partial commit; those create another failure mode.

## Suggested acceptance tests

For `remember`, `update`, and `forget`, inject a failure after the first representation mutates and assert:

- both representations retain their pre-operation state;
- no transaction remains open;
- success behavior stays backward compatible.

## Related work

PR #582 touches `memory.py` for global cross-session update semantics, but does not establish a general Dual-Write atomicity/rollback contract. Keep this issue separate unless the maintainer explicitly chooses a shared design.

dplush:
Additional evidence: the public MCP batch path has the same underlying dual-write transaction-ownership problem, and needs to be part of #726's contract rather than patched incidentally in #582.

Read-only audit on current main / a local #580 candidate found:

- `apply_beam_batch()` wraps operations in `_deferred_commits(beam.conn)`, whose success path unconditionally calls `_real_commit()`. A batch therefore cannot safely run within an already-open caller-owned transaction.
[...]

PR #760, opened 2026-08-15
fix: preserve caller transactions in memory batches

dplush:
## Summary

Establishes the first #726 foundation slice: `mnemosyne_batch` now respects transaction ownership instead of committing or rolling back an already-open caller transaction.

- uses connection-local, unique savepoints for caller-owned and nested batch/guard scopes
- prevents guarded `forget` failures from rolling back the caller transaction
- prevents vector ingestion from bypassing active commit deferral via `_real_commit()`
- keeps audit events deferred until the batch scope succeeds

## Proof

Focused public batch regressions cover:

- caller-owned `remember -> update`, followed by outer rollback
- failure inside `forget_working()` after the guard is entered, preserving caller work and emitting no audit event
- nested success/failure savepoint behavior and deferral-flag restoration
- active vector path without an inner real commit or premature publication

Local matrix: `17 passed` (`test_mnemosyne_batch_tool.py` + `test_batch_transaction_owner_726.py`), Ruff and `git diff --check` clean.

## Scope

This does **not** solve the separate wrapper Legacy + BEAM dual-write atomicity contract. That remains tracked by #726 as a subsequent slice.

Refs #726

## Summary

- Preserves caller-owned transactions in `mnemosyne_batch`.
- Uses connection-local savepoints for nested batches and guarded operations.
- Prevents failed operations from committing, rolling back, or corrupting caller data.
- Defers audit events until batch success.
- Preserves commit deferral during vector ingestion.
- Restores nested deferral state after success or failure.

## Architectural impact

- Strengthens local-first behavior by keeping transaction control with the caller.
- Improves atomicity for working-memory and vector-storage writes.
- Protects privacy by keeping transaction and audit-event handling local.
- Does not change tiered-memory behavior across working, episodic, or BEAM memory.
- Does not change retrieval, consolidation, veracity, synchronization, Hermes, MCP, or CLI interfaces.
- Does not change the separate wrapper Legacy and BEAM dual-write contract tracked by `#726`.
- Adds regression coverage for transaction ownership, nested savepoints, guarded `forget_working()` failures, audit suppression, deferral restoration, and vector ingestion.
- The savepoint model is the right call for nested memory operations because it isolates failures without taking ownership of the caller’s transaction.
- The changes improve maintainability by making transaction ownership and nested deferral state explicit.
- No change erodes the local-first design or introduces a remote data path.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -1305,58 +1305,61 @@ def _guarded_transaction(conn: sqlite3.Connection):
 
 @contextlib.contextmanager
 def _deferred_commits(conn: sqlite3.Connection):
-    """Suppress nested commit() calls so the caller can wrap many
-    sub-helpers in a single transaction.
+    """Defer nested commits without stealing a caller-owned transaction.
 
-    Pairs with `_BeamConnection`'s `_defer_commit` flag. If the
-    passed connection isn't a `_BeamConnection` (e.g., a test
-    constructed `BeamMemory` with a raw sqlite3 connection, or a
-    legacy caller built its own conn), the context manager degrades
-    to a no-op -- inner commits still fire, performance regression
-    isn't fixed for that code path but correctness is preserved.
-
-    Threading: `_BeamConnection._defer_commit` is per-connection.
-    BeamMemory uses thread-local connections (see _get_connection),
-    so the flag is visible only to the calling thread. A future
-    refactor that shares the connection across threads would need
-    a lock here.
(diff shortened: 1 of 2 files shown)


dplush:
Heads-up before merge: this is the narrow #726 BEAM transaction-ownership foundation only. It preserves caller-owned SQLite transactions with connection-local savepoints, keeps vector writes inside the deferred batch boundary, and adds focused regressions. Wrapper legacy/BEAM dual-write atomicity remains explicitly out of scope for a separate follow-up. CI, CLA, CodeRabbit, independent review, and the final current-head gate are green.

PR #768, opened 2026-08-16
fix: suppress expected FastEmbed pooling warning

jonasrombach:
## Summary
- suppress FastEmbed’s expected mean-pooling migration warning for `paraphrase-multilingual-MiniLM-L12-v2`
- restrict the warning filter to `UserWarning`
- add a fresh-process regression test

## Verification
- `pytest tests/test_embedding_optout.py tests/test_multilingual_local_recall.py -q` (31 passed)
- Ruff `F,RUF022` checks passed
- real FastEmbed model load passed under `-W error` with no warning

## Summary

- Suppresses the expected FastEmbed mean-pooling migration warning for `multilingual-e5-large` and `paraphrase-multilingual-MiniLM-L12-v2`.
- Restricts the filter to `UserWarning`.
- Adds a fresh-process regression test with no stderr output.

## Architectural impact

- No changes to retrieval, consolidation, veracity, sync, Hermes, MCP, or CLI integration.
- No changes to privacy posture or local-first guarantees.
- No benchmark methodology changes.
- Improves maintainability by documenting and testing expected model behavior without hiding unrelated warning types.

## Verification

- 31 specified tests passed.
- Ruff `F,RUF022` checks passed.
- FastEmbed model loading succeeded under `-W error` without warnings.


dplush:
Thanks for the narrow suppression. The production change is correct, but the regression currently turns unrelated stderr output into a failure: Python 3.11 emits an ONNX Runtime PCI-discovery line, while `assert result.stderr == ""` requires all stderr to be empty.

Could you please make one focused follow-up batch?

1. Assert that the specific matched FastEmbed `UserWarning` is absent, rather than asserting empty stderr.
2. Cover both configured model strings: `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` and `intfloat/multilingual-e5-large`.
3. Add a matching non-`UserWarning` control that remains visible, so the category boundary is explicit.
4. Correct the comment from FastEmbed `>=0.7` to `>=0.6`.
5. Add a concise factual `CHANGELOG.md` entry under Unreleased for the reduced expected FastEmbed startup/model-load noise.

Then please rebase onto current `main` (post-#750) and rerun CI. The change stays a good small logging-only fix; no embedding, model-selection, indexing, or recall behavior should change.

AxDSan (changes requested):
The suppression itself is fine. The test asserting it is over-specified, and that is the whole failure. One line fixes it.

## What is actually failing

`assert result.stderr == ""` demands a **completely silent** stderr. On GitHub runners onnxruntime emits an unrelated warning before anything of ours runs:

```
[W:onnxruntime:Default, device_discovery.cc:134 GetPciBusId] Skipping pci_bus_id
for PCI path at "/sys/devices/LNXSYSTM:00/LNXSYBUS:00/ACPI0004:00/MSFT1000:00/..."
because filename "..." did not match expected pattern of [0-9a-f]+:[0-9a-f]+:...
```

That is the runner's Hyper-V PCI topology (`MSFT1000`), which onnxruntime cannot parse into a bus id. Nothing to do with FastEmbed mean pooling, and nothing this PR causes. The rest of the run is clean: **1 failed, 3417 passed**.

It passes on my machine because a real PCI path parses fine, and fails on CI because a virtualised one does not. Any assertion of the form "stderr is empty" will keep finding new ways to be false as onnxruntime adds diagnostics.

## The fix

Assert the absence of the warning you are actually suppressing, not the absence of all output:

```python
assert "mean pooling" not in result.stderr   # or the exact marker string
```

That still fails loudly if the suppression regresses, and stops failing when an unrelated component talks to stderr.

## Worth knowing

This is the same family as **#850**, `blkid: not found` from onnxruntime on minimal Linux, filed by @dplush from the 4.0.0b1 canary. Two different environments, two different unparseable-hardware complaints, both from onnxruntime, both landing on stderr. If #850 ends up suppressing that one, the two probably want the same mechanism rather than separate patches, and this test would be the natural place to assert both are gone.

I also updated this branch from `main` before re-running, since its last run predated the CI hang fix in #867 by ten days. That was not the cause; the failure reproduces on current main.

Issue #728, opened 2026-08-13
[BUG] MCP tools/list advertises schemas without corresponding handlers


dplush:
## Summary

The MCP server publishes schemas for eight tools that its dispatcher cannot execute. Clients can select them from `tools/list`, but every `tools/call` fails with `Unknown tool`.

## Evidence

Audited on current `main` `85a4b88b506252ac522285f35ff86792731c65f3`:

- `ALL_TOOL_SCHEMAS` includes `mnemosyne_triple_end`, `mnemosyne_sync_push`, `mnemosyne_sync_pull`, `mnemosyne_sync_status`, `mnemosyne_persona_promote`, `mnemosyne_persona_demote`, `mnemosyne_persona_list`, and `mnemosyne_persona_reinforce`: `mnemosyne/tool_schemas.py:858-870`.
- `_TOOL_HANDLERS` does not define handlers for those names: `mnemosyne/mcp_tools.py:1105-1134`.
- Missing entries become `ValueError: Unknown tool: ...`: `mnemosyne/mcp_tools.py:1151-1155`.
- The server publishes the schema list without handler filtering: `mnemosyne/mcp_server.py:113-125`.

A controlled local `tools/list`/dispatch probe returned 36 advertised tools; all eight names above failed at dispatch.

## Expected

Every MCP-advertised tool is callable, or unsupported tools are absent from `tools/list`.

## Actual

The advertised public surface is larger than the callable surface, producing deterministic client errors.

## Decision needed

Please choose one MCP contract:

1. publish only names with MCP handlers, or
2. implement handler parity for the eight schemas.

The first option is the smallest safe correction. The second exposes additional APIs and should be reviewed as a feature/compatibility decision.

## Suggested regression

At the MCP boundary, assert the advertised name set matches the dispatcher handler set, then run a `tools/list` to `tools/call` smoke for every published tool using safe arguments or a structured validation-path assertion.

## Related work

PR #723 addresses a different missing MCP handler: `mnemosyne_forget_canonical`. It does not cover this eight-tool mismatch.

dplush:
Maintainer decision requested: I recommend option 1, publishing only schemas with an MCP handler.

These eight tools have never been callable through MCP, so filtering them from `tools/list` restores the advertised-to-callable contract without adding new public MCP APIs.

[...]

PR #770, opened 2026-08-16
feat(mcp): advertise optional bank parameter in all bank-aware tool schemas (#769)

### LINKED ISSUE #769: MCP server tool schemas omit the optional bank parameter — #173 fixed only the hermes_plugin wrapper
## Summary

`#173` (closed, fixed 2026-05-25, commit `fd736761ac`) added the `bank` parameter to the tool schemas in the **`hermes_plugin/`** wrapper. The MCP server's own schemas were not part of that fix. As of the latest release **v3.15.1 (2026-07-30)**, none of the server's advertised tool schemas in `mnemosyne/tool_schemas.py` include `bank`, while every handler continues to honor it.

## Verified on v3.15.1

- `mnemosyne/mcp_tools.py`: `_resolve_bank(arguments)` → `arguments.get("bank") or MNEMOSYNE_MCP_BANK or "default"` — all tool handlers route through it.
- `mnemosyne/tool_schemas.py`: `RECALL_SCHEMA`, `REMEMBER_SCHEMA`, `STATS_SCHEMA`, `TRIPLE_ADD_SCHEMA`, `TRIPLE_QUERY_SCHEMA`, `GET_SCHEMA` — **no `bank` property in any of them**.
- Consequence: an MCP client that introspects `tools/list` sees no way to select a bank; an agent calling the tools without the undocumented parameter silently operates on the `default` bank. Multi-instance agent deployments (one server, one private bank per instance) currently rely on client-side code hard-coding the parameter name into the arguments dict.

## Proposed change

1. Add an optional `bank: string` property to the input schemas in `mnemosyne/tool_schemas.py` for every bank-aware tool.
2. Document the resolution precedence in the schema description: explicit `bank` → `MNEMOSYNE_MCP_BANK` env → `default`.
3. Keep it optional — an absent `bank` behaves exactly as today (backward compatible).

## Out of scope

No change to bank-routing semantics (consistent with the maintainer's non-goal on #548).

## Acceptance criteria

- `tools/list` advertises `bank` on the bank-aware tools.
- A schema-driven MCP client can call `mnemosyne_recall` with `bank` targeting a named bank and receive results from that bank only.
- Omitting `bank` still resolves env → `default`; existing tests pass.

## Related

- #173 (plugin-wrapper fix — this issue covers the server half)
- #548 (SDK entry point; routing non-goal)

tiangao88:
Every MCP-callable tool whose handler resolves a memory bank now exposes an optional `bank` parameter in its schema, with resolution precedence documented (explicit `bank` argument -> `MNEMOSYNE_MCP_BANK` env var -> `default`). A shared `BANK_PROPERTY` constant keeps all 23 bank-aware schemas consistent.

Exclusions are explicit and unchanged:
- `mnemosyne_validate` keeps its own `bank` enum (`private|surface`) with a different meaning
- `mnemosyne_shared_*` tools operate on the shared surface and resolve no bank
- non-MCP or non-bank-aware tools gain nothing

## Changes
- **`mnemosyne/tool_schemas.py`**: shared `BANK_PROPERTY` applied to all bank-aware schemas (20 previously missing `bank`, 3 upgraded from ad-hoc entries)
- **`tests/test_tool_surface_parity.py`**: new parity test derives bank-awareness from the handler registry, so any future handler/schema drift fails CI
- **`tests/test_mcp_server.py`**: obsolete "bank is not in the schema" assertions replaced with positive schema checks + a named-bank write/recall smoke test proving same-bank isolation from the default bank
- **`scripts/generate-docs.py`**: docs generator now resolves shared constant references (previously schemas referencing them were silently dropped from the output)
- **`docs/api/tool-schema.mdx`**: regenerated (37 tools, 29 MCP)
- **`CHANGELOG.md`**: entry under Unreleased

## Testing
- `tests/test_mcp_server.py` + `tests/test_tool_surface_parity.py`: 89 passed
- `scripts/verify-docs.py`: OK (generated docs match the code)
- Full `tests/` sweep: green apart from pre-existing test-ordering pollution in `test_hermes_memory_provider.py` (passes in isolation; identical failures on pristine main)

Fixes #769

## Summary

- Adds an optional `bank` parameter to 23 bank-aware MCP tools.
- Documents precedence: explicit `bank`, `MNEMOSYNE_MCP_BANK`, then `default`.
- Preserves default behavior and existing bank-routing semantics.
- Enables schema-driven clients to target named banks through `tools/list`.
- Verifies named-bank write and recall isolation from the default bank.

## Architecture and Local-First Impact

- Does not change tiered memory, retrieval, consolidation, veracity, or sync architecture.
- Does not change storage or network behavior.
- Preserves Mnemosyne’s local-first design.
- Improves named-bank separation without creating a new privacy or security boundary.

## Agent Integration Surfaces

- Aligns MCP schemas with existing handler capabilities.
- Leaves `mnemosyne_validate` and `mnemosyne_shared_*` unchanged.
- Includes no Hermes or CLI changes.

## Maintainability

- Centralizes the shared definition in `BANK_PROPERTY`.
- Adds parity tests for handler bank resolution and schema declarations.
- Updates documentation generation to resolve shared schema constants.
- Regenerates tool-schema documentation and updates the changelog.

This is the right call because it aligns the MCP contract with existing handler behavior while preserving compatibility and local-first guarantees.


tiangao88:
All CodeRabbit review comments addressed in the amended commit `5090b29`:

1. **Correct the omitted-bank behavior description** (inline, `tool_schemas.py:20`) — `BANK_PROPERTY` now states: omit to use `MNEMOSYNE_MCP_BANK` when set, otherwise the default bank; named banks use separate local SQLite files and do not cross-read or cross-write.
2. **Assert that `bank` is not required** — added `assert "bank" not in schema["required"]` to the remember/recall schema tests, plus a parity-collection check that fails if any bank-aware tool ever declares `bank` as required.
3. **Expose bank resolution in the generated API reference** — `scripts/generate-docs.py` now renders a `Description` column in tool parameter tables (matching the existing config tables), so the `bank` precedence + isolation note appears in the regenerated `tool-schema.mdx` (verified via `verify-docs.py`).

Tests: `test_mcp_server.py` + `test_tool_surface_parity.py` — 89 passed locally; CI re-running on the new head.

dplush:
Thanks — the MCP bank contract itself looks good on the current head. Independent checks confirmed the exact 23 normal bank-aware handlers, intentional exclusions, explicit/env/default precedence, named-bank isolation, generated docs, and the focused 89-test matrix.

One small follow-up before approval: #770 introduced a degraded failure path in `scripts/generate-docs.py`. `_dict_keys_of_assignment()` now falls through with `None` when the assignment is absent, because its `return []` ended up after the new helper definitions. That replaces the existing controlled fatal diagnostic with `TypeError: 'NoneType' object is not iterable`.

Could you move `return []` back to the end of `_dict_keys_of_assignment()` and remove the unreachable copy? Please add a focused regression for the absent/unparseable assignment path if practical. The generator still fails closed today, so this is only a narrow diagnostics fix; no change to the bank schema scope is needed.

The schema-driven parity guard is otherwise sufficient: it is derived from actual handler behavior and mutation-verified to catch missing, stray, and required `bank` properties. After this small update, fresh CI/CodeRabbit, and a current-head recheck, this should be ready to approve.

AxDSan (approved):
Approving. The thing I most wanted to check was whether this advertises a parameter the handlers ignore, which is exactly the #728 failure mode. It does not.

## Verified end to end, not just read

Two banks through the real MCP dispatch, isolated data dir:

```
bankA sees zebra: True
bankB sees zebra: False
banks on disk:  bankA  bankB
stats bankA: {'stats': {'total_memories': 1, ...
```

Separate SQLite files, no cross-read, and `mnemosyne_stats` honours the argument. So all 24 advertised `bank` parameters are backed by handlers that actually resolve one.

The stated exclusions hold as written:

- `mnemosyne_validate` keeps its own `bank` with `enum: ['private', 'surface']`, a different concept, untouched.
- All four `mnemosyne_shared_*` tools report `has bank: False`.

`tests/test_mcp_server.py` + `tests/test_tool_surface_parity.py`: **89 passed**.

## Why this is the right shape

`BANK_PROPERTY` as one shared constant is what makes the parity test meaningful. Twenty-three hand-written copies of the same description would have drifted within two releases, and the failure mode of drift here is silent: a client reads one tool's description, assumes the same precedence everywhere, and writes into the wrong store. Documenting the precedence (explicit argument, then `MNEMOSYNE_MCP_BANK`, then `default`) in the shared constant means it is stated once and cannot disagree with itself.

The backward-compatibility framing is right too. Omitting `bank` keeps the previous behavior exactly, so no existing client changes meaning, and the description says so rather than leaving callers to infer it.

Worth noting for whoever tracks the beta: this closes a real usability gap for per-project banks. It came up in a user question about workspace-level recall this week, where the honest answer was "banks do what you want, but the parameter is not discoverable in every tool that accepts it." After this, it is.

dplush:
Superseded by #939 and #941, which landed the bank-aware MCP schema and validation semantics on main. This branch now conflicts with main and would reintroduce older tool boundaries. Closing as superseded; any standalone docs-generator improvements can be recovered separately if still useful.

Issue #342, opened 2026-06-17
[BUG] _maybe_auto_sleep() times out using sleep_all_sessions + daemon thread SQLite race


ruangraung:
## Description

Two related bugs in `hermes_memory_provider/__init__.py`:

1. **`_maybe_auto_sleep()` uses `sleep_all_sessions()` which times out on large working sets.** With 1,000+ memories across multiple sessions, processing all sessions always exceeds the timeout. Auto-sleep fires every 10 turns but never completes.

2. **Daemon threads share SQLite connection with main thread.** Both `_maybe_auto_sleep()` and `on_session_end()` run `beam.sleep()` in daemon threads that reuse `self._beam.conn`. When the main thread continues `sync_turn()` (writing new memories), concurrent writes on the same connection cause episodic INSERT to fail silently.

## Environment

- **OS:** Debian Trixie (Linux 6.12.90, x86_64)
- **Python version:** 3.11.15
- **Mnemosyne version/commit:** mnemosyne-memory v3.7.0, mnemosyne-hermes v0.1.8 (local patches applied)
- **Installation method:** pip install (via Hermes venv)
- **Hermes version:** v0.16.0 (2026.6.5)

## Reproduction Steps

**Bug 1 (timeout):**

1. Set `MNEMOSYNE_AUTO_SLEEP_ENABLED=1` and `MNEMOSYNE_AUTO_SLEEP_TIMEOUT=300` in `.env`
2. Accumulate 1,000+ working memories across multiple sessions
3. Wait for auto-sleep to fire (every 10 turns)
4. Check logs: auto-sleep always times out

```bash
grep "auto-sleep" ~/.hermes/logs/agent.log
# Expected: "Mnemosyne auto-sleep completed within Xs"
# Actual: "Mnemosyne auto-sleep timed out after 300s — consolidation deferred"
```

**Bug 2 (SQLite race):**

1. Enable auto-sleep with a session that has ongoing writes
2. Check episodic memory after auto-sleep runs
3. Some INSERT operations fail silently

```bash
# Check logs for errors during auto-sleep
grep "episodic INSERT" ~/.hermes/logs/agent.log
```

## Expected Behavior

1. Auto-sleep should complete within the timeout and consolidate working memories into episodic summaries
2. Daemon threads should not interfere with main thread SQLite writes

## Actual Behavior

1. Auto-sleep always times out because `sleep_all_sessions()` processes ALL sessions instead of just the current one
[...]

Issue #687, opened 2026-08-10
[BUG] Hermes auto-sleep can consolidate unrelated sessions in a shared DB


dplush:
## Summary

Hermes-provider auto-sleep is initiated from the active provider/session lifecycle, but its threshold and eligibility checks currently inspect the shared working-memory store, and its worker then calls `sleep_all_sessions()` when available. In a shared multi-session database, one auto-sleep invocation can therefore consolidate eligible working memories from other sessions.

## Evidence

This behavior is present on current `main`, whose standalone provider declares version `0.6.0` (not currently released as a `v0.6.0` GitHub release), and in the released `v0.5.0` tag.

The current path captures the active session identity and creates an isolated worker with it, but threshold and eligibility use unfiltered shared-store reads, then the broad operation is selected solely by capability:

```python
sleep_all_sessions = hasattr(beam_ref, "sleep_all_sessions")
...
(
    sleep_beam.sleep_all_sessions
    if sleep_all_sessions
    else sleep_beam.sleep
)()
```

`BeamMemory.sleep_all_sessions()` enumerates eligible rows grouped across all sessions before invoking session-scoped `sleep()` for each. The worker connection avoids cross-thread SQLite use, but it does not constrain the operation to the triggering `session_id`.

A local dry-run did not perform any consolidation.

## Why this matters

Auto-sleep is a per-turn/provider lifecycle hook. In a shared Hermes database, it is surprising for an automatic trigger from one conversation to cause maintenance writes for unrelated conversations, especially when those conversations have not themselves crossed a session-scoped threshold.

## Requested decision

Is global consolidation from an automatic per-session lifecycle hook intentional?

If the intended contract is that automatic consolidation remains session-scoped, the narrow implementation would be:

- keep an isolated worker/SQLite connection;
- calculate threshold and eligibility for the triggering `session_id`;
- construct the worker with that session identity and context;
- call `sleep()` for auto-sleep;
- retain `sleep_all_sessions()` only for explicit/manual all-session maintenance.

## Acceptance contract

A synthetic two-session regression should prove, when session-scoped auto-sleep is the intended contract:

1. A provider lifecycle turn in Session A initiates auto-sleep while the shared DB has eligible old working memories for both A and B.
2. The asynchronous worker completes deterministically in the test.
3. Triggering auto-sleep from A consolidates only A.
4. B's working rows and consolidation state remain unchanged.
5. SQLite integrity remains `ok`.

No production memory content or identifiers are needed for the test.

[...]

PR #772, opened 2026-08-17
fix(hermes): scope auto-sleep to triggering session; honor auto_sleep_enabled (#771)

### LINKED ISSUE #771: [BUG] A single MCP write consolidates a replica's entire `mcp_{bank}` backlog — 407 → 147 working rows, and `auto_sleep_enabled: false` does not prevent it
Related: #687 (auto-sleep crosses sessions in a shared DB), #653 (MCP session id hardcoded to `mcp_{bank}`).
This report adds a case those two do not cover: a **replicated shared-surface DB**, where the damage is
immediate and large rather than gradual.

## Summary

On a machine that holds a *replica* of a shared-surface database, the first write through the MCP server
consolidates every `working_memory` row belonging to session `mcp_default` in one sweep. In our case that
destroyed 261 of 407 rows (64%) on a single `mnemosyne_remember` call.

Setting `auto_sleep_enabled: false` does **not** prevent it, even though `mnemosyne config get` confirms the
setting is loaded.

## Environment

- `mnemosyne-memory` 3.14.0 (also reproduced against the 3.15.1 wheel's `mcp_server.py` transport list)
- Python 3.12, Linux aarch64 (Ubuntu), and a macOS 3.11 origin machine
- Streamable HTTP MCP server (`streamable_http_server.py` wrapper calling
  `mnemosyne.mcp_tools.get_tool_definitions` / `handle_tool_call`)
- Topology: macOS origin ⇄ Linux replica, `mnemosyne-sync-relay` (sync server) + `mnemosyne-surface-capture`
  on the replica, bidirectional sync over an SSH tunnel

## Reproduction

1. On machine A (origin), use the MCP server normally for some weeks. Rows accumulate with
   `session_id = "mcp_default"` (hardcoded per #653).
2. Replicate the shared-surface DB to machine B.
3. On machine B, start an MCP server against that replica and issue **one** `mnemosyne_remember` call.

Observed on B:

```
working_memory  407 -> 147     (261 rows removed)
gists             1 ->   2     (the sweep produced one gist)
episodic_memory   0 ->   0
```

Row-level diff of the 261 removed vs the 146 kept:

| | count | session_id | created range |
|---|---|---|---|
| removed | 261 | `mcp_default` (100%) | 2026-07-21 23:08 → 2026-07-22 17:17 |
| kept | 146 | `shared-hermes` (100%) | 2026-07-22 18:27 → 2026-08-15 21:13 |

All 407 rows were `scope=global` and owned (`sync_surface_id` non-NULL) before the write. The split is
exactly by `session_id`: everything under the MCP's own fixed session id is swept; rows written by the
sync relay under a different session id survive.

Reproduced 3×: twice against the live replica, once against an offline copy with a fresh server process.

## `auto_sleep_enabled: false` does not stop it

```
$ mnemosyne config get auto_sleep_enabled
auto_sleep_enabled = False

# same single write against a pristine copy of the same DB
rows before: 407
rows after:  147
gists:       2
```

This is consistent with #687's observation that the sleep path is selected by capability rather than by
configuration:

```python
sleep_all_sessions = hasattr(beam_ref, "sleep_all_sessions")
... (sleep_beam.sleep_all_sessions if sleep_all_sessions else sleep_beam.sleep)()
```

A `hasattr` check cannot be disabled from `config.yaml`, so operators have no supported way to opt out.

## Why a replica makes this severe

On the origin machine the same code is harmless: `mcp_default` rows are consolidated incrementally as they
age, so there is never a backlog. A replica receives those rows already aged but **not** consolidated, so it
carries consolidation debt the origin already paid down. The first MCP write on the replica then discharges
weeks of that debt at once.

The result is silent, lossy divergence between two surfaces that sync is otherwise keeping consistent — the
origin retains 407 discrete memories while the replica holds 147 plus a gist, and sync has no way to
reconcile that.

## Impact

- 64% of working memory on the replica collapsed into a gist on one tool call, with no warning or dry-run
- Divergence is invisible to the sync layer, which reports healthy
- There is no configuration switch to prevent it
- Any deployment that replicates a shared surface to a second host (the documented pattern in #543) is exposed

## Suggested directions

1. Constrain auto-sleep to the triggering `session_id` (this is #687's proposal; it would fix this case too).
2. Honour `auto_sleep_enabled: false` on the capability-selected path.
3. Let the MCP server's `session_id` be configurable (#653) so a replica's server cannot inherit the origin's
   session identity.
4. Consider guarding consolidation when a row's `sync_surface_id` marks it as replicated rather than locally
   authored.

Happy to supply the before/after row dumps or test any patch — the reproduction is scripted and runs against
an offline copy.

ddddddddwp:
Fixes #771: a single MCP write consolidated a replica's entire `mcp_{bank}` backlog (407 -> 147 working rows), and `auto_sleep_enabled: false` did not prevent it.

## Root cause

- `_maybe_auto_sleep()` in `integrations/hermes/src/mnemosyne_hermes/__init__.py` selected the sleep path by capability probing: `sleep_all_sessions = hasattr(beam_ref, "sleep_all_sessions")`. `BeamMemory` always implements that method, so auto-sleep always ran `sleep_all_sessions()`, sweeping every session in the database. This was a regression from #342 (which fixed auto-sleep to use session-scoped `beam.sleep()`) reintroduced by the #604 rebind refactor; the root provider copy (`hermes_memory_provider/__init__.py`) kept the correct session-scoped implementation and test contract.
- The Hermes integration's `_read_config_key()` only read the Hermes `memory.mnemosyne.auto_sleep` key and never bridged to the core `MnemosyneConfig`, so the documented `auto_sleep_enabled` key (set via `mnemosyne config set`) never reached the provider.

## Changes

- **fix**: `_maybe_auto_sleep()` always runs session-scoped `sleep()` on a worker beam bound to the triggering `session_id`; the `hasattr` capability probe and ternary selection are removed. Manual `mnemosyne_sleep(all_sessions=True)` keeps its explicit all-sessions sweep.
- **fix**: `_apply_provider_config()` reads `auto_sleep_enabled` as a fallback after `auto_sleep`; `_read_config_key()` now bridges to the core `MnemosyneConfig` (parity with the root provider), so `auto_sleep_enabled: false` actually disables auto-sleep.
- **test**: new regression tests assert the worker beam only runs `sleep()`, and that `auto_sleep_enabled: false` disables auto-sleep. The packaged-provider parity test now asserts the worker's `sleep()` contract.
- **docs**: CHANGELOG entry under Unreleased/Fixed.

## Summary

- Restricts Hermes automatic working-memory consolidation to the triggering `session_id` and database.
- Prevents automatic `sleep_all_sessions()` calls from consolidating unrelated sessions.
- Preserves explicit all-session behavior for `mnemosyne_sleep(all_sessions=True)`.
- Supports core `auto_sleep_enabled` through the `MnemosyneConfig` fallback path.
- Disables automatic sleep when `auto_sleep_enabled: false` is set without a Hermes override.

## Architectural impact

- Protects working-memory isolation during the consolidation pipeline.
- Reduces cross-session privacy exposure and replica divergence for shared-surface databases.
- Preserves local-first behavior by keeping automatic consolidation local to the triggering session.
- Improves Hermes and MCP integration safety without changing their public entities.
- Does not change episodic memory, BEAM behavior, retrieval strategies, veracity processing, sync protocols, CLI behavior, or benchmark methodology.
- Adds regression coverage for session-scoped sleep and disabled automatic sleep.
- Updates packaged-provider parity coverage and documents the focused fix in the changelog.

## Maintainability

- Removes capability-based branching from automatic sleep selection.
- Uses one deterministic worker path for session-scoped consolidation.
- Bridges Hermes configuration reads to the core `MnemosyneConfig`.
- Keeps explicit all-session behavior available for intentional manual operations.


ddddddddwp:
Fixed in `ff351c8`. The regression test now leaves the Hermes `config.yaml` empty of `auto_sleep_enabled` and mocks `mnemosyne.core.config.get_config()` to return `False` only for that key (None otherwise), so it pins the core `MnemosyneConfig` bridge in `_read_config_key()` itself and would fail if the fallback were removed.

Verified locally: `tests/test_mnemosyne_hermes_session_switch.py -k auto_sleep` passes (5/5), full session-switch file passes (35/35), ruff clean on the changed file.

dplush (changes requested):
Thanks — the direct `sleep_all_sessions()` sweep and the `auto_sleep_enabled` gap are real improvements. I found one remaining contract hole that keeps this unsafe to merge for #687/#771.

`_auto_sleep_snapshot_locked()` still calls `get_working_stats()` and `_count_unconsolidated_before()` without a session filter. A backlog in Session B can therefore cross the threshold / eligibility check, reserve Session A’s reflection budget, and trigger A’s worker even when A has no qualifying backlog. Also, session-bound `sleep()` still calls the currently global `degrade_episodic()` path, so an automatic A-trigger can mutate B-derived episodic state.

Please make the complete automatic path session-scoped: trigger counts, eligibility, and any downstream degradation. Add the #687 two-session SQLite regression: trigger A with qualifying rows in both sessions; prove only A changes, B working/episodic/derived state stays unchanged, B-only backlog does not trigger A, and `PRAGMA integrity_check` is `ok`. Keep explicit `all_sessions=True` as the sole all-session boundary.

The focused suite and full CI are green on `ff351c8`; this is a contract/data-integrity blocker, not a CI issue.

dplush:
Scope update to my earlier review: I recommend keeping this PR as the narrow immediate safety fix for #771. The change from capability-selected `sleep_all_sessions()` to the triggering session’s worker `sleep()`, plus honoring `auto_sleep_enabled`, removes the reported bulk working-memory sweep.

The remaining complete-auto-sleep contract is broader and should be tracked as a follow-up under #687: session-scoped threshold/eligibility and reflection budget, the global `degrade_episodic()` semantics, and the end-to-end two-session SQLite/integrity regression. Please do not expand this focused PR into that Core contract.

One small correction before merge: please narrow the CHANGELOG wording so it does not claim all auto-sleep work is session-scoped. It should specifically say that automatic *working-memory consolidation* no longer calls `sleep_all_sessions()` and instead uses the triggering session worker. Once that wording is adjusted, I will clear the earlier request-changes review.

ddddddddwp:
Thanks for clarifying the scope. I’ve updated the CHANGELOG in `ee54e77` to describe the focused fix: automatic working-memory consolidation now uses the triggering session’s worker `sleep()` instead of `sleep_all_sessions()`. The broader auto-sleep contract remains deferred to #687 as requested.

Could you please take another look and update your review when you have a chance?

dplush (approved):
Re-reviewed on `ee54e77`. The CHANGELOG now accurately scopes this to automatic working-memory consolidation, and keeps the broader auto-sleep lifecycle contract with #687.

The focused tests, full current-head CI, CLA, and current CodeRabbit check are green. Approving this narrow #771 fix.

Issue #682, opened 2026-08-10
plugin.yaml name/version stale at hermes-mnemosyne 0.4.0 (dist is 0.5.0) - double registration


icommstudios:
## Problem

The Hermes dashboard shows two plugin rows for a single installed package (mnemosyne-hermes 0.5.0), and the second row carries a stale name and version.

## Root cause - stale plugin.yaml

The package ships an in-package `plugin.yaml` manifest that was never updated when the package version moved:

`mnemosyne_hermes/plugin.yaml` (lines 1-2):
```yaml
name: hermes-mnemosyne          # old name; dist is "mnemosyne"
version: 0.4.0                  # stale; dist is 0.5.0
```

This file is discovered when Hermes scans the user plugins directory. The symlink at `~/.hermes/plugins/mnemosyne` points at the package directory, and Hermes reads the package's own `plugin.yaml` from there - revealing the stale name/version.

## Root cause - double registration

The same package registers via two discovery paths:

1. Pip entry point `[hermes_agent.plugins]` -> produces row: `mnemosyne / entrypoint / 0.5.0`
2. The installer-created symlink in the user plugins directory (reads `plugin.yaml`) -> produces row: `hermes-mnemosyne / user / 0.4.0`

Hermes's plugin discovery does not dedupe by installed dist, so one package lists twice.

## Suggested fix

Two options, not mutually exclusive:

1. (Minimal) Sync `plugin.yaml` name to `mnemosyne` and version to `0.5.0` to match the dist. This collapses the two rows to the same identity.

2. (Better) Address the double registration: the installer creates a plugins-dir symlink (`install.py:install_plugin()`, line 786) that duplicates the pip entry point group `[hermes_agent.plugins]` discovery. Ideally the installer detects the pip entry point is already discoverable and skips the symlink, or Hermes dedupes rows that resolve to the same installed dist.

## Important caveat (confirmed by Hermes core maintainers)

The plugins-dir symlink is NOT redundant - it is the actual load path for the memory provider. Hermes resolves memory providers via `plugins/memory/__init__.py find_provider_dir()` (line 124), which scans bundled `plugins/memory/<name>/` then user `$HERMES_HOME/plugins/<name>/`. The pip entry point group `hermes_agent.memory_providers` is NOT consumed by hermes-agent core. So any fix that removes the symlink would break provider resolution on a fresh session.

## Environment

- mnemosyne-hermes 0.5.0 (pip-installed in Hermes venv)
[...]

PR #793, opened 2026-08-18
fix(hermes-provider): share one initialized provider with plugin tools

supervised-nl:
register() built a second MnemosyneMemoryProvider for PluginManager
handlers. Hermes Desktop deferred those tools onto the empty instance,
so mnemosyne_remember/stats returned "not initialized" while the CLI
and MemoryManager used the live DB.

Reuse one module-level provider and lazy-initialize on first primary
tool call when Hermes never called initialize().

## Description

Hermes can load mnemosyne-hermes twice in one process: as the MemoryManager
provider and as a PluginManager plugin. `register()` constructed a **second**
`MnemosyneMemoryProvider` and bound every `mnemosyne_*` tool to that object.
`initialize()` only ran on the MemoryManager instance.

Result: Desktop / deferred `tool_call` returned
`{"status": "memory_unavailable", "reason": "Mnemosyne not initialized"}`
while `mnemosyne store`, `mnemosyne stats`, and `hermes memory status`
used the live SQLite DB. `_beam is None` and `_init_error is None` — init
was never attempted on the tool object.

This is not a skip-context miss and is present since at least plugin v0.5.0.

**Fix:** one module-level provider for both registration paths; lazy
`initialize()` on the first primary-context tool call if Hermes never
called it. Cron/subagent skip-contexts still skip.

## Related Issue

No issue number — this token cannot open issues on mnemosyne-oss/mnemosyne.
Related but different: #682 (duplicate plugin list rows), #102 (surfaced
the failure instead of a silent no-op).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [x] Hermes provider tests: `pytest integrations/hermes/tests/test_plugin_tool_initialization.py tests/test_persona_plugin_registration.py tests/test_canonical_tools.py`

New tests:
- `register()` constructs a single provider for MemoryManager and tools
- plugin `mnemosyne_remember` stores without a prior `initialize()`

`test_plugin_registration_honors_tool_allowlist_before_initialize` already
fails on unmodified `main` in this environment; not introduced here.

Core suite `pytest tests/ -v` was not run (change is hermes-provider only).

## Checklist

- [ ] Version bumped in `mnemosyne/__init__.py` (Hermes-provider-only; no core version bump)
- [ ] README updated if user-facing behavior changed (not needed)

## Summary

- Hermes now reuses one module-level `MnemosyneMemoryProvider` for `MemoryManager` and `PluginManager`.
- Primary-context tool calls lazily initialize the provider when Hermes has not called `initialize()`.
- Cron and subagent skip-contexts retain their existing behavior.
- Initialization failures remain visible to callers.

## Architectural impact

This is the right call for the Hermes integration. Shared provider state prevents inconsistent memory access and duplicate provider construction.

The change does not alter working, episodic, or BEAM memory tiers. It does not alter retrieval, consolidation, veracity, or synchronization. It does not change Mnemosyne’s local-first or privacy posture. The provider remains local, and the PR adds no remote data path.

The PR changes the Hermes integration only. It does not change MCP or CLI surfaces.

- Added tests for shared provider construction.
- Added a test for plugin memory storage without prior initialization.
- Updated the changelog.
- Replaced an outdated fallback-parser assertion with a test for the intended provider behavior.
- Hermes-specific tests passed.
- The full core test suite was not run.
- The CLA assistant confirmed that all committers signed the CLA.


AxDSan:
Heads up on what I just did to this PR, and why: I used the maintainer "update branch" action to merge current `main` into `fix/hermes-plugin-uninitialized-tools`. That adds a merge commit to your branch. Nothing of yours was rewritten or dropped.

The reason is that this PR's CI was not failing on anything you wrote. `tests/test_mcp_streamable_http.py` drove an authenticated SSE stream by hand and never delivered the initial `http.request` message that ASGI requires. mcp 2.0.0 answered anyway, so the bug was invisible; mcp 2.1.0 started reading the request body to enforce a size limit, so the handler correctly blocked and the test wedged. Every job then ran to the six hour ceiling and reported nothing. That is what your four red test jobs were.

The fix landed in #867. I first tried simply re-running your jobs, which was wrong of me and cost you another six hours of red: a re-run replays the *same* merge commit, and that commit predates the fix. Only a freshly computed merge picks it up, hence the branch update.

Your new run is going now and should behave normally. Sorry for the noise on your PR, none of it was yours.

AxDSan (commented):
Reviewed properly now that CI is trustworthy again. Sorry this sat: it was blocked on the harness, not on you. Good diagnosis, and the writeup made it fast to verify.

## What I checked rather than took on trust

- `register()` in `mnemosyne_hermes` did construct a second `MnemosyneMemoryProvider` and bind every tool to it, while `initialize()` only ever ran on the MemoryManager instance. Your description of the failure is exact.
- **Leaving `hermes_memory_provider/` untouched is correct, not an omission.** Its `register()` registers the CLI command and then delegates tool registration to `hermes_plugin`; it never constructs a second provider. So the bug genuinely lives in one of the two implementations. Worth stating out loud because the file layout invites the opposite assumption.
- The `"hermes_default"` fallback in `_ensure_initialized_for_tools` matches the provider's own default at `__init__.py:734`, so the lazy path cannot land tools on a different session than the rest of the provider uses. That was the thing I most wanted to be wrong about and it holds.

## One finding worth acting on: the parity assertion was deleted wider than the behavior change

The removed block in `tests/test_hermes_provider_parity.py` sat inside `for module in provider_modules.values():`, so it asserted the uninitialized `memory_unavailable` response for **both** implementations. Only `mnemosyne_hermes` changed. Deleting the whole assertion therefore also drops the only coverage of `hermes_memory_provider`'s uninitialized response, in a file whose docstring is "Parity checks for the two Hermes Mnemosyne provider implementations."

It had to stop asserting the old behavior for `mnemosyne_hermes`, so the deletion is the minimum that makes CI green. But the minimum here quietly converts a parity guarantee into an untested divergence, and the next person reading the file will not know the two providers now differ on purpose.

Suggestion: split it rather than drop it. Keep the existing assertion for `hermes_memory_provider`, and for `mnemosyne_hermes` assert the new behavior, that an uninitialized provider lazy-initializes on a primary-context tool call and does *not* return `memory_unavailable`. That turns a deletion into a regression test for the thing this PR fixes, and it documents the divergence where someone will actually see it.

## Two minor notes, neither blocking

**`globals().get("_provider")`.** The function already declares `global _provider`, so a plain `if _provider is None` reads the same value. If the indirection is working around something specific, a one-line comment would help; if not, simplify. Related: `assert _provider is not None` disappears under `python -O`, so it is type narrowing rather than a guarantee.

**No lock in `_get_or_create_provider()`.** The whole point is exactly one instance, and two concurrent `register()` calls could still produce two. Plugin registration is almost certainly single-threaded at load, so this is a note rather than a request. Mentioning it because the invariant is the feature.

## One thing to call out in the description

Moving `self._maybe_retry_init()` inside the `try` also changes where its exceptions land: they are now caught by the `except Exception` below and returned as a structured payload, instead of propagating out of `handle_tool_call`. I think that is an improvement and consistent with the C27 structured-response design right underneath it. It is just an error-surface change that the PR body does not mention, and it deserves a line so it is not discovered later as a surprise.

Nothing here blocks. The parity split is the one I would like to see before merge; the rest is your call.

supervised-nl:
@AxDSan Split the uninitialized tool-call assertion in tests/test_hermes_provider_parity.py as you asked:

- hermes_memory_provider still asserts memory_unavailable when uninitialized
- mnemosyne_hermes asserts lazy-init on a primary-context tool call (not memory_unavailable)

Also merged current main. CI is green.

EmberRoot:
Confirmed same root cause in production (Hermes v0.5.0, mnemosyne-hermes v0.5.0).

**Real-world trigger path:** Not just Hermes Desktop — the bg-review (background review) fork is the most common trigger in production. The fork is created with `skip_memory=True`, inherits the parent's `enabled_toolsets` (including "memory"), and routes `mnemosyne_*` tool calls through the PluginManager path — hitting the uninitialized module-level provider.

**Failure pattern (from production logs):**
```
bg-review fork → mnemosyne_remember → provider B handle_tool_call
→ _beam is None → "Mnemosyne not initialized"
```
All mnemosyne_* calls in bg-review forks fail with this error. The main agent's MemoryManager path works fine (provider A is initialized). This has been present since at least 8/29 in our deployment.

**More context:**
- The `register()` function (line 2756) creates a separate `_provider` that's never `initialize()`d
- `MemoryManager` path creates a separate `MnemosyneMemoryProvider` via `register_memory_provider()` — that one IS initialized
- Two instances, only one initialized
- The PR's fix (shared module-level provider + lazy init) is the right approach

Thanks for the fix — this has been producing noise in our logs for a while. Would be great to see it merged.

dplush (changes requested):
Thanks for fixing the duplicate normal-memory provider initialization. I found three lifecycle/surface-contract blockers on the current head:

1. `handle_tool_call()` discards caller lifecycle values before lazy initialization. A first PluginManager call from cron, a subagent, a non-default session/profile, etc. can initialize as `hermes_default` / `primary` / `cli` and write to the wrong session or DB. Please bind the active tool/host context atomically before lazy init, or fail closed when it is incomplete. Add uninitialized cron/subagent, non-default-session, and profile-path regressions.

2. `register()` builds the PluginManager tool map once, while provider allowlists can be re-resolved on a later `initialize(..., hermes_home=...)`. That can advertise an uncallable tool or omit an allowed one after a home/config change. Please bind registration and runtime dispatch to the same frozen/updated config source, with real `register()`-path coverage for omitted/null/empty/restricted/unknown allowlists and a home/reload transition.

3. The PR title promises shared lifecycle for plugin tools, but sync tools use a global `SyncAdapter(None)` and persona tools a separate `PersonaAdapter`; both bypass `_ensure_initialized_for_tools()`. Either attach these handlers to the context-bound shared provider and test their session isolation, or explicitly narrow the title/changelog/scope to normal memory tools.

The existing positive test proves a normal tool can write, but should also read/recall that exact ID through the same shared surface and cover the isolation cases above. Rebase is additionally required (`BEHIND`), but it will not resolve these contract gaps by itself.

PR #800, opened 2026-08-18
fix(release): the pre-push tag comparison must allow equality

AxDSan:
Regression I introduced in #794, found by pushing `v4.0.0b1` for real.

Replacing the `sort -V` comparison with a PEP 440 one, I required the pushed tag to be **strictly greater** than the latest existing tag in its namespace. The comparison it replaced deliberately allowed equality:

```sh
if [ "$(... sort -V | head -1)" = "$tag_version" ] && [ "$tag_version" != "$latest_ver" ]
```

`git tag --list` includes the tag being pushed, because it already exists locally by the time a pre-push hook runs. So `LATEST_VER` is routinely `TAG_VER` itself, and strictly-greater rejected **every first push of a new tag**:

```
ERROR: Tag '4.0.0b1' is not ahead of existing core (non-v0) tag 'v4.0.0b1'
```

#794's checks compared two distinct versions and never covered a tag against itself, which is the case that actually occurs in production. The ordering logic was right; the boundary was not.

Only a tag strictly behind its namespace is an error now. Re-verified the ordering table including the self-comparison, and confirmed with `git push --dry-run` of the real tag, which runs the hook without publishing:

```
✓ Tag 'v4.0.0b1' matches the release version ('4.0.0b1') in mnemosyne/__init__.py
 * [new tag]         v4.0.0b1 -> v4.0.0b1
```

## Summary

- Fixes pre-push tag comparison for first pushes of locally existing tags.
- Allows equality with the latest tag.
- Rejects only tags that are strictly older within the same namespace.
- Updates related error messages.

## Architectural impact

- No impact on working, episodic, or BEAM memory tiers.
- No impact on retrieval, consolidation, veracity, sync, or benchmark systems.
- No impact on Hermes, MCP, or CLI integration surfaces.
- No change to privacy posture or local-first guarantees.
- Improves maintainability by aligning hook behavior with Git’s local tag state.
- Dry-run validation confirmed that `v4.0.0b1` now passes the hook.


dplush:
Hi @AxDSan , #800’s original checks are green, but the branch is now behind main. Could you please rebase and force-push it? Once current-head CI and review rerun, we can merge the release-unblocking fix.

Issue #552, opened 2026-07-27
bug(query-cache): global invalidation does not reach already-open Enhanced Recall peers


dplush:
## Summary

Issue #550 fixes an invalidating `BeamMemory` instance and a fresh cache instance. A separate live-peer case remains: an already-open peer `BeamMemory` can retain an opaque Enhanced Recall result in its in-memory cache after another session invalidates a global memory.

## Reproduction on `c9af909`

1. Create two `BeamMemory` instances, sessions A and B, on the same DB.
2. A stores a `scope="global"` memory.
3. B warms an exact Enhanced Recall request containing that memory.
4. A successfully calls `invalidate(memory_id)`.
5. B repeats the identical request without being recreated.

Result: B still returns the invalidated ID from its live `_opaque` cache. The same DB has been invalidated, but B does not consult it on an in-memory opaque hit.

## Why separate from #550

#550 intentionally remains a narrow single-owner/fresh-instance fix. This case needs an explicit cross-instance cache-coherence contract, likely involving a shared invalidation generation or another process-safe mechanism. It should not be folded into the #550 hotfix.

## Decision needed

Define whether global-memory invalidation must be immediately visible to already-open peer instances, and select the coherence mechanism. Non-goal: per-request SQLite polling or an unrelated cache redesign without a measured contract.

dplush:
@AxDSan I reproduced #552 on current main with two already-open `BeamMemory` peers. The primary DB is correct after invalidation, but peer B still returns the global memory from its local opaque Enhanced Recall cache.

I recommend a durable `recall_cache_epoch` in the memory DB: bump it in the same transaction as a successful invalidating mutation, include it in the v2 Enhanced Recall cache key, and keep the existing cache delete as hygiene only. This gives already-open peers and separate local processes read-after-invalidate coherence without per-hit row validation or a process registry.

Proposed guarantee: after a successful invalidation returns, every subsequently started Enhanced Recall on an instance sharing that DB must not return the invalidated global memory from an old cache value. Concurrent recalls already in flight may complete against their earlier snapshot.

Would that be the intended coherence contract for #552?

PR #803, opened 2026-08-19
fix: preserve committed mutations when cache invalidation fails

### LINKED ISSUE #594: bug(query-cache): define post-commit invalidation failure contract
## Problem

Recall-visible mutations commit their database change and then call `_invalidate_query_cache()`. If the cache `DELETE` or commit fails, the mutation may already be durable while the public API raises an exception.

## Scope

Define and test one explicit failure contract across the existing post-commit cache invalidation paths (`forget_working()`, `invalidate()`, `supersede()` and siblings): a successful memory mutation must not be reported as failed merely because cache reconciliation fails. Choose a durable/reconcilable invalidation mechanism or an explicitly documented retry/error policy.

## Non-goals

Do not fold the concurrent stale-publication race into this issue. That belongs to #552.

Found while narrowing #593.

dplush:
## Summary

Fixes #594.

A cache invalidation failure after a mutation has committed must be warning-only: callers keep the mutation success result and the durable database state remains authoritative.

## Contract and scope

- **Boundary:** only direct post-commit cache calls in `BeamMemory.invalidate()` and `BeamMemory.forget_working()`.
- **Non-goal:** no cache-publication or concurrent stale-publication behavior changes (#552 remains out of scope).
- **Why this boundary:** these methods currently commit/delete first and then let `_invalidate_query_cache()` escape, producing a false failure after durable state changes.

## Reproducer

`uv run pytest tests/test_enhanced_recall_cache.py -k post_commit_mutations_survive_query_cache_invalidation_failure -q` currently fails on this draft head: the selected mutation raises `RuntimeError` after the row was committed/changed.

The next commit will make that post-commit invalidation warning-only and retain the existing normal invalidation coverage.

## Summary

- Preserves committed `invalidate()` and `forget_working()` mutations when enhanced-recall cache invalidation fails after commit.
- Treats post-commit cache failures as warning-only for transaction-owning operations.
- Retains cache errors for caller-owned and deferred transactions, so failures can still roll back uncommitted mutations.
- Keeps the durable database authoritative and records cache failures for later reconciliation.

## Architectural impact

- Improves the local-first guarantee by preserving durable memory mutations despite secondary cache failures.
- Does not change retrieval, consolidation, veracity, sync, Hermes, MCP, or CLI integration surfaces.
- Does not change privacy behavior or introduce external services.
- Does not address cache publication or concurrent stale-publication races.

## Maintainability

- Adds regression coverage for post-commit, caller-owned, and deferred transaction paths.
- Documents the cache invalidation failure contract in `CHANGELOG.md`.
- Separating transaction-owned invalidation from nested invalidation is the right call. It makes commit semantics explicit and prevents cache errors from falsely reporting successful durable mutations as failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
+- **Committed memory invalidations no longer report failure when enhanced-recall cache eviction fails (#594).** The mutation remains successful and the cache error is logged for reconciliation.
 - **Hermes providers no longer clear the shared host LLM backend while another primary provider remains active (#551).**
 - **The OpenAI-compatible modality retry test is deterministic under load (#798).** Its localhost stub handles one request at a time and records response statuses, so the 401/no-retry contract is checked against the response actually served.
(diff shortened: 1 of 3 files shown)


dplush:
@coderabbitai review

PR #815, opened 2026-08-20
fix(recovery): harden backup and restore paths

bayhnf:
Split from bundled #788 per dplush feedback. Focused backup/restore hardening and regression matrix.

## Summary

This PR hardens Mnemosyne’s local backup and restore paths.

- Backups use configured paths, unique names, SQLite online backups, chunked SHA-256 checksums, atomic publication, and failure cleanup.
- Restore validates metadata and checksums, rejects WAL/SHM sidecars, and denies `SQLITE_ATTACH`.
- Restore stages files, preserves originals, verifies integrity, fsyncs changes, rolls back after post-replacement failure, and reacquires writer exclusion after inode replacement.
- Recovery handles malformed metadata, emergency-restore attempt counts, `rotate_backups(keep=0)`, concurrent deletion, and optional sqlite-vec support.
- The CLI requires an explicit restore path.
- Regression tests cover staging visibility, locking, cleanup, atomic replacement, integrity checks, extension handling, and environment-based paths.

## Architectural impact

- **Local-first guarantees:** Strengthened through fail-closed validation, atomic replacement, rollback, cleanup, and fsync handling.
- **Privacy posture:** Improved. Restore blocks database attachment and disables extension loading after sqlite-vec setup. No backup encryption was added.
- **Retrieval, consolidation, veracity, and sync:** No changes.
- **Agent integration:** No Hermes or MCP changes. The CLI now requires an explicit restore path.
- **Maintainability:** Improved through explicit recovery invariants and regression coverage.

The right call is to isolate recovery hardening and enforce fail-closed restore behavior. Reported verification includes 30 recovery tests, 40 related tests, clean Ruff checks, and a clean diff check.

### DIFF
--- a/mnemosyne/dr/recovery.py
+++ b/mnemosyne/dr/recovery.py
@@ -13,7 +13,7 @@ import shutil
 import sqlite3
 from datetime import datetime
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 
 def get_default_paths():
@@ -38,31 +38,77 @@ def get_default_paths():
     else:
         root = Path(os.environ.get("HERMES_HOME", Path.home() / ".hermes"))
         data_dir = root / "mnemosyne" / "data"
-    backup_dir = Path(os.environ.get("MNEMOSYNE_BACKUP_DIR", data_dir.parent / "backups"))
+    backup_dir = Path(
+        os.environ.get("MNEMOSYNE_BACKUP_DIR", data_dir.parent / "backups")
+    )
     db_path = data_dir / "mnemosyne.db"
     return data_dir, backup_dir, db_path
+
+
+def _allocate_unique_backup_path(backup_dir: Path, timestamp: str) -> Path:
+    """Atomically allocate a unique backup filename.
+
+    Uses ``O_CREAT | O_EXCL`` so two concurrent backups can never both select
+    and write the same name (a check-then-create sequence races). Retries with
+    a short random suffix until an exclusive create succeeds.
+    """
+    import secrets
+
+    suffix = ""
+    for _ in range(64):
+        name = f"mnemosyne_backup_{timestamp}{suffix}.db.gz"
+        candidate = backup_dir / name
+        try:
+            fd = os.open(str(candidate), os.O_WRONLY | os.O_CREAT | os.O_EXCL)
+            os.close(fd)
+            return candidate
+        except FileExistsError:
+            suffix = "_" + secrets.token_hex(3)
+    raise RuntimeError(f"Could not allocate a unique backup filename in {backup_dir}")
+
+
+def _unique_staged_path(db_path: Path) -> Path:
+    """A staged-restore path unique per invocation, so concurrent restores to
+    the same target cannot clobber each other's staging file."""
+    import secrets
+
+    token = secrets.token_hex(4)
+    return db_path.with_name(f"{db_path.name}.{os.getpid()}.{token}.restore_staged")
+
+
+def _fsync_dir(path: Path) -> None:
+    """fsync a directory so rename/replace metadata reaches durable storage."""
+    try:
+        fd = os.open(str(path), os.O_RDONLY | os.O_DIRECTORY)
+    except (OSError, ValueError):
+        return
+    try:
+        os.fsync(fd)
+    except OSError:
+        pass
+    finally:
+        os.close(fd)
 
 
 def create_backup(db_path: Path = None, backup_dir: Path = None) -> Dict:
     """
     Create a compressed backup of the database.
-    
+
     Returns:
         Dict with backup_path, size, checksum, and timestamp
     """
     _, default_backup_dir, default_db = get_default_paths()
     db_path = db_path or default_db
     backup_dir = backup_dir or default_backup_dir
-    
+
     if not db_path.exists():
         raise FileNotFoundError(f"Database not found: {db_path}")
-    
+
     backup_dir.mkdir(parents=True, exist_ok=True)
-    
+
     timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
-    backup_name = f"mnemosyne_backup_{timestamp}.db.gz"
-    backup_path = backup_dir / backup_name
-    
+    backup_path = _allocate_unique_backup_path(backup_dir, timestamp)
+
     # Use sqlite3 online backup API instead of shutil.copyfileobj.
     # sqlite3.backup() is lock-aware (acquires read-lock), includes
     # uncommitted WAL frames, and is atomic — it won't produce a torn
@@ -75,14 +121,8 @@ def create_backup(db_path: Path = None, backup_dir: Path = None) -> Dict:
     # when copying vec0 virtual tables, AND dst.iterdump() (used to
     # serialize the in-memory backup to gzipped SQL) fails the same
     # way when introspecting the destination's vec0 schema.
-    # Mirrors the graceful-fallback pattern in core/beam.py.
-    def _load_sqlite_vec(conn):
-        try:
-            import sqlite_vec
-            conn.enable_load_extension(True)
-            sqlite_vec.load(conn)
-        except (ImportError, sqlite3.OperationalError):
-            pass  # optional extra; absence just means no vec0 tables
+    # Uses the module-level helper, which always re-disables extension
+    # loading afterward so no connection leaks an enabled state (I-1).
     _load_sqlite_vec(src)
     dst = sqlite3.connect(":memory:")
     _load_sqlite_vec(dst)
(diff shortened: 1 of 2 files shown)


dplush:
Thanks for the hardening work. I found two blocking issues that should be addressed before merge:

1. `restore_backup()` executes the decompressed backup with `executescript()` but does not install an SQLite authorizer. A crafted dump can use `ATTACH DATABASE` to write outside the target path. I reproduced this: the external database was created even though the restore later failed. Please deny at least `sqlite3.SQLITE_ATTACH` on the staging connection and add a regression test confirming that the target and filesystem remain unchanged.

2. `create_backup()` reserves the final filename before the write starts. If SQLite backup, gzip, checksum, or metadata writing fails, the zero-byte backup remains. It is then picked up by listing, rotation, health checks, and emergency restore. Please remove the reserved backup and any partial sidecar on every failure path.

I also noticed that `emergency_restore()` always reports `attempts: 1`, `list_backups()` aborts on malformed sidecar JSON, and `rotate_backups(keep=0)` deletes nothing. These should be corrected as part of the recovery-path hardening.

bayhnf:
**Review feedback addressed — head **\n\nAddressed the concrete recovery correctness and security findings from the latest review:\n\n- Backup reservation cleanup: failed backup/dump/gzip/metadata writes now remove the reserved archive and partial sidecar before re-raising.\n- Hashing: shared chunked  avoids whole-file reads for DB and compressed backup checksums.\n- sqlite-vec fallback: extension enabling is inside the guarded block;  and  are treated as optional capability failures, and disabling is defensive.\n- Restore security: staged dump execution now denies  through the SQLite authorizer.\n- Emergency restore reports the actual number of attempted backups.\n- Malformed backup metadata is skipped by  instead of aborting the entire listing.\n-  sorts backups before selecting the latest.\n-  validates non-negative , handles , and tolerates concurrent deletion.\n-  now closes the connection through a context manager.\n- CLI  without a path now exits non-zero with a clear error.\n\nVerification:\n- : **27 passed, 1 skipped**\n- E902 No such file or directory (os error 2)
--> mnemosyne/dr/recovery.py:1:1

E902 No such file or directory (os error 2)
--> tests/test_recovery_paths.py:1:1

Found 2 errors.: pass\n- : clean\n- Commit author/branch remains .\n\nThe remaining review items are primarily additional regression-test expansion and docstring coverage; no production-side change was made outside .

bayhnf:
**Review feedback addressed — head `bcfa60e`**

Addressed the concrete recovery correctness and security findings from the latest review:

- Backup reservation cleanup: failed backup/dump/gzip/metadata writes now remove the reserved archive and partial sidecar before re-raising.
- Hashing: shared chunked `_file_sha256` avoids whole-file reads for DB and compressed backup checksums.
- sqlite-vec fallback: extension enabling is inside the guarded block; `AttributeError` and `sqlite3.NotSupportedError` are treated as optional capability failures, and disabling is defensive.
- Restore security: staged dump execution now denies `sqlite3.SQLITE_ATTACH` through the SQLite authorizer.
- Emergency restore reports the actual number of attempted backups.
- Malformed backup metadata is skipped by `list_backups` instead of aborting the entire listing.
- `health_check` sorts backups before selecting the latest.
- `rotate_backups` validates non-negative `keep`, handles `keep=0`, and tolerates concurrent deletion.
- `verify_integrity` now closes the connection through a context manager.
- CLI `restore` without a path now exits non-zero with a clear error.

Verification:
- `tests/test_recovery_paths.py`: **27 passed, 1 skipped**
- `ruff check mnemosyne/dr/recovery.py tests/test_recovery_paths.py`: pass
- `git diff --check`: clean
- Commit author/branch remains `bayhnf/contrib/s1-recovery`.

The remaining review items are primarily additional regression-test expansion and docstring coverage; no production-side change was made outside `mnemosyne/dr/recovery.py`.

mnemosyne/dr/recovery.py
```diff
@@ -38,336 +38,662 @@ def get_default_paths():
...
+        # lock immediately after replace. A zero-window design needs an in-place
+        # restore or a separately governed sentinel lock.
+        if preserved_existed:
+            shutil.copy2(db_path, preserved_path)
+        os.replace(staged_path, db_path)
+        staged_path = None  # consumed by replace
+        # The old-inode lock no longer covers the path; release it and
+        # re-acquire on the replacement inode BEFORE any fsync or verify, so
+        # the vulnerable interval is just the release+acquire syscall pair.
+        lock_conn = _release_writer_lock(lock_conn)
```
dplush, line 485: **Blocking: writer exclusion is not continuous across replacement.** `_release_writer_lock()` closes the old-inode connection before `_reacquire_writer_lock()` opens the replacement inode. Between these calls, another Mnemosyne writer can open the new path, acquire `BEGIN IMMEDIATE`, commit, and then `restore_backup()` can verify and return success with that interleaved write included. The lock held before `os.replace()` cannot protect the new inode. Please use a recovery/sentinel lock that every writer honors across `os.replace()`, or restore in place while retaining the held lock, and add a deterministic test that coordinates a competing writer in this exact gap.

mnemosyne/dr/recovery.py
```diff
@@ -38,336 +38,662 @@ def get_default_paths():
...
+
     backup_dir.mkdir(parents=True, exist_ok=True)
-    
+
     timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
-    backup_name = f"mnemosyne_backup_{timestamp}.db.gz"
-    backup_path = backup_dir / backup_name
-    
-    # Use sqlite3 online backup API instead of shutil.copyfileobj.
+    backup_path = _allocate_unique_backup_path(backup_dir, timestamp)
```
dplush: **Blocking: the reserved backup path is still published to the rotation glob.** `_allocate_unique_backup_path()` creates the final `mnemosyne_backup_*.db.gz` name before the backup is complete. `rotate_backups()` can therefore unlink that path while `create_backup()` is writing it. The writer then continues through an unlinked inode, creates the sidecar afterward, and may return a backup path that no longer exists (or leave an orphan sidecar). Please write to invocation-unique staging names that do not match the rotation glob and atomically publish the completed payload and metadata with `os.replace()`, with cleanup limited to staging files. Add a coordinated backup/rotation race test.

dplush (changes requested):
## Review

I found two remaining data-integrity races that should be fixed before merge. The targeted recovery tests pass (27 passed, 1 skipped), and CI is green, but these windows are not covered by the current tests. See the inline comments.

dplush:
Thanks for addressing the first round of recovery findings. I reviewed the current head and found two remaining blocking races:

1. Writer exclusion is not continuous across `os.replace()`. The old-inode lock is released before the replacement inode is locked, allowing a concurrent writer to commit in that window.
2. `create_backup()` writes directly to the final rotation glob. `rotate_backups()` can delete the file while it is still being written, leaving a missing backup or orphaned sidecar.

Please address both with deterministic race tests before merging. The targeted recovery tests currently pass (`27 passed, 1 skipped`), but they do not cover these windows.

bayhnf:
**Review feedback addressed — head `65961b9`**

Addressed the two blocking data-integrity races highlighted by @dplush:

1. **Backup staging isolation from rotation glob:**
   - `_allocate_unique_backup_path()` now claims names via hidden `.reserve` markers and provisions private staging paths (`.{name}.staging-...`) that do not match `mnemosyne_backup_*.db.gz`.
   - In-progress backups remain completely invisible to concurrent `rotate_backups()` or other glob readers until payload and metadata are atomically published via `os.replace()`.
   - Failure cleanup is strictly limited to staging and reservation files; published backups are never touched.
   - Added coordinated test `test_backup_staging_files_not_visible_to_rotation_race`.

2. **Continuous writer exclusion across replacement handoff:**
   - Eliminated the unlocked gap during replacement: `_reacquire_writer_lock(db_path)` acquires `BEGIN IMMEDIATE` on the replacement inode *before* `_release_writer_lock(lock_conn)` releases the unlinked old-inode connection.
   - Any competing writer attempting `BEGIN IMMEDIATE` across the replacement transition remains blocked (`database is locked`).
   - Added deterministic race test `test_competing_writer_cannot_enter_during_inode_replacement_handoff`.

**Verification:**
- `tests/test_recovery_paths.py`: 30 passed
- `tests/test_schema_init_serialization.py` + `tests/test_hermes_doctor_bank_routing.py`: 40 passed
- `ruff check`: clean
- `git diff --check`: clean

bayhnf:
**Review feedback addressed — head `3fbb386`**

Addressed the feedback from CodeRabbit:
1. Wrapped `src` and `dst` connections in `create_backup` in deterministic `try ... finally: close()` blocks.
2. Wrapped `sqlite3.connect` in `verify_integrity` with `contextlib.closing` to explicitly close connection handles.
3. Catch `(OSError, json.JSONDecodeError)` in `list_backups` to skip unreadable/corrupted sidecar files.
4. Replaced the timing sleep in `test_competing_writer_cannot_enter_during_inode_replacement_handoff` with an explicit `threading.Event` handshake (`competitor_done`).

PR #825, opened 2026-08-22
test(beam): stop the valid_until fixture failing for two hours every day

AxDSan:
## Description

`test_stored_offset_bearing_valid_until_chronologically_filtered` stores a space-separated naive `valid_until` at `now + 2h` and asserts a lexical filter drops it, on the premise that a space separator sorts before a `T` separator.

That premise only holds while the **date components match**. Between 22:00 and 00:00 UTC the naive value rolls into tomorrow, sorts *after* aware-UTC now, the lexical filter keeps it, and `assert space_mid not in lexical` fails on a completely unchanged tree.

Two hours of red per day, on a test whose own control is the thing breaking.

Demonstrated directly:

```
run time (UTC)  now                          space valid_until      lexically included?
12:00           2026-08-21T12:00:00+00:00    2026-08-21 14:00:00    False
21:59           2026-08-21T21:59:00+00:00    2026-08-21 23:59:00    False
22:00           2026-08-21T22:00:00+00:00    2026-08-22 00:00:00    True   <-- fails
23:56           2026-08-21T23:56:00+00:00    2026-08-22 01:56:00    True   <-- fails
```

Found because two unrelated PRs went red on it with CI starting at 23:56 and 23:59 UTC. One of them (#824) changes no code at all, which is what ruled out both PRs as the cause.

The naive value is now clamped to the current UTC date so the trap the fixture sets actually holds. The final second of the UTC day is skipped explicitly, since no naive value can be both chronologically future and lexically earlier than now inside it.

**Behavior under test is unchanged.** This is a defect in the fixture, not in `valid_until` handling. The `#525` semantics it guards are untouched.

## Related Issue

No issue. `#525` follow-up, found during CI triage on #823 and #824.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [x] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

Swept the clamp across every minute of the day:

```
minutes where the sanity assertion would still fail: 0
```

| gate | result |
|---|---|
| the previously failing test | 1 passed |
| `tests/test_beam.py -k "valid_until or offset or naive or utc"` | 12 passed |
| `ruff==0.15.22` on the changed file | clean |

## Checklist

No version bump: the tree is already on `4.0.0b1` and this lands inside that pending release. No README change; nothing user-facing moves.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V2pLMWfZENtNZYZW4oVsxN)_

## Summary

This PR fixes a time-dependent BEAM test fixture failure near UTC midnight. The fixture keeps its naive `valid_until` timestamp on the current UTC date and requires a 30-second execution margin.

## Architecture impact

- **Memory architecture:** No changes to working, episodic, or BEAM memory.
- **Sync and privacy:** No changes. Local-first guarantees remain unchanged.
- **Agent integrations:** No changes to Hermes, MCP, or CLI surfaces.
- **Benchmark methodology:** No changes. The fixture now provides deterministic test behavior.

## Maintainability

The pure `_same_utc_day_future(now)` helper isolates UTC boundary logic from the behavior under test. Deterministic margin, boundary, near-midnight, and full-day tests improve fixture reliability.

This is the right call because it fixes the test defect without changing production behavior or public APIs.

## Validation

- `test_beam.py`: 106 passed, 13 skipped.
- Tests cover the 22:00 boundary, near-midnight behavior, offset handling, clamping, and all 1,440 minutes.
- Ruff checks pass.
- Changelog updated.


AxDSan:
@coderabbitai the Major finding is correct and I have fixed it. Pushed in `dcbd1fb`.

**Confirmed before fixing.** At 23:59:58.999 UTC the clamp yields 23:59:59.000, and the old guard only tested `_space_future <= _now`:

```
now          : 2026-08-21T23:59:58.999000+00:00
_space_future: 2026-08-21T23:59:59+00:00
skip fires?  : False
validity     : 0.001 seconds
```

Against a test that runs in roughly 0.9s, the row expires during the writes and assertions that follow. I fixed a two-hour flake and left a one-second one, which is not a fix.

**What changed.** The decision moved into `_same_utc_day_future(now)`, a pure function, which returns `None` when the clamped value would not outlive the caller. I used a **30-second** margin rather than the suggested 5: the margin has to exceed worst-case runtime on a loaded CI runner rather than local runtime, and the only cost of being generous is a slightly wider skip window at the very end of the UTC day.

**Your second point, taken in part.** Extracting the decision makes both branches testable without the real clock, which was the substance of it. `TestSameUtcDayFuture` pins the plain-offset case, the clamp, the exact 22:00 boundary, the millisecond case above, and sweeps all 1440 minutes asserting the property the fixture actually depends on: the stored value is chronologically future and lexically earlier than aware-UTC now.

**Declined, with reason.** The same comment asks for coverage of retrieval strategies, temporal queries, BEAM benchmarks, cross-session recall, MCP tools and Hermes provider parity. That is the `tests/**` path-instruction default and it does not fit this change, which corrects how one fixture builds a timestamp. None of those surfaces are touched, and adding assertions about them here would put unrelated coverage in a place nobody would look for it.

Gates: `test_beam.py` 106 passed, 13 skipped. `ruff 0.15.22` clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #828, opened 2026-08-22
[TRACKING][FEATURE] Deployment-informed memory lifecycle and integration improvements


bayhnf:
# Deployment-informed memory lifecycle and integration improvements

## Context and evidence standard

These proposals came from a Mnemosyne setup that we personally operate and have extended for a real Hermes integration. We exercised provider migration, write/recall policy, consolidation, recovery, identity scope, diagnostics, and agent lifecycle flows beyond isolated happy paths.

Private deployment observations are discovery and design evidence, not claims about every upstream deployment. Data below is sanitized. Proposed changes are local-first, opt-in unless stated otherwise, and include compatibility/rollback notes. Existing official issues are linked instead of duplicated.

This is a tracking/index issue, not one implementation request. We are willing to develop accepted items as small, dependency-ordered PRs. PRs should use `Addresses Exx in this tracker`; they must not close the tracker.

## Status vocabulary

- `PROPOSED_NEW` — no semantic equivalent found in 315 issues / 500 PRs.
- `EXISTING/PARTIAL` — official work covers part or all of the capability.
- `READY_FOR_GUIDANCE` — design and evidence exist; maintainer direction needed before code.
- `BLOCKED` — prerequisite contract or hardening is missing.
- `IMPLEMENTED_LOCALLY` — exercised in our private setup, not yet an upstream claim.

## Improvements index

| ID | Improvement | Status | Existing overlap | Detail |
|---|---|---|---|---|
| E01 | Instrumented provider migration: shadow capture, recall diff, gates, canary, soak, rollback | `PROPOSED_NEW` / `IMPLEMENTED_LOCALLY` | No framework-level equivalent found | [proposal](https://github.com/mnemosyne-oss/mnemosyne/issues/828#issuecomment-5378586315) |
| E02 | Safe recall metadata projection for policy-aware consumers | `PROPOSED_NEW` / `READY_FOR_GUIDANCE` | No read-side equivalent found; PR #644 is write-side | [proposal](https://github.com/mnemosyne-oss/mnemosyne/issues/828#issuecomment-5378586315) |
| E03 | Multi-agent reader-side visibility matrix | `EXISTING/PARTIAL` | #327, #653, #761 | [details](https://github.com/mnemosyne-oss/mnemosyne/issues/828#issuecomment-5378586528) |
[...]

PR #829, opened 2026-08-22
feat(recall): add scoped metadata projection

bayhnf:
Addresses E02 in #828.

## What changed

Recall results stay compact by default. Callers can now pass `metadata_keys=[...]` to `Mnemosyne.recall()` (and the module-level wrapper) to project selected top-level metadata after the normal scoped recall has completed.

The projection is deliberately allowlisted and bounded:

- no change to default recall output or ranking;
- scope and identity filters run before metadata projection;
- at most one metadata query per physical tier;
- synthetic fact rows are left untouched;
- maximum 500 results, 32 keys, and 16 KiB projected metadata per result;
- malformed stored metadata degrades to an empty projection;
- enhanced-recall cache entries remain compact.

## Why

The provider/policy integration needs project and provenance metadata after recall. The current result contract omits stored `metadata_json`, forcing consumers into per-result lookups. This adds a safe, explicit seam without exposing arbitrary metadata by default.

## Verification

Clean isolated environment (`TZ=UTC`, embeddings disabled, leaked `MNEMOSYNE*` and `NVIDIA_EMBEDDING*` variables unset):

- focused projection tests: **14 passed**;
- adjacent recall/scoring/cache/package tests: **233 passed, 1 skipped**;
- combined focused + adjacent: **247 passed, 1 skipped**;
- trial: 40 projected results, one metadata query, zero scope leaks, zero unrequested-key leaks, 41 ms;
- Ruff and `git diff --check`: clean.

The full core suite was **3292 passed, 36 failed, 15 skipped**. The failures are in pre-existing repair, optional sync-crypto, MCP/package surfaces outside this diff; the changed projection tests and adjacent recall suite are green.

No live Mnemosyne instance was used.

## Summary

- Adds opt-in `metadata_keys` projection to `Mnemosyne.recall()` and the module-level `recall()`.
- Preserves compact results, ranking, filtering, and synthetic fact rows by default.
- Hydrates metadata only after scope and identity filtering.
- Supports working and episodic tiers with one metadata query per physical tier.
- Limits projection to 500 results, 32 keys, and 16 KiB per result.
- Keeps unrequested metadata in local SQLite storage.
- Preserves compact enhanced-recall cache entries.

## Architecture and Privacy

The change adds a bounded projection step across working memory, episodic memory, and BEAM result handling. It does not change retrieval strategies, ranking, consolidation, veracity, or synchronization.

The allowlist and scope checks reduce metadata exposure. Malformed metadata produces an empty projection. Projection reads only local SQLite working and episodic rows after recall filtering. No remote service or synchronization path is added.

This preserves the local-first design. Hermes, MCP, and CLI integrations can use the existing recall APIs and request only required fields. No direct integration changes are required.

## Maintainability and Verification

The projection logic is isolated in `_hydrate_recall_metadata`. Explicit validation and size limits bound query and memory costs. The tests cover tier hydration, scope enforcement, malformed metadata, cache behavior, explain responses, and module-level recall. Documentation and the changelog define the API contract.

The export snapshot changes improve transaction consistency and completeness accounting. They are separate from metadata projection and add regression coverage.

Verification passed 247 focused and adjacent tests, with one skipped, plus documentation, Ruff, and diff checks. The full core suite retains pre-existing failures outside this change. Benchmark methodology is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Added
 
+- **Recall metadata can be projected explicitly without expanding default results.** `recall(..., metadata_keys=[...])` bulk-loads selected top-level metadata keys only for rows that passed that recall call's filters, with at most one query per physical memory tier. It is opt-in, skips synthetic rows, limits pages to 500 rows and projections to 32 keys / 16 KiB per result, and never exposes unrequested metadata.
 - **Multimodal memory: images, video and audio can become recallable memories (RFCs 0002, 0003, 0004).** `BeamMemory.remember_media(ref)` takes a reference to a piece of media, registers it, describes it through a configured modality provider, and writes the description back as an ordinary memory that hybrid recall already understands. Nothing about text recall changes.
 
   The stack is additive throughout. Two sidecar tables, `media_assets` and `media_moments`, are created `IF NOT EXISTS` by their own store when a bank is opened, so existing databases acquire them with no migration step and no change to any existing table. No new package dependency is introduced.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -137,7 +137,19 @@ results = mem.recall(
 )
 ```
 
-Returns a list of dicts with keys: `id`, `content`, `score`, `source`, `timestamp`, `importance`, `scope`, `metadata`.
+Returns compact result dicts with ranking, source, lifecycle and scope fields. Stored `metadata_json` is not included by default.
+
+Pass `metadata_keys` when a policy-aware consumer needs selected top-level metadata keys:
+
+```python
+results = mem.recall(
+    "deployment policy",
+    top_k=10,
+    metadata_keys=["project_id", "policy_version"],
+)
+```
+
+Projection happens after ranking and scope filtering, never changes recall counters, skips synthetic rows, and issues at most one query per physical memory tier. It accepts at most 500 results and 32 keys, with a 16 KiB projected-metadata cap per result. Unrequested metadata is never returned, and enhanced-recall caches remain compact.
 
 **Examples:**
 
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -133,7 +133,8 @@ results = mem.recall(
     temporal_halflife: float = None,      # Hours for decay (default: 168)
     vec_weight: float = None,             # Override vector scoring weight
     fts_weight: float = None,             # Override FTS scoring weight
-    importance_weight: float = None       # Override importance weight
+    importance_weight: float = None,      # Override importance weight
+    metadata_keys: list[str] = None       # Optional top-level metadata allowlist
 )
 ```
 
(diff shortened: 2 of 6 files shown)


bayhnf:
Addressed the documentation boundary note in `c2ecacf`: the signature now includes `metadata_keys`, and the docs state explicitly that projection reads only local SQLite working/episodic rows after recall filtering. Projected metadata is not added to enhanced-cache payloads.

`generate-docs --check`, `verify-docs`, Ruff, and `git diff --check` are clean.

dplush (changes requested):
The metadata projection itself is well bounded: it is opt-in, allowlisted, tier-qualified, keeps enhanced-cache payloads compact, and does not mutate returned cache objects.

I found two separate export/restore data-integrity blockers that need to be resolved before this can merge:

1. `export_to_file()` builds the payload through multiple independent read sequences, then calculates `meta["completeness"]` afterwards. A concurrent write can therefore make the payload and its completeness manifest describe different DB states. Please derive all exported surfaces and completeness from one consistent read transaction/snapshot, with a concurrency regression proving the manifest matches the exported state.

2. For an omitted column with an unknown SQLite expression default, completeness currently counts only source rows where that field is non-NULL. A source NULL can restore as a generated value such as `CURRENT_TIMESTAMP`, yet the export may claim `complete: true`. The unknown-default branch must conservatively count every row, with regression coverage for this NULL-to-default case.

The current Python 3.13 matrix is also red (`test_client_errors_are_not_retried[403]`), while the other test jobs were cancelled after fail-fast. That test is outside this diff, but the base was green, so please reproduce/classify it and provide a fresh complete matrix after the data-integrity fixes.

bayhnf:
CodeRabbit posted two outside-diff findings on `mnemosyne/core/memory.py` after the #822/#825 merge landed in this branch:

1. **Export completeness snapshot** (lines 1034–1037) — `export_to_file()` reads BEAM, legacy, annotation, canonical, and sync data in separate queries before `_export_completeness()` runs in a later sequence. A concurrent write between reads means the JSON payload and its `"complete"` flag can describe different states.
2. **Unknown restore-default affected-row count** (lines 157–166) — when `_sqlite_restore_default()` returns the unknown-default sentinel, the branch counts only non-NULL rows; an omitted column with a `CURRENT_TIMESTAMP` expression default can replace an original NULL but reports zero affected rows.

Both are real data-integrity concerns in the export/restore path. However, they were introduced by upstream PRs #822/#825 (not by this PR), and fixing them requires a consistent-read-snapshot redesign of `export_to_file()` plus a predicate change in the restore flow — both outside this PR’s metadata-projection scope.

I will track these as separate focused issues/PRs so they receive proper review rather than being bundled here.

bayhnf:
The retriggered CI run (`b995809`) confirms the failure is reproducible, not a one-off flake: `tests/test_modality_openai_compat.py::test_client_errors_are_not_retried[403]` fails on every Python version when the stub server’s second queued response (200) is consumed by the initial `configured()` fixture call instead of the test body. The fixture pre-populates two replies for retry testing but the client-errors parametrize only expects one request — the stub’s `(200, _OK_REPLY)` fallback leaks into the assertion.

This is a test-fixture bug in upstream `test_modality_openai_compat.py`, not caused by this PR. I will file a separate focused issue with the fix rather than bundling it here.

bayhnf:
Filed the modality test flake as a separate issue: #841. It is not caused by this PR — the same fixture pattern can affect any branch that touches `test_modality_openai_compat.py`.

dplush:
Hey @bayhnf the fix for #841/#844 just landed. Please rebase and hopefully the CI is running again.
👍

bayhnf:
Rebased onto current main with #844 included, and addressed both data-integrity blockers in a391d5a:

- portable export payload and completeness are now read through one shared SQLite connection and one read snapshot;
- export preserves caller-owned transactions;
- unknown SQLite expression defaults conservatively count every source row, including NULL rows that restore to generated values;
- regression coverage exercises a concurrent WAL writer, caller transaction ownership, and CURRENT_TIMESTAMP with a NULL source value.

Final CI run 32676230694 is fully green: Python 3.10, 3.11, 3.12, and 3.13, plus docs, lint, build, CodeRabbit, and CLA. The branch is CLEAN / MERGEABLE with no unresolved review threads.

@dplush please re-review when convenient.

PR #830, opened 2026-08-22
feat(mcp): multi-token SSE auth with per-agent identity (MNEMOSYNE_MCP_TOKENS)

### LINKED ISSUE #761: Feature request: multiple named auth tokens for the SSE MCP server (per-agent identity)
## Use case

We run one Mnemosyne SSE instance per host serving several local AI agents
(family assistant, admin assistant, CI jobs). Currently the SSE server accepts a
single MNEMOSYNE_MCP_TOKEN, so all agents share one credential:

- no per-agent audit distinction (who stored/validated what),
- rotating one agent's token rotates everyone's,
- revoking a single agent requires downtime for all.

## Request

Support multiple named tokens, e.g.:

    MNEMOSYNE_MCP_TOKENS='{"hermes-family": "tok1", "hermes-admin": "tok2"}'

with the token name recorded as author/agent identity on memories it creates
(or at minimum in the audit log). Backwards-compatible: single
MNEMOSYNE_MCP_TOKEN keeps working.

## Current workaround

We run one instance per agent on separate ports (8082, 8083, ...) with separate
SQLite banks — works, but multiplies memory/CPU for the embedding model on
CPU-only hosts and complicates backup/retention.

Thanks for the project — local-first memory with fastembed ONNX runs great on
CPU-only Ryzen boxes.

dwmb5055:
Implements #761.

## What

The SSE MCP server gains a JSON-object env var mapping **named bearer tokens**:

```bash
MNEMOSYNE_MCP_TOKENS='{"hermes-family": "tok1", "hermes-admin": "tok2"}' \
  mnemosyne mcp --transport sse --host 0.0.0.0 --port 8080
```

Each client sends its own token as usual (`Authorization: Bearer tok1`). The **name** of the matched token becomes the author identity on memories that client creates, giving per-agent audit attribution from a single instance — plus per-agent token rotation and revocation without downtime for the others.

## Why

Single shared `MNEMOSYNE_MCP_TOKEN` means: no per-agent audit distinction, rotating one agent's credential rotates everyone's, revoking one agent requires restart for all. The workaround (one instance per agent, separate ports/banks — see #761) multiplies memory/CPU for the embedding model on CPU-only hosts.

## How

- `_resolve_sse_auth` returns `{name: token}`; `MNEMOSYNE_MCP_TOKENS` takes precedence, legacy `MNEMOSYNE_MCP_TOKEN` is unchanged (maps to the single entry `"default"`) — fully backward compatible.
- Malformed JSON, empty names/secrets, or duplicates **refuse startup** with actionable errors — fail-closed, consistent with the S1 hardening policy.
- `_BearerTokenMiddleware` matches constant-time per candidate (existing pure-ASGI middleware preserved, SSE frames never buffered) and records the matched name in a contextvar (`mnemosyne/runtime_context.py`, avoids circular import) plus `scope["state"]`.
- `mcp_tools._create_instance` resolves author identity: explicit arg > `MNEMOSYNE_AUTHOR_ID` > token name — no client-side cooperation needed.

## Tests

- 10 new: env parsing (precedence, malformed/non-object/empty/duplicate all raise with the env var named), loopback skip, per-name accept, wrong-token 401, contextvar identity inside authenticated requests, scope-state propagation.
- 2 updated for the dict return type.
 - `test_s1_mcp_sse_auth.py` **71/71**, adjacent `test_multi_agent_identity.py`
   + `test_mcp_server.py` **96/96**, full suite **3394 passed / 36 skipped**. 
- Note: middleware tests use an instant-responding helper route instead of `GET /sse`, which hangs TestClient by design (endless stream).

## Docs

`docs/cli-reference.md` (multi-agent tokens section) + CHANGELOG entry.

Closes #761

## Summary

- Adds named multi-token authentication through `MNEMOSYNE_MCP_TOKENS` for MCP SSE and Streamable HTTP.
- Uses token names as authoritative memory identities in multi-token mode.
- Rejects conflicting `author_id` values before writes.
- Preserves legacy `MNEMOSYNE_MCP_TOKEN` behavior.
- Validates malformed, empty, non-string, blank, duplicate, and normalized-collision configurations.
- Binds identity to transport sessions and rejects credential changes.
- Does not change tiered memory, retrieval, consolidation, veracity, sync, benchmarks, or the core data model.

- MCP and Hermes agents can share one local Mnemosyne instance with distinct memory attribution.
- Agents can rotate or revoke credentials without separate instances, ports, or banks.
- Transport-native session ownership prevents identity switching. This is the right call.
- `ContextVar` separates request identity from MCP-dependent tool code.
- Named-token authentication applies on loopback and non-loopback hosts.
- CLI and Hermes documentation describe configuration precedence, validation, attribution, and TLS requirements.
- The change does not add tenant-bank isolation or alter shared-surface tool behavior.
- Existing working, episodic, and BEAM memory behavior remains unchanged.
- Existing sync behavior and benchmark methodology remain unchanged.

## Privacy and maintainability

- Per-agent identities improve audit attribution and reduce shared-credential exposure.
- Startup validation fails closed for invalid or ambiguous configuration.
- Bearer tokens remain visible to the transport layer. Non-loopback deployments require TLS.
- The change adds no cloud service or remote dependency. It preserves the local-first design.
- Existing remote embedding and LLM options remain unchanged and can still send configured content or queries to providers.
- Focused runtime-context code and transport tests limit coupling between authentication and memory tools.
- Regression coverage includes parsing, precedence, session ownership, attribution, disconnect cleanup, Streamable HTTP, and spoofing protection.


dwmb5055:
Addressed round-2 findings in 2e9c3be:                                       
                                                                                  
     - **Bounded session-identity map**: `session_identities` is capped at 10k    
   entries with FIFO eviction, so a client flooding unique `session_id`s cannot   
   grow memory unboundedly. The map is enforcement-only metadata — session        
   validity itself remains the transport's job. (I kept binding to first-sight    
   rather than only transport-established sessions because the middleware cannot  
   observe the transport's internal writer registry; the cap plus the transport's 
   own 404s bound the exposure. Happy to switch to a TTL if preferred.)           
     - **Empty/whitespace secrets**: parser tests now parameterize empty and      
   whitespace-only names AND secrets, guarding against `Authorization: Bearer `   
   matching a blank configured secret.                                            
                                                                                  
     Suite: 51/51 auth tests, 96/96 adjacent, ruff F,RUF022 clean, CI green on all
   four Python versions (b4345e9).

dwmb5055:
CI is also green on the latest commit 2e9c3be.

dplush:
Thanks for the update. I think the remaining session-lifecycle issue is still a merge blocker.

With the supported `mcp==2.0.0` SSE transport, `SseServerTransport.connect_sse()` creates the UUID internally and emits it only afterwards in the SSE `endpoint` event. The initial `GET /sse` therefore has no `session_id` for this middleware to bind.

That makes the first normal `POST /messages/?session_id=<uuid>` the first request that can create `session_identities[uuid]`. A client can open SSE with token A, then send that first message with token B. The middleware accepts and binds B before transport validation, while `server.run()` is still executing in the long-lived SSE task context established with A. A B-submitted tool request can therefore be attributed to A.

The 10k FIFO cap bounds memory, but it does not establish the claimed connection identity. It also lets arbitrary authenticated message IDs occupy or evict entries without a transport session and leaves stale entries after disconnect.

Please bind the token identity at the actual transport-session creation boundary, reject a mismatched token only for an active bound session, and remove the binding when that session closes. The regression should establish SSE with A, read the emitted endpoint UUID, send a real MCP message with B, and assert rejection with no tool execution or attribution side effect.

dwmb5055:
You're right — thank you for the precise analysis. 44a030b reworks the     
   binding to use the transport's native session ownership instead of my        
   parallel registry:                                                           
                                                                                
     - The middleware now sets `scope["user"]` to an `AuthenticatedUser` whose  
   `client_id` is the matched token name. `SseServerTransport.connect_sse()`    
   binds that principal into `_session_owners` at the actual session-creation   
   boundary (the GET /sse that created the session), `handle_post_message`      
   rejects a different credential exactly as if the session did not exist, and  
   `connect_sse()`'s finally block pops the binding on disconnect. The          
   first-sight/FIFO map is gone entirely — no parallel state to bound, evict, or
   expire, and stale entries after disconnect can't occur.                      
                                                                                
     - The regression follows your script: a real SSE session over uvicorn+httpx
   (a background reader keeps the stream alive so the session survives), the    
   emitted `endpoint` event yields the transport-generated UUID, then a real MCP
   `ping` is POSTed with token B — rejected with 404 and no session side effects
   — and with token A — accepted (202). A second test closes the stream and     
   asserts the binding is dropped (POST with the original token also 404s).     
                                                                                
     One practical note from writing that test: the SSE response must keep being
   read while POSTs are made; aborting the line iterator closes the httpx       
   response and tears the session down. That's the close-path working as        
   designed, but worth knowing for anyone writing similar tests.                
                                                                                
     Suite: 49/49 auth tests, 99/99 adjacent (identity/mcp_server/wheel), ruff  
   clean on changed files.

dwmb5055:
Round-4 findings addressed in 4d8eea2:                                     
                                                                                
     - **No silent degradation**: `AuthenticatedUser`/`AccessToken` are now     
   resolved once at app-build time; if the auth classes are unavailable the     
   server refuses to start instead of proceeding without the session-ownership  
   principal. The changelog's unconditional wording now matches unconditional   
   behavior.                                                                    
                                                                                
     - **Test hygiene**: `uvicorn` guarded with `importorskip` like `httpx`; the
   endpoint event is asserted (clear failure message instead of `KeyError:      
   'sid'`); leftover Polish comments unified to English.                        
                                                                                
     - **End-to-end author attribution**: new test opens a real SSE session with
   token A, completes `initialize` and a `tools/call` round-trip over the       
   stream, and resolves the author exactly as the tool handler does (contextvar 
   set by the middleware → `_create_instance`) — asserted to equal the token    
   name (`hermes-family`). Combined with the transport-level cross-token        
   rejection test, this covers both halves of the feature's claim: identity     
   enforcement and identity attribution.                                        
                                                                                
     Suite: 50/50 auth tests (incl. the E2E), 99/99 adjacent, 47 pass + 3 skip  
   in the CI-like environment (no uvicorn/httpx installed), ruff clean on       
   changed files.

dwmb5055:
Found and fixed the real cause of the intermittent CI failures — not timing after all:

Running the full suite exactly as CI does (`MNEMOSYNE_NO_EMBEDDINGS=1`, all files) reproduced it deterministically: an earlier suite shadows `sys.modules["logging"]` with a plain object, and uvicorn's startup then fails with `No module named 'logging.StreamHandler'` inside my SSE session tests only — which is why it never failed in isolation and moved between Python versions depending on suite ordering. This is a known upstream condition; `core/sync_server.py` already ships an identical recovery for it.

`37e6fe1` adds an autouse fixture that restores the real stdlib `logging` module around the SSE session tests (mirroring the sync_server workaround) and extends the close-path deadline. Full suite now passes as CI runs it: **3310 passed, 36 skipped, 0 failed** (previously 3 failed); isolated runs unchanged; ruff clean.

dplush (changes requested):
Thanks for the transport-lifecycle rework. The SSE A→B session-switch protection is now on the correct boundary: the transport binds the principal at session creation, rejects a mismatched POST, and removes the binding on disconnect.

One merge blocker remains in the feature’s core audit-attribution contract.

`mcp_tools._create_instance()` still gives a client-supplied `author_id` precedence over the authenticated token identity. On this branch, a client authenticated as token A can call `mnemosyne_remember` with `author_id="token-B"` and create a row attributed to B without possessing B’s token.

For authenticated multi-token SSE, please make the transport-bound token name authoritative, or reject a conflicting client-supplied `author_id`. If backward compatibility requires explicit `author_id`, retain it only for local/stdio or otherwise unauthenticated paths. Please add an end-to-end regression proving that token A cannot persist a memory as token B.

Two smaller follow-ups while touching this path:

- Accept the bearer auth scheme case-insensitively (`bearer` is valid HTTP syntax too).
- Add `httpx` to the relevant test dependency. The real SSE transport tests currently use `pytest.importorskip("httpx")`, so they are skipped in the declared dependency environment rather than necessarily exercised by CI.

This is pre-merge only; I found no impact on current `main`.

dwmb5055:
All three findings addressed:

**Merge blocker — token identity authoritative.** `mcp_tools._create_instance()` now resolves the author from the transport-bound token name first. When a client-supplied `author_id` conflicts with the authenticated identity, the call is rejected with a clear error (returned as a tool `is_error` result, no row written) instead of silently attributing to the spoofed identity. A matching `author_id` is accepted, and an omitted one resolves to the token name. Explicit `author_id` / `MNEMOSYNE_AUTHOR_ID` remain authoritative only on unauthenticated paths (stdio/local), as suggested.

New regression coverage, end to end over a real authenticated SSE session (initialize → tools/call): token A calls `mnemosyne_remember` with `author_id="token-B"` → the result is `isError: true` with the conflict message, and nothing is attributed to B. Unit tests additionally cover the matching/omitted/unauthenticated resolution matrix.

**Bearer scheme case-insensitivity.** The middleware now parses the scheme per RFC 9110 (`bearer`/`BEARER` authenticate; any other scheme, or an empty credential, is still 401). Covered by a middleware test driving all three spellings through the same auth path.

**httpx in test dependencies.** Added `httpx>=0.27` to both the `test` and `dev` extras in `pyproject.toml`, so the real-SSE regression tests are exercised in the declared dependency environment instead of silently skipping.

dplush (changes requested):
The native SSE session ownership fix is now real: the A→B transport-session swap is rejected, binding is removed after stream close, and conflicting authenticated tool authorship is rejected before a write. The focused transport suite also passes with declared httpx/uvicorn dependencies.

Two current-head blockers remain:

1. `MNEMOSYNE_MCP_TOKENS` coerces non-string JSON keys/values with `str(...).strip()`. For example, `{"a": 1, "b": null, "c": true}` becomes valid predictable credentials `"1"`, `"None"`, and `"True"` rather than failing startup. Require JSON strings for both token names and secrets before stripping; add cases for null, booleans, numbers, lists, and objects.

2. The documented identity contract contradicts the implementation. The docs say explicit `author_id` still wins and single-token deployments are unaffected, but a legacy single-token deployment is normalized to the bound `default` identity and a conflicting explicit author is rejected. Please make an intentional compatibility decision, document it accurately, and add regression coverage. Either preserve the prior single-token explicit-author behavior, or document the breaking authoritative-token policy and state that conflicting `author_id` is rejected.

This PR is also behind current main; after these fixes it needs a rebase and fresh full gates.

dplush:
For the remaining compatibility decision, please preserve the existing single-token contract.

- `MNEMOSYNE_MCP_TOKENS` is the opt-in multi-agent mode: its matched token name is authoritative, and a conflicting client `author_id` must be rejected before any write.
- The legacy single `MNEMOSYNE_MCP_TOKEN` continues to secure the SSE transport and bind session ownership, but must not introduce a new enforced `default` author identity. Explicit `author_id` and the existing environment fallbacks keep their prior behavior there.

This keeps the new per-token audit guarantee where operators explicitly opt in, without turning the existing single-token deployment into a breaking authorship change. Please add regressions for both modes and update the docs to state the distinction clearly.

Separately, reject non-string JSON names and secrets in `MNEMOSYNE_MCP_TOKENS` instead of coercing them with `str()`; malformed values must fail closed at startup. After these changes, please rebase onto current `main` and request a fresh review.

dwmb5055:
Both blockers addressed, and the branch is rebased onto current `main` (single clean commit `e3fb312` on top of `0001316`).

**Fail-closed token parsing.** `MNEMOSYNE_MCP_TOKENS` entries must now be JSON strings for both the name and the secret; non-string values (`1`, `null`, `true`, lists, objects) raise at startup with an actionable message instead of being coerced. Regression cases cover all five value shapes plus a numeric-name/numeric-value combination; valid string mappings still parse unchanged.

**Single-token contract preserved.** As requested in the comment: the legacy single `MNEMOSYNE_MCP_TOKEN` continues to authenticate and bind session ownership, but introduces **no** author identity — explicit `author_id` and the `MNEMOSYNE_AUTHOR_ID` fallback keep their prior precedence exactly. The authoritative-token policy (conflicting client `author_id` rejected before any write) is exclusively the opt-in `MNEMOSYNE_MCP_TOKENS` multi-agent contract. This distinction is now stated in `docs/cli-reference.md` and the changelog, and both modes have dedicated regressions (single-token: authenticated request resolves to no identity + explicit author preserved; multi-token: spoofed `author_id` rejected end-to-end over a real SSE session).

**Rebase.** The rebase also meant integrating with the new unified HTTP transport stack from `main`: the multi-token mode now works on both the SSE and Streamable HTTP transports through the shared bearer middleware (`token=` legacy / `tokens=` multi-agent), keeping the single-token code path byte-identical in behavior. Full suite on the rebased branch: 3391 passed, 36 skipped, 0 failed; ruff clean.

mnemosyne/mcp_server.py
```diff
@@ -161,6 +175,68 @@ def _resolve_transport_security(host: str):
...
+    Raises RuntimeError with an actionable message on malformed JSON,
+    non-object payloads, empty mappings, empty names/secrets, and
+    duplicate secrets (two names sharing one secret would make per-agent
+    attribution ambiguous, since authentication matches the first name
+    for a presented token).
+    """
+    import json as _json
+
+    try:
+        parsed = _json.loads(raw)
```
dplush: **[P1] Reject duplicate names before choosing a credential.** `json.loads()` has already collapsed exact duplicate JSON members here, and the later `name.strip()` permits a normalized collision such as `"agent"` plus `" agent "`; both configurations silently select the later secret for the same authenticated principal. That contradicts the documented fail-closed contract and can rotate which credential owns an agent identity without an error. Parse with `object_pairs_hook` and reject exact *and post-normalization* duplicate names before populating the mapping, with regressions for both forms.

tests/test_s1_mcp_sse_auth.py
```diff
@@ -322,3 +325,741 @@ def test_401_response_has_www_authenticate_header(self, monkeypatch):
...
+                    content=_json.dumps(payload),
+                )
+                assert r.status_code == 202
+
+            # initialize + initialized, potem tools/call remember
+            post({"jsonrpc": "2.0", "id": 1, "method": "initialize",
+                  "params": {"protocolVersion": "2024-11-05", "capabilities": {},
+                             "clientInfo": {"name": "t", "version": "0"}}})
+            post({"jsonrpc": "2.0", "method": "notifications/initialized"})
+            post({"jsonrpc": "2.0", "id": 2, "method": "tools/call",
```
dplush, line 816: **[P2] Make the claimed attribution regression actually verify an authenticated write.** This sends `rag_query`, accepts either the initialize or tool response, and later sets the ContextVar manually before constructing an instance. It never proves that an SSE `mnemosyne_remember` call persisted the token name as `author_id`. Exercise a successful `mnemosyne_remember` over this session and assert the persisted row's attribution.

tests/test_s1_mcp_sse_auth.py
```diff
@@ -322,3 +325,741 @@ def test_401_response_has_www_authenticate_header(self, monkeypatch):
...
+                break
+            time.sleep(0.05)
+        assert server.started, "uvicorn did not start within 30s"
+        port = server.servers[0].sockets[0].getsockname()[1]
+        return server, thread, f"http://127.0.0.1:{port}"
+
+    def _build(self, monkeypatch, tokens):
+        monkeypatch.setenv("MNEMOSYNE_MCP_TOKENS", _json.dumps(tokens))
+        from mnemosyne.mcp_server import _build_sse_app
+        return _build_sse_app(host="0.0.0.0")
```
dplush, line 621: **[P2] Add a public Streamable HTTP multi-token regression.** This PR also wires the multi-token middleware into `_build_streamable_http_app`, but the new end-to-end transport coverage only builds the SSE app. Cover authenticated Streamable HTTP initialization, cross-token session rejection, and token-bound write attribution so that advertised transport does not rely solely on implementation symmetry.

dplush (changes requested):
Independent Builder review and Sol security gate both found a fail-closed configuration gap in the multi-token auth boundary. Please fix that blocker and make the transport/attribution regressions exercise the documented public behavior before re-review.

dwmb5055:
All three findings addressed, plus the docs/CHANGELOG wording aligned.

**[P1] Duplicate token names now refuse startup, in both forms.**
`_parse_tokens_env` parses with an `object_pairs_hook` that flags exact
duplicate JSON members *before* `json.loads` collapses them to the last
value, and the normalization pass tracks stripped names, so `"agent"` and
`" agent "` collide post-strip and refuse startup too. Both errors name the
offending entries verbatim (raw spellings, not the normalized one, so the
operator can find them). Regressions pin all three shapes: exact duplicate
(`{"a": "s1", "a": "s2"}`), whitespace collision (`{"agent": "s1",
" agent ": "s2"}`), and both spellings appearing in the collision error
(`TestMultiTokenParserEdgeCases`).

**[P2] The attribution regression now proves a persisted, authenticated
write.** `test_author_attribution_end_to_end` was rewritten: it no longer
probes the contextvar manually. It drives a real `tools/call
mnemosyne_remember` (no author_id) over the authenticated SSE session with
`MNEMOSYNE_DATA_DIR` isolated to tmp, asserts the success result read back
over the same SSE stream, then opens the SQLite database and asserts the
persisted `working_memory` row's `author_id` equals the matched token name.

**[P2] Streamable HTTP is now covered end to end.** New
`TestStreamableHttpMultiToken` drives a real uvicorn server against
`_build_streamable_http_app(host="0.0.0.0", json_response=True)`: an
authenticated initialize, cross-token session rejection (token B presenting
session A gets 404 "Session not found" — the SDK's stateful session manager
compares the principal from `scope["user"]` against the session owner), the
owning token keeping full access, a second principal opening its own
session, and token-bound write attribution (`mnemosyne_remember` over
authenticated Streamable HTTP → persisted row attributed to the token name).
No product change was needed here: the middleware already sets
`scope["user"]` on both transports, and the stateful manager enforces
ownership from it — the gap was exactly the missing coverage, now closed.

Docs/CHANGELOG updated to state the duplicate-name forms precisely and that
session-ownership binding applies on both transports (SSE
`_session_owners`, Streamable HTTP stateful manager principal check).

Local verification (Python 3.14, `MNEMOSYNE_NO_EMBEDDINGS=1`):
`test_s1_mcp_sse_auth.py` 71/71 (was 43), adjacent
`test_multi_agent_identity.py` + `test_mcp_server.py` 96/96, full suite
**3394 passed, 36 skipped, 0 failed** in 97 s (the only non-pass was the
pre-existing `test_snapshot_growth`, which is xdist-unsafe — it counts
snapshots in the shared `~/.hermes/mnemosyne/stats` and races with a
parallel worker; it passes serially — and one env-specific TestClient hang
under Python 3.14 deselected, unchanged by this PR and green in CI's
3.10–3.13), ruff clean on the CI selects (E9,F63,F7,F82 + F,RUF022 on
changed files).

dplush:
Thanks for the update. I reviewed the current head (`85b9706c`) and the earlier SSE / Streamable HTTP attribution findings look addressed.

I still see a few merge blockers:

1. `MNEMOSYNE_MCP_TOKENS` must fail closed when it is present but empty or whitespace-only. The current `or ""` path treats that as unset and can silently fall back to the legacy token.

2. Please restore the unrelated CLI, export-backup, and Streamable HTTP host/origin documentation that was removed from `docs/cli-reference.md`, plus the corresponding CHANGELOG entries. The current non-loopback example should also require HTTPS via a reverse proxy or secure tunnel, and its Authorization header is malformed.

3. The spoofing regression test should prove the documented “rejected before any write” guarantee: use an isolated data directory and assert that `spoof probe` was not persisted after the rejected request.

The PR is also currently conflicting / dirty and has no current-head GitHub Actions run. Once these are addressed, rebased, and verified on the new head, I’m happy to re-review.

dwmb5055:
All three blockers addressed, and the branch is rebased onto current `main`
(`e84d147` on top of `183c0a9`) — the docs sections are main's own content
again, with the multi-token material layered on top.

**[1] Blank `MNEMOSYNE_MCP_TOKENS` fails closed.** `_resolve_multi_tokens`
now distinguishes unset (`None` → legacy single-token contract, unchanged)
from present-but-blank: an empty or whitespace-only value raises at startup
with an actionable message instead of silently falling back to
`MNEMOSYNE_MCP_TOKEN`. Regressions pin `""`, `"   "`, and `"\t"` (all
refused) plus the unset-still-falls-back control (`TestTokensEnvFailClosed`).

**[2] Docs and CHANGELOG restored from main.** `docs/cli-reference.md` and
`CHANGELOG.md` are main's versions with only the multi-token additions
layered in — verified by diffing against `main`: zero deleted lines, the
unrelated CLI/export-backup/Streamable HTTP host-origin material is back.
The multi-token section now also (a) states that bearer tokens are cleartext
headers and a non-loopback bind must terminate TLS (reverse proxy or secure
tunnel) in front of the server, and (b) shows the client header as a proper
`curl -H "Authorization: Bearer tok1" …` example. The CHANGELOG entry
documents the blank-value refusal too.

**[3] Spoofing regression proves "rejected before any write".**
`test_spoofed_author_id_rejected_end_to_end` now runs with
`MNEMOSYNE_DATA_DIR` isolated to tmp and, after asserting the error result,
opens the SQLite store and asserts no `working_memory` row containing the
spoofed payload exists (under any author; an absent database is treated as
the stronger proof).

Local verification (Python 3.14, `MNEMOSYNE_NO_EMBEDDINGS=1`):
`test_s1_mcp_sse_auth.py` 75/75, adjacent multi-agent/MCP suites 149/149
(one env-specific TestClient hang under 3.14 deselected, unchanged by this
PR and green in CI's 3.10–3.13), full suite 4088 passed / 54 skipped with
8 failures that reproduce identically on clean `origin/main` in this
environment (6× `test_beam_e3_additive_sleep`, 2× the hermes-provider pair
that needs the `integrations/hermes` extra CI installs) — unrelated to this
diff, ruff clean on the CI selects.

dplush:
Thanks for the substantial update. I re-reviewed the current head (`49509178`).

The transport attribution, cross-token rejection, spoofing no-write proof, current CI/CLA, and the restored documentation sections all look much better. I still see one core contract issue before approval:

`_resolve_http_auth()` returns through the loopback bypass before evaluating `MNEMOSYNE_MCP_TOKENS`. As a result, on the default loopback host:

- a present-but-empty `MNEMOSYNE_MCP_TOKENS` value still starts silently, and
- valid named-token configuration is ignored, so there is neither authentication nor per-agent identity.

That conflicts with the documented named-token opt-in / fail-closed contract. Please either validate the named-token configuration before the loopback bypass and add loopback regressions, or explicitly decide and fully document the intentionally reduced loopback semantics.

Please also align `docs/cli-reference.md`, CHANGELOG, `--help`, and the relevant docstrings with the final behavior. They currently disagree on whether named tokens satisfy the non-loopback authentication requirement and whether blank configuration is always a startup error.

Finally, the non-loopback activation/error logs should explicitly state that bearer tokens require TLS termination via a reverse proxy or secure tunnel.

The head is also behind current `main`, so please rebase after the behavior and documentation are settled. Then a fresh CI, CLA, and review pass will be needed.

dwmb5055:
**Core contract — named tokens now evaluated before the loopback bypass.**
`_resolve_http_auth()` calls `_resolve_multi_tokens()` first, on every host:

- a valid `MNEMOSYNE_MCP_TOKENS` mapping opts the server into multi-agent
  mode on **every** host — bearer auth with per-agent identity is enforced
  even on loopback (the builders install the multi-token middleware; the
  "loopback-only, no auth" log path is no longer reached in this mode);
- a present-but-blank value refuses startup **on every host, loopback
  included** — it can no longer silently start behind the bypass;
- unset still falls back to the legacy contract (open loopback;
  `MNEMOSYNE_MCP_TOKEN` required on non-loopback).

I chose the first of your two options (validate before the bypass) rather
than documenting reduced loopback semantics: the opt-in contract is now
host-independent, which is the least surprising behavior. Loopback
regressions: blank-on-loopback (parametrized), valid-tokens-on-loopback
(require_auth=True, token=None), unset+loopback legacy control, plus the
early-round test that asserted the old "loopback ignores multi-token"
behavior was inverted to assert the new contract
(`TestTokensEnvFailClosed`, `TestMultiTokenResolve`).

**Docs aligned to this final behavior** — `docs/cli-reference.md`,
CHANGELOG, the module docstring, `_resolve_http_auth`/`_run_sse`
docstrings, and the builders' activation logs now all state: named tokens
are evaluated on every host and satisfy the authentication requirement by
themselves; blank configuration is always a startup error; bearer tokens
are cleartext headers and a non-loopback bind requires TLS termination via
a reverse proxy or secure tunnel. The activation/error logs for both
transports and both modes (single and multi) now say so explicitly.

**Rebase:** the branch is rebased onto current `main` (single commit on
top of `39d2e45`); CI, CLA and your re-review will run on the fresh head.

Local verification (Python 3.14, `MNEMOSYNE_NO_EMBEDDINGS=1`):
`test_s1_mcp_sse_auth.py` 84/84, adjacent MCP suites 144/144 (1
env-specific TestClient hang under 3.14 deselected, unchanged by this PR
and green in CI), full suite 4092 passed / 55 skipped with
8 failures that reproduce identically on clean `origin/main` in this
environment (6× `test_beam_e3_additive_sleep`, 2× the hermes-provider pair
that needs the `integrations/hermes` extra CI installs) — unrelated to this
diff, ruff clean on the CI selects.

dplush (changes requested):
Thanks, this is close. One important documentation gap remains before approval:

`scripts/generate-docs.py` does not add `MNEMOSYNE_MCP_TOKENS` to `ENV_ONLY_DESCRIPTIONS`. As a result, generated `docs/api/configuration.mdx` still lists 46 environment-only variables and omits the new multi-token configuration entirely.

Please add a factual entry covering its JSON mapping format, precedence over `MNEMOSYNE_MCP_TOKEN`, loopback behavior, and startup validation, then regenerate the configuration reference.

The remaining `test`-extra suggestion is not a CI blocker. CI installs the `mcp` extra and the transport/authentication suite runs there.

we are on it...

dplush (approved):
Reviewed `471929b0`: the generated docs, auth and TLS contract, test isolation, and fresh transport suite all check out. CI, CLA, and CodeRabbit are green. The remaining dependency thread is non-blocking because CI installs the MCP extra that supplies Starlette and Uvicorn.

dwmb5055:
Heads-up after today's push, in case it helps the final look:

- **Rebased onto current `main`** (`03f15d1`) — same seven commits, new hashes only; the equivalent of the previously approved tip `471929b0` is `b41dedf`.
- **`bcc53ec`** addresses CodeRabbit's minor finding: the legacy single-token statement in `CHANGELOG.md` and `docs/cli-reference.md` now says the token authenticates requests but binds **no session-owning principal** — matching `_BearerTokenMiddleware`, which sets `scope["user"]` only in multi-token mode. Docs-only change, no code touched.
- CI is fully green on `bcc53ec` (tests 3.10–3.13, docs-check, lint, build, CLA); CodeRabbit review completed with no new findings.

dplush (approved):
Reviewed the full merge-base diff at bcc53ec0. Named-token parsing, transport-native session ownership, authoritative author attribution, legacy compatibility, and generated documentation are consistent. Focused MCP/auth tests pass against mcp 2.0.0 and 2.2.0; no blocking findings.

Issue #827, opened 2026-08-22
[TRACKING][BUG] Deployment-informed correctness and reliability findings


bayhnf:
# Deployment-informed correctness and reliability findings

## Context and evidence standard

These findings came from a Mnemosyne setup that we personally operate and have extended for a real Hermes integration. Running the system across real write, recall, consolidation, recovery, scope, provider, and lifecycle paths exposed the items below.

Private deployment observations are discovery signals, not upstream proof. New bug claims are independently reproduced against current upstream code with sanitized fixtures. Existing findings link to their official issue/PR instead of duplicating them. No private memory content, credentials, hostnames, IPs, absolute private paths, or live identifiers are included.

This is a tracking/index issue, not a request for one large patch. We are willing to implement accepted items as small focused PRs. A PR should use `Addresses Bxx in this tracker`; it must not close the tracker.

## Status vocabulary

- `VERIFIED_NEW` — reproduced on current upstream and no semantic duplicate found.
- `EXISTING` — already tracked upstream; this tracker adds deployment-informed evidence/context only.
- `FIXED/MERGED` — accepted upstream.
- `OPEN_PR` — a focused fix already exists.
- `NEEDS_REPRODUCTION` — discovery signal only, not a confirmed upstream bug.

## Findings index

| ID | Finding | Status | Existing issue / PR | Detail |
|---|---|---|---|---|
| B01 | Concurrent fresh-DB initialization races check-before-ALTER schema guards | `VERIFIED_NEW` | No semantic duplicate found | [evidence](https://github.com/mnemosyne-oss/mnemosyne/issues/827#issuecomment-5378533682) |
| B02 | CLI operational failures leaked backend details or wrong success/error contracts | `FIXED/MERGED` | PR #814 | [details](https://github.com/mnemosyne-oss/mnemosyne/issues/827#issuecomment-5378534117) |
| B03 | Backup/restore and JSON import can leave unsafe or partial recovery state | `EXISTING` / `OPEN_PR` | #640, #727, PR #815 | [details](https://github.com/mnemosyne-oss/mnemosyne/issues/827#issuecomment-5378534117) |
| B04 | Hygiene candidate failures need per-item transaction isolation | `OPEN_PR` | PR #816 | [details](https://github.com/mnemosyne-oss/mnemosyne/issues/827#issuecomment-5378534117) |
[...]

PR #839, opened 2026-08-23
fix(shmr): keep non-dense imports usable without NumPy

### LINKED ISSUE #834: bug(shmr): keep non-dense imports usable without NumPy
## Summary

`mnemosyne.core.shmr` imports NumPy unconditionally. A minimal installation without NumPy cannot import the module, even when dense SHMR functionality is not requested.

Tracked from #827, B16.

## Verified current boundary

A fresh import with NumPy blocked raises `ModuleNotFoundError` at module import time.

## Decision needed

Confirm the capability contract:

- Is dense SHMR optional in the minimal installation?
- If yes, which dense SHMR entry points should emit a clear capability error?
- Which non-dense paths must remain importable without NumPy?

## Intended narrow follow-up after decision

Use a guarded or lazy NumPy import. Keep dense APIs explicit about the missing capability. Do not attempt to reimplement dense SHMR without NumPy.

## Acceptance criteria

- Importing non-dense SHMR surfaces succeeds without NumPy.
- Dense-only API use fails clearly and deterministically when NumPy is absent.
- Existing dense behavior remains covered when NumPy is installed.

## Non-goals

- Making dense SHMR NumPy-free.
- Changing SHMR scoring semantics.

bayhnf:
Fixes #834.

## What changed

`mnemosyne.core.shmr` imported NumPy unconditionally, so a minimal installation without NumPy could not even import the module — even though PR #31 established the base-install contract and the issue body confirmed that dense SHMR remains optional.

This change:

- replaces the unconditional `import numpy as np` with a guarded import (`np = None` on failure);
- adds `from __future__ import annotations` so `np.ndarray` type hints stay unevaluated;
- introduces `ShmrDenseCapabilityUnavailable(RuntimeError)` with a stable message (`shmr_dense_unavailable: numpy is not installed`);
- guards all dense-only entry points (`_embed`, `_cosine_similarity`, `_compute_harmony_score`, `harmonize`, `recall_beliefs`) so they fail deterministically before any work or state mutation;
- leaves non-dense surfaces (`FACTS_SCHEMA_SQL`, `_init_schema`, `_call_llm`, `_extract_json_from_llm_output`, `get_resonance_log`) fully usable without NumPy.

No scoring-semantic changes. No dense fallback without NumPy.

## Verification

Clean isolated environment (`TZ=UTC`, embeddings disabled):

- focused no-NumPy/dense regression tests: **11 passed**
- existing SHMR tests (with NumPy): **all passed**, confirming unchanged behavior
- full core suite: **3287 passed, 36 failed, 15 skipped = 98.9% pass rate**
  - 36 failures are pre-existing baseline issues in repair fd-binding, optional sync-crypto deps, and packaging/MCP surfaces — none touch SHMR or this diff
- Ruff clean; `git diff --check` clean

Addresses B16 in #827.

## Summary

This PR makes NumPy optional for `mnemosyne.core.shmr`.

- Non-dense SHMR surfaces remain importable and usable without NumPy.
- Dense entry points fail before work or state mutation with `ShmrDenseCapabilityUnavailable`.
- Existing dense behavior and scoring semantics remain unchanged when NumPy is installed.
- No dense fallback implementation was added.

## Architectural impact

The change preserves the working, episodic, and BEAM memory tiers. It does not change retrieval, consolidation, veracity, synchronization, or benchmark methodology.

Hermes, MCP, and CLI surfaces receive no direct behavior changes. Local non-dense workflows can import SHMR without NumPy. Dense workflows receive the stable error `shmr_dense_unavailable: numpy is not installed`.

The guarded import removes an unnecessary runtime dependency. It adds no network access, remote service, or data-sharing path. It preserves Mnemosyne’s local-first design and privacy posture.

## Maintainability

`ShmrDenseCapabilityUnavailable` defines a clear contract for a missing dense capability. `_require_numpy()` centralizes the capability check. Postponed annotations preserve static typing without requiring NumPy at runtime.

The regression tests cover importability, non-dense operation, deterministic dense failures, no-mutation behavior, NumPy-enabled behavior, exact error text, and safe subprocess path handling.

This is the right call because it removes an unnecessary dependency without changing dense semantics or adding an unsupported fallback.

### DIFF
--- a/tests/test_shmr_numpy_optional.py
+++ b/tests/test_shmr_numpy_optional.py
@@ -90,6 +90,7 @@ def test_non_dense_surface_usable_without_numpy():
 
 
 def test_dense_entry_points_fail_deterministically_without_numpy(tmp_path):
+    db = str(tmp_path / "harmonize.db")
     result = _run_without_numpy(
         f"""
         from mnemosyne.core.shmr import (
@@ -102,7 +103,7 @@ def test_dense_entry_points_fail_deterministically_without_numpy(tmp_path):
         )
         import sqlite3
 
-        db = r"{tmp_path / "harmonize.db"}"
+        db = {db!r}
         beam_conn = sqlite3.connect(db)
 
         class _Beam:
@@ -138,12 +139,13 @@ def test_dense_entry_points_fail_deterministically_without_numpy(tmp_path):
 
 def test_no_state_mutation_before_capability_error(tmp_path):
     """The dense guard must run before schema creation or row writes."""
+    db = str(tmp_path / "state.db")
     result = _run_without_numpy(
         f"""
         from mnemosyne.core.shmr import recall_beliefs, ShmrDenseCapabilityUnavailable
         import sqlite3
 
-        db = r"{tmp_path / "state.db"}"
+        db = {db!r}
         beam_conn = sqlite3.connect(db)
         beam_conn.row_factory = sqlite3.Row
 
--- a/tests/test_shmr_numpy_optional.py
+++ b/tests/test_shmr_numpy_optional.py
@@ -89,9 +89,9 @@ def test_non_dense_surface_usable_without_numpy():
     assert result.returncode == 0, result.stdout + result.stderr
 
 
-def test_dense_entry_points_fail_deterministically_without_numpy():
+def test_dense_entry_points_fail_deterministically_without_numpy(tmp_path):
     result = _run_without_numpy(
-        """
+        f"""
         from mnemosyne.core.shmr import (
             _cosine_similarity,
             _embed,
@@ -100,12 +100,21 @@ def test_dense_entry_points_fail_deterministically_without_numpy():
             recall_beliefs,
             ShmrDenseCapabilityUnavailable,
         )
+        import sqlite3
+
+        db = r"{tmp_path / "harmonize.db"}"
+        beam_conn = sqlite3.connect(db)
+
+        class _Beam:
+            conn = beam_conn
+            session_id = "test"
 
         for call in (
             lambda: _embed("probe"),
             lambda: _cosine_similarity(None, None),
             lambda: _compute_harmony_score([], []),
-            lambda: recall_beliefs(None, "query"),
+            lambda: recall_beliefs(_Beam(), "query"),
+            lambda: harmonize(_Beam()),
         ):
             try:
                 call()
@@ -114,15 +123,14 @@ def test_dense_entry_points_fail_deterministically_without_numpy():
             else:
                 raise AssertionError("dense call unexpectedly succeeded")
 
-        class _FakeBeam:
-            session_id = "test"
-
-        try:
-            harmonize(_FakeBeam())
-        except ShmrDenseCapabilityUnavailable:
-            pass
-        else:
-            raise AssertionError("harmonize unexpectedly succeeded")
+        tables = [
+            row[0]
+            for row in beam_conn.execute(
+                "SELECT name FROM sqlite_master WHERE type='table'"
+            ).fetchall()
+        ]
+        assert "harmonic_beliefs" not in tables
+        assert "memory_resonance_log" not in tables
         """
     )
     assert result.returncode == 0, result.stdout + result.stderr
--- a/tests/test_shmr_numpy_optional.py
+++ b/tests/test_shmr_numpy_optional.py
@@ -157,6 +157,10 @@ def test_no_state_mutation_before_capability_error(tmp_path):
             recall_beliefs(_Beam(), "query")
         except ShmrDenseCapabilityUnavailable:
             pass
+        else:
+            raise AssertionError(
+                "recall_beliefs unexpectedly returned without raising"
+            )
 
         tables = [
             row[0]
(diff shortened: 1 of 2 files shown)


dplush (changes requested):
Thanks, this is a focused implementation and the runtime capability boundary is correct.

One small regression-test change is needed before approval. The supported contract says dense entry points fail before doing work or mutating state. `harmonize()` is currently called with a fake beam, so the test verifies only the exception type.

Please call `harmonize()` with the existing SQLite-backed beam, assert the exact stable message `shmr_dense_unavailable: numpy is not installed`, and confirm that neither `harmonic_beliefs` nor `memory_resonance_log` was created. This makes the no-mutation guarantee regression-proof for both public dense entry points.

I verified the current implementation already raises before `_init_schema()`, so this is test coverage only, not a production defect.

bayhnf:
Addressed the test-coverage gap in `9dfbf69`.

`harmonize()` is now called with a real SQLite-backed beam (not a fake), and the test asserts the exact stable message `shmr_dense_unavailable: numpy is not installed` plus that neither `harmonic_beliefs` nor `memory_resonance_log` was created — confirming the no-mutation guarantee for both public dense entry points (`recall_beliefs` and `harmonize`).

All 11 focused tests pass; Ruff and `git diff --check` clean.

tests/test_shmr_numpy_optional.py
```diff
@@ -0,0 +1,193 @@
...
+            _cosine_similarity,
+            _embed,
+            _compute_harmony_score,
+            harmonize,
+            recall_beliefs,
+            ShmrDenseCapabilityUnavailable,
+        )
+        import sqlite3
+
+        db = r"{tmp_path / "harmonize.db"}"
```
dplush: Please serialize this path as a Python literal, for example `{str(tmp_path / "harmonize.db")!r}`, rather than embedding it in `r"..."`. A double quote or newline in `tmp_path` makes the child script invalid before the SHMR capability boundary is tested. The same fix is needed for the generated `state.db` path below.

dplush (changes requested):
The original database-backed `harmonize()` regression is now correct: it asserts the stable capability message and verifies no schema tables are created.

One small test-correctness fix remains before approval. The subprocess snippets embed the temporary DB paths inside raw double-quoted Python literals. A path containing a double quote or newline makes the child script invalid before it can exercise the no-NumPy contract. Please use a Python-safe representation such as `{str(tmp_path / "harmonize.db")!r}` at both generated-path sites, then rerun the focused suite.

This is test portability only; I found no production SHMR defect in the current implementation.

bayhnf:
Addressed the path-serialization note in `b27d6b2`.

Both subprocess tests (`harmonize` and `recall_beliefs`) now serialize `tmp_path` via `{str(...)!r}` instead of embedding it in an `r"..."` literal, so double quotes or newlines in the temporary path no longer produce invalid child-script syntax before the capability boundary is tested.

tests/test_shmr_numpy_optional.py
```diff
@@ -0,0 +1,195 @@
...
+        beam_conn.row_factory = sqlite3.Row
+
+        class _Beam:
+            conn = beam_conn
+            session_id = "test"
+
+        try:
+            recall_beliefs(_Beam(), "query")
+        except ShmrDenseCapabilityUnavailable:
+            pass
```
dplush, line 159: Please add an `else: raise AssertionError("recall_beliefs unexpectedly succeeded without numpy")` after this `except`. Without it, an unexpected normal return can still continue to the table assertions, letting this focused capability regression pass for the wrong reason.

dplush (changes requested):
Current-head re-review: the NumPy capability boundary is correct, the database-backed `harmonize()` no-mutation proof is present, and the path serialization fix is correct. Focused no-NumPy tests pass locally; the full CI matrix is green.

One final test-only robustness change remains. In `test_no_state_mutation_before_capability_error`, add an `else: raise AssertionError(...)` after the `except ShmrDenseCapabilityUnavailable` block. Otherwise an unexpected normal return from `recall_beliefs()` can still reach the no-table assertions and make this specific regression pass for the wrong reason.

This is redundant with the broader dense-entry-point test and is not a production concern, but it makes the targeted regression self-contained. After that, please request a fresh approval on the new head.

bayhnf:
Addressed the test-robustness note in `dcf885a`.

`test_no_state_mutation_before_capability_error` now has an `else: raise AssertionError(...)` after the `except ShmrDenseCapabilityUnavailable` block, so an unexpected normal return from `recall_beliefs()` fails the test immediately instead of reaching the no-table assertions and passing for the wrong reason.

All 6 no-NumPy subprocess tests pass; Ruff clean.

bayhnf:
Re-requesting review on the new head `dcf885a`. The `else: raise AssertionError(...)` guard has been added to `test_no_state_mutation_before_capability_error` so an unexpected normal return from `recall_beliefs()` fails immediately. All 6 no-NumPy subprocess tests pass; Ruff clean.

bayhnf:
@dplush Fresh review requested on head `dcf885a`. The `else: raise AssertionError` guard is now in place so `test_no_state_mutation_before_capability_error` cannot pass for the wrong reason.

dplush (approved):
Approved on the current head. The final no-NumPy regression guard is present; the database-backed no-mutation and path-serialization coverage remains correct. Focused tests, the full CI matrix, and the independent current-head review are clean.

Issue #840, opened 2026-08-23
[BUG] memoria_kg fills with hardcoded regex-extraction rows while LLM-extracted "kg" triples are parsed away


Prediction6767:
## Description

Two defects conspire to leave the knowledge-graph layer broken in opposite ways. Both have been present since the v3.0.0/v3.1.0 era and remain in **3.15.1** (latest PyPI release).

**1. The real path is dropped.** `EXTRACTION_PROMPT_TEMPLATE` (mnemosyne/core/extraction.py) explicitly asks the LLM for a `"kg"` category of subject–predicate–object triples during fact extraction. However, `_parse_facts()` iterates only `('facts', 'instructions', 'preferences', 'timelines')` — every triple the model returns for `"kg"` is parsed away and never reaches any store. As a result the canonical temporal store (`TripleStore`, backed by the `triples` table) remains empty indefinitely even when extraction runs successfully with `extract=True`.

**2. The prototype path writes low-quality rows.** A synchronous regex classifier — its own comment calls it *"Phase 2: MEMORIA regex-based extraction (always-on, zero-LLM-cost)"*, docstring: *"Uses regex patterns matching the BEAM benchmark oracles"* (mnemosyne/core/beam.py) — runs inside every `remember()` call and inserts hardcoded rows into `memoria_kg`:

- `subject` is always the literal `'user'`; only three predicates exist as string literals (`negation`, `decision`, `requires`); confidence values are hardcoded constants (`0.75` / `0.65`); `message_idx` is always `0`.
- Turn text is sliced upstream (500-char user / 800-char assistant caps in the Hermes provider integration) before scanning, and captures end at raw Python slices (`obj[:80]`, no word-boundary guard), so objects frequently end **mid-word**, exactly where the input was cut.
- Any text following trigger verbs (`decided to|chose to|opted for|selected|picked|switching to`) qualifies as a "decision", so conversational filler gets stored as decisions.
- The negation handler strips the negation word and stores what *follows* it under predicate `negation` — storing an affirmative reading of the sentence under a negation label (same word-boundary defect family as #507 / #559).
- No UNIQUE constraint on `(session_id, subject, predicate, object)` and no application-level dedup, so identical rows accumulate.

## Environment

- **OS:** Windows 11
- **Python version:** CPython 3.11
- **Mnemosyne version/commit:** mnemosyne-memory **3.15.1** (latest PyPI), verified against the official sdist
[...]

PR #847, opened 2026-08-24
Gate the regex memoria_kg prototype writer behind MNEMOSYNE_REGEX_KG (#840, containment)

Prediction6767:
## Description

First half of the #840 fix, split as requested by @dplush: containment only.

The regex `memoria_kg` writer built for early BEAM benchmark oracles inserts hardcoded rows — `subject='user'`, predicate in {negation, decision, requires}, confidence in {0.75, 0.65} — whose objects end at raw Python slices (`obj[:80]`, `decision[:...]`), producing mid-word junk like `killswi`. These rows are consumed by structured recall (negation/entity-action specialists and the CR/MR router), so the junk reaches consumers; it is not just internal telemetry.

This change makes all three prototype branches opt-in via `MNEMOSYNE_REGEX_KG` (env-only flag, not a `config.yaml` key; default off via `_env_truthy`):

- One gate at the top of the branch cluster inside `BeamMemory.extract_and_store_facts`. Remember, dedup-update, batch and the Hindsight importer all funnel through this function, so a single flag covers every write path.
- Metric/date/version/sequence extraction still runs unconditionally.
- Historical rows are never deleted or reinterpreted.
- Empty-table recall behavior is covered by new tests (the specialists return their documented fallback dicts when the table is empty).

**Merge order:** land this PR (#847) before #848. #848 still includes this commit for review convenience; merging #848 first would empty this PR.

**memoria_kg consumers (AxDSan):** with the writer default-off, CR/MR specialists that still `SELECT … FROM memoria_kg` receive no *new* rows. We accept temporary dormancy for new data; historical rows stay. Repointing those readers at TripleStore is follow-up work, not part of this containment PR.

A follow-up PR (#848) wires the LLM-extracted `kg` triples into the TripleStore — kept separate per the review discussion on #840 so each can be evaluated on its own.

## Related Issue

Part of #840 (containment half)

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] Existing tests still pass (`pytest tests/ -v`) — extraction/parity/triples suites match pristine-main results
- [x] New tests added for any new functionality — gate default-off, opt-in restore, metric-branch untouched, empty-KG recall fallback (6 cases)
- [x] Manual verification — grep audit of every `memoria_kg` write site; all funnel through the gated function

## Checklist

*(version bump left to the release cut; no user-facing README change)*


AxDSan (commented):
The containment is right and the diagnosis is convincing: `obj[:80]` producing `killswi` is indefensible, and gating rather than deleting is the correct first move. `tests/test_kg_triple_wiring.py`: **6 passed**. Not requesting changes, but there is one consequence neither PR states, and I would like it stated before this merges.

## What goes quiet when the default flips

`memoria_kg` is not only written by those three branches, it is read by two structured-recall specialists:

- `beam.py:5910` — negation predicates (`SELECT subject, object, message_idx FROM memoria_kg ...`)
- `beam.py:5936` — entity-action pairs, predicates `requires` / `decision`

With `MNEMOSYNE_REGEX_KG` off by default, nothing writes those predicates any more, so on a fresh store both specialists return nothing for the rest of time. Existing rows are preserved, as the description says, so an existing store degrades gradually rather than instantly, which is arguably the more confusing failure: negation recall keeps working on old memories and silently stops covering new ones.

**#848 does not fill the gap.** It writes to the `triples` table via `TripleStore`, not to `memoria_kg`, so the two `SELECT ... FROM memoria_kg` specialists are not fed by the LLM path either. After both PRs land, the regex writer is off and the new triples go somewhere those two readers do not look.

That may well be the right end state: junk rows feeding a specialist is worse than no rows. But it is a capability being removed, and right now it is removed as a side effect rather than as a decision. I would like one of three things in the description before merge:

1. The specialists are also retired or repointed at `TripleStore`, in a follow-up that is named here, or
2. an explicit statement that negation/entity-action structured recall is accepted as dormant until then, or
3. `_store_kg_triples` additionally writes the `memoria_kg` shape those readers expect.

## The BEAM question

The description says these rows were built "for early BEAM benchmark oracles" and that they feed the CR/MR router. That makes this a change that can move benchmark numbers, and our only end-to-end BEAM result of record (65.2% at 100K on v3.0.0) predates it.

There is an open question about a re-run on the current tree reportedly scoring far lower, which nobody has attributed yet. I am not claiming a connection, and I would rather not guess: what I want is that this does not become another unattributed delta. A before/after CR/MR measurement with the flag on and off, even a small one, would settle whether disabling the writer costs anything measurable, and would be worth more than the argument we would otherwise have later.

## Small note

The gate reads the env var directly rather than going through the config resolver, so `mnemosyne config set` cannot control it and the documented `config.yaml > env > default` precedence does not apply. That is defensible for a prototype escape hatch that we expect to delete, and inconsistent with how every other toggle behaves. Worth a one-line comment saying which it is, so the next person does not "fix" it by wiring it into config and accidentally give it a longer life.

@dplush this is the split you asked for; the cross-PR gap above is the bit I would want your read on.

Prediction6767:
Updated after @AxDSan / @dplush feedback on the #840 split:

- Merged current `main`; `MNEMOSYNE_REGEX_KG` now goes through `_env_truthy` (env-only, default off).
- PR body no longer `# DRAFT:`; CodeRabbit auto-summary that attributed TripleStore wiring to this containment PR was removed.
- Explicit product answer on memoria_kg CR/MR consumers: temporary dormancy for *new* data after containment; historical rows untouched; TripleStore repoint is follow-up (not dual-write here).
- Merge order: this PR (#847) before #848.

AxDSan (changes requested):
Decision: containment is the right first step and the merge order (this, then #848, then #855) stands. One blocker.

The diff removes the concurrency block and every timeout-minutes from .github/workflows/ci.yml. Those are what stop a hung matrix job from burning six hours (see #871 and #898). I assume the branch was cut from an older main and the ci.yml hunk is an accident, but merging it as-is would strip the protection. Please drop every change to ci.yml and rebase over current main; the beam.py side conflicts now after today's merges.

The MNEMOSYNE_REGEX_KG gate itself is fine as an env-only flag for a benchmark-oracle writer.

Prediction6767:
Friendly ping — all review feedback on this PR is addressed and it's ready for re-review (rebased on current `main`, `MNEMOSYNE_REGEX_KG` goes through `_env_truthy` per the #840 split discussion).

Happy to make any further changes. If it's a matter of bandwidth rather than the code, a "not right now" is fine too — we're running this containment locally in the meantime, so upstream adoption is a nice-to-have for us, not a blocker.

Thanks!

dplush:
Thanks for the containment work. I agree that the env-only default-off gate is the right narrow first step.

I cannot approve this head yet:

1. The diff removes the CI concurrency guard and all job timeouts from `ci.yml`. Please drop those unrelated workflow changes completely.
2. The PR now conflicts with current `main`, including `CHANGELOG.md`. Please rebase and resolve that before the next review.
3. The metric regression assertion is currently only `after >= before`, so it can pass without any metric row being written. Please assert the expected metric record and value directly.

The temporary CR/MR dormancy for newly created data is understood as the explicit containment trade-off and belongs to the planned follow-up, not this PR.

After the rebase, a clean CI run and fresh review on the new head, I am happy to take another look. Thanks!

PR #848, opened 2026-08-24
Wire LLM-extracted kg triples into the TripleStore (#840, contract)

### LINKED ISSUE #840: [BUG] memoria_kg fills with hardcoded regex-extraction rows while LLM-extracted "kg" triples are parsed away

Prediction6767:
## Description

Second half of the #840 fix, split as requested by @dplush: the KG contract. Stacks on #847 (containment) — please review and **merge #847 first**.

The extraction prompt asks the model for a `kg` category of subject-predicate-object triples, but `_parse_facts()` only flattens facts/instructions/preferences/timelines, so every returned triple was dropped on the floor.

This change:

- Parses `kg` triples into typed SPO tuples (`_parse_kg_triples`) instead of dropping them.
- Validates before storing (`validate_kg_triples`): non-empty fields; hedge / conversational-filler rejection via `_KG_FILLER_PREFIXES` (hedges such as `maybe` / `i think`, plus conversational openings including `you said` and `let me` / `let me check`, matched as stem with trailing-space-or-end); word-boundary-safe object truncation; lowercase snake_case predicates; within-batch dedup.
- Writes validated triples to the bank's TripleStore with `valid_from` set to the extraction date and provenance `source='llm_extraction'`; supersession keeps one current truth per subject+predicate.
- Adds an end-to-end test: mocked LLM JSON → `remember(extract=True)` → readable back through both `TripleStore.query()` and the real triple-query handler.

**Merge order:** #847 lands on `main` first, then this PR. This branch currently includes the containment commit for review convenience; merging this first would empty #847. (Upstream has no `contain-memoria-kg` base branch, so the stack is documented rather than GitHub-enforced.)

**memoria_kg consumers (AxDSan):** LLM triples go to `TripleStore`, while CR/MR specialists still `SELECT … FROM memoria_kg`. With the regex writer default-off (#847), those specialists receive no *new* rows. We accept temporary dormancy for new data; historical rows stay. Repointing those readers at TripleStore is follow-up work outside these PRs.

**Seam note:** upstream's parity suite patches `extract_facts_safe`, so the beam fact path keeps calling it unchanged; the KG path goes through a new best-effort `extract_triples_for_beam()` accessor. A small content-keyed LRU shares one extraction pass between the two, so `remember()` still makes a single LLM call.

Also ports this branch onto current main, adopting the malformed-reasoning-trace rejection guards at all four LLM call sites (tuple-shaped returns).

## Related Issue

Closes #840 (together with #847)

## Type of Change

- [x] Bug fix
- [x] Test improvement

## How Has This Been Tested?

- [x] Existing tests still pass — extraction/parity/triples suites match pristine-main results (only pre-existing Windows temp-cleanup failures remain, identical on vanilla main)
- [x] New tests added — parser/validator cases plus the end-to-end JSON→query-surface test (16 total in `test_kg_triple_wiring.py`)
- [x] Manual verification — labeled wheel installed locally; live store reads triples back through the query surface; filler cases for `you said` / `let me check` / bare `maybe` now reject as claimed

## Checklist

## Summary

- Wires LLM-extracted `kg` triples into the temporal `TripleStore`.
- Validates, normalizes, deduplicates, and supersedes triples by subject and predicate.
- Stores extraction dates in `valid_from` and uses `source='llm_extraction'`.
- Gates the legacy regex `memoria_kg` writer behind `MNEMOSYNE_REGEX_KG`.

## Architectural impact

This fixes the consolidation path used by `remember(extract=True)`. Validated triples now reach temporal storage instead of being discarded. The existing fact-extraction API remains compatible. New cached accessors expose triple results without duplicating LLM extraction.

Working memory, episodic memory, BEAM storage, retrieval strategies, veracity logic, and sync behavior remain unchanged. Triple queries now retrieve persisted LLM triples through the existing query surfaces.

Disabling the legacy regex writer by default is the right call. It prevents low-quality duplicate rows while preserving historical data and allowing future TripleStore reader migration.

## Privacy and local-first behavior

The change uses the existing host, local, and remote LLM paths. It adds no new external service or telemetry path. Local-LLM load failures remain non-fatal and return `([], [])`.

The default behavior reduces unwanted persistence of conversational filler in `memoria_kg`. No change erodes the local-first design.

## Integration surfaces

Hermes continues to use the existing `remember()` and extraction pipeline. `TripleStore.query()` and the triple-query handler now expose stored LLM triples. No MCP or CLI changes are reported.

## Maintainability

Validation is centralized in `validate_kg_triples()`. Tests cover parsing, null and filler rejection, normalization, truncation, deduplication, provenance, dates, supersession, caching, and public query surfaces. Malformed reasoning traces receive guards at four LLM call sites.

The separation between fact extraction and triple extraction preserves compatibility and supports temporal knowledge-graph storage. PR `#847` must merge before this change.


AxDSan (changes requested):
The wiring is good and the validator is real, but the description promises two behaviors the code does not have, and I would rather that be fixed than discovered by whoever trusts it.

## Verified working

`tests/test_kg_triple_wiring.py` + `tests/test_extraction_integration.py`: **22 passed**. Exercised the validator directly:

| Input | Result |
|---|---|
| `('user','prefers','neovim')` | kept |
| `('','prefers','x')` | dropped |
| `('user','prefers',None)` | dropped, no raise |
| `('user','Prefers Strongly','x')` | → `prefers_strongly` |
| 599-char object | truncated to 299, not ending mid-space |
| same triple twice | deduped to one |

Word-boundary-safe truncation, snake_case normalization, empty/None rejection and within-batch dedup all do what they say. Parsing `kg` instead of dropping it on the floor is plainly right, and reusing the same LLM pass through a content-keyed cache rather than a second call is the correct instinct.

## The description is wrong about filler rejection

The body says: *conversational-filler rejection ("you said", "let me check")*. Neither is rejected:

```
subject='you said'              -> [('you said', 'prefers', 'x')]
object ='let me check'          -> [('user', 'prefers', 'let me check')]
subject='let me check the logs' -> kept
```

`_KG_FILLER_PREFIXES` is:

```
['i guess ', 'i mean ', 'i think ', 'it seems ', 'kinda ', 'maybe ',
 'not sure ', 'perhaps ', 'possibly ', 'probably ', 'sorta ', 'well ',
 'what ', "what's ", 'whats ', 'whether ']
```

Hedges, not conversational openings. Both named examples are absent. Fix whichever is wrong, the list or the sentence, but they must agree: this is the exact class of drift where someone later reads the description, assumes dialogue fragments are filtered, and stops checking.

While in there: every prefix carries a trailing space, so a subject that is exactly `"maybe"` or `"well"` with nothing after it does not match. Probably deliberate to avoid eating `"Wellington"`, but it means a one-word filler survives. A trailing-space-or-end match would close it.

## Merge-order hazard

`#848` contains `#847`'s commit (`529a5aa contain memoria_kg regex prototype writer...`) and both target `main`. Merge #848 first and #847 becomes empty; merge #847 first and this rebases cleanly. The description says to review #847 first, which is right, but nothing enforces the *merge* order. Either retarget this at `contain-memoria-kg` so GitHub sequences it, or say plainly in both descriptions that #847 lands first.

## Carried from my #847 review

`_store_kg_triples` writes to the `triples` table, while the two structured-recall specialists at `beam.py:5910` and `beam.py:5936` read `memoria_kg`. With the regex writer off by default and the LLM triples going elsewhere, negation and entity-action structured recall are fed by nothing on new data. That needs to be a stated decision in one of these two PRs rather than an emergent property of landing both.

Prediction6767:
Addressed the CHANGES_REQUESTED items from @AxDSan:

1. Filler rejection now matches the described examples — `_KG_FILLER_PREFIXES` includes `you said` / `let me` / `let me check`, with stem trailing-space-or-end matching so bare hedges like `maybe`/`well` drop without eating `Wellington`.
2. Merge order stated baldly in both PR bodies: land #847 first, then #848 (upstream has no `contain-memoria-kg` base to retarget against).
3. memoria_kg consumer gap: accepted temporary dormancy for new rows; historical rows stay; TripleStore repoint = follow-up.

Also merged current `main` into the stack and routed the `MNEMOSYNE_REGEX_KG` gate through `_env_truthy` (env-only). Re-requesting review.

Prediction6767:
Addressed the three CodeRabbit minors on the new head (e1c067d):

1. **JSON null triple fields are rejected before `str()` conversion** — a payload like `[[null, "uses", "sqlite"]]` no longer becomes the literal string `"None"` and is dropped at parse time instead of persisting junk.
2. **Predicates now normalize to strict snake_case** — whitespace and hyphen runs fold to a single underscore, boundary underscores are trimmed, so `works-at` and `works at` can no longer persist as separate current truths (supersession keys on subject+predicate).
3. **`_extract_facts_impl` returns the declared `([], [])` tuple** on local-LLM load failure instead of a bare `[]`, matching the documented `(facts, triples)` shape.

+4 test cases in `tests/test_kg_triple_wiring.py` (18 passed locally, ruff clean). The 6 `test_extraction_integration.py` failures are the pre-existing Windows temp-lock issue (identical on unmodified HEAD).

Prediction6767:
Follow-up (74ad82a) for the CodeRabbit note on separator-only predicates: after snake_case normalization, a predicate that collapses to empty (`-`, `___`, `  - -  `) is now rejected instead of persisted. Regression test added (19 pass locally).

Prediction6767:
Friendly ping — the CHANGES_REQUESTED items from @AxDSan were all addressed on 2026-08-29 (filler rejection matches the described examples, plus follow-ups `74ad82a`/`e1c067d` covering the CodeRabbit minors, including the separator-only-predicate note from the last review round).

One heads-up on CI: the tests (3.10–3.13), docs-check, and build are green on head `74ad82a`; only `lint` shows red, and its annotation is an actions/checkout Node-deprecation warning ending in `exit code 128` — a runner/git infra hiccup, not a lint violation in the diff. A maintainer re-run of that job should clear it (I can't re-run workflows as an outside contributor).

Ready for another look when you have bandwidth — and if this split doesn't fit the roadmap right now, just say so and we'll keep carrying it locally without pressure. Thanks!

dplush:
Thanks for the substantial follow-up. The parser and validation work are heading in the right direction, but I still see a few merge blockers on the current head:

1. **Session scope must not leak through KG extraction.** `remember()` defaults to `scope="session"`, while the new TripleStore writes are bank-wide and carry no session, author, channel, or scope information. A triple extracted in one session is therefore queryable from another session in the same bank. Please decide and implement the intended visibility contract before merging.

2. **The advertised single extraction pass is not happening yet.** `_extract_and_store_facts()` calls both `extract_facts_safe()` and `extract_triples_for_beam()`. The former bypasses `_EXTRACT_CACHE`, so `remember(extract=True)` currently invokes the backend twice. Please make the fact and KG accessors share one cached extraction result and add a call-count regression test.

3. **Predicate normalization does not yet enforce snake_case.** Whitespace and hyphens are normalized, but predicates such as `works@acme`, `works/acme`, and `works.acme` remain accepted. That permits distinct current-truth keys for semantically equivalent predicates. Please either enforce a strict allowed predicate alphabet or narrow the documented contract.

4. **Please clean up the integration state before the next review.** The branch is behind and conflicting with `main`; it also carries an unintended CI workflow rollback and a whole-file CRLF churn in `CHANGELOG.md`. The current lint job is red because its shallow diff step cannot find a merge base. Please rebase after #847, restore unrelated CI changes, normalize the changelog diff, and bring the full check set back to green.

The focused KG tests pass locally, but the scope boundary and duplicate-LLM-call behavior need to be resolved before this is safe to merge.

PR #855, opened 2026-08-25
fix(extraction): dedupe items repeated across parse categories

arcalumis:
## Description

`_parse_facts()` flattens the JSON extraction result with `extend()` across `('facts', 'instructions', 'preferences', 'timelines')` and no dedupe, so a statement the model files under more than one category is returned once per category.

Those categories overlap **by design** — the prompt's own examples make that unavoidable:

- `instructions`: *"always use tabs"* is also a fact about the user
- `timelines`: *"release on 2024-12-01"* is also a fact

So a model that files one statement under every category it belongs to is answering the prompt correctly, and the parser punishes it for that. The duplicates land in the `facts` table, which is keyed only on `fact_id` and so has no constraint to absorb them.

This is model-dependent, which is why it can go unnoticed: a model that leaves `instructions`/`timelines` empty is never affected. Observed with Qwen3.5-9B against a 3-sentence input — 6 items returned for 4 distinct statements:

```json
{
  "facts": ["The new POS system goes live October 1st",
            "Servers must clock out through the tablet by the walk-in, not the front terminal",
            "Tip-outs will be calculated nightly instead of weekly",
            "Marcus is training the closers"],
  "instructions": ["Servers must clock out through the tablet by the walk-in, not the front terminal"],
  "timelines": ["The new POS system goes live October 1st"]
}
```

Both repeats are correct categorisations. After this change the same payload yields the 4 distinct statements.

Items are deduped on trimmed, case-folded text with the first occurrence winning, so `facts` entries keep their position and ordering is unchanged for callers. Genuinely distinct statements are untouched.

### Scope

Deliberately narrow — **the `kg` category is not touched here.** `_parse_facts()` also drops `kg` entirely, but that is #840 and the fix is already designed in #847/#848 (parse into typed SPO tuples and write to the `TripleStore`, not into the flat fact list). This change is orthogonal to those and should not conflict with them.

The legacy newline-splitting path is also left alone; the cross-category defect is specific to the JSON branch.

## Related Issue

None open for this specific defect — #840 covers the neighbouring `kg` drop. Happy to file one first if you'd prefer the issue-first flow.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor / performance
- [ ] Test improvement
- [ ] CI / build tooling

## How Has This Been Tested?

- [x] Existing tests still pass — `MNEMOSYNE_NO_EMBEDDINGS=1 pytest tests/ -v`
- [x] New tests added
- [x] Manual verification

Three tests added to `tests/test_extraction.py`:

| test | asserts |
|---|---|
| `test_parse_facts_json_dedupes_across_categories` | the overlapping payload above returns 2 items, `facts` order preserved |
| `test_parse_facts_json_dedupe_ignores_case_and_padding` | comparison is trimmed and case-folded |
| `test_parse_facts_json_keeps_distinct_items` | four genuinely different statements all survive |

The first two fail on `main` and pass with the change; the third is a regression guard that passes either way.

Manually verified end-to-end through `extract_facts()` against a live OpenAI-compatible endpoint with two different models — Qwen3.5-9B (populates the overlapping categories, previously 6 items for 4 statements) and Gemma-4-26B-A4B (leaves them empty, previously unaffected). Both now return 4 items, 4 unique.

## Checklist

- [ ] Version bumped in `mnemosyne/__init__.py` — **deliberately not done.** `__version__` is currently `4.0.0b1` and there are a number of PRs open; bumping here seemed more likely to cause a conflict than to help. There is precedent for deferring in the current `Unreleased` notes ("the version bump is deferred to that PR to avoid a duplicate bump"). Happy to add it if you'd rather it ship with the change.
- [ ] README updated if user-facing behavior changed — not applicable, no documented behaviour changes
- [x] Code follows the project's principles — no new dependencies; the change is contained to one function and adds no imports

## Summary

- `_parse_facts()` now removes duplicate statements across supported categories.
- Deduplication uses trimmed, case-folded text.
- The first occurrence remains unchanged and preserves order.
- Supported categories are detected by key presence.
- Empty, null, scalar, and unusable supported categories return no facts.
- Unsupported and `kg`-only payloads retain existing fallback behavior.
- `kg` handling and legacy newline parsing remain unchanged.

## Architectural impact

- **Core memory architecture:** Improves fact quality before persistence. The PR does not change working, episodic, or BEAM memory tiers.
- **Retrieval and consolidation:** Prevents duplicate facts from reaching the AnnotationStore, facts table, and `fact_recall`. Retrieval strategies and consolidation remain unchanged.
- **Veracity system:** No changes affect validation, confidence, or veracity handling.
- **Privacy and local-first guarantees:** Deduplication uses local text normalization only. The PR adds no data collection, remote processing, or synchronization. The local-first design remains intact.
- **Agent integration:** No changes affect Hermes, MCP, or CLI integration surfaces.
- **Sync layer:** No changes.

## Validation and maintainability

- Tests cover cross-category deduplication, normalization, ordering, distinct items, unusable values, fallback behavior, and storage-path uniqueness.
- Storage tests verify that each overlapping fact reaches the AnnotationStore, facts table, and `fact_recall` exactly once.
- Reported CI hangs occurred on pristine `main` and were tracked separately.
- The changelog documents the behavior.
- Preserving `kg` handling and legacy parsing limits regression risk. This is the right call for a focused maintenance change.


dplush (changes requested):
Thanks for the focused fix. The cross-category duplicate path is real, and keeping `kg` out of this PR is the right boundary.

Please address the two current-head findings before merge:

1. A valid supported JSON payload whose entries are all empty or whitespace must return `[]`. On this head it falls through to the partial-JSON fallback and can turn keys such as `instructions` and `preferences` into stored facts. Track the presence of a supported category and return the deduped list even when empty, while preserving the malformed/unsupported fallback and the `kg`-only path.
2. Make the distinct-items regression assert the complete expected list in first-occurrence order, rather than only its length.

One release-note cleanup: please reduce the new `CHANGELOG.md` entry to a concise factual Fixed bullet. The current entry repeats the full rationale and is much too large for an Unreleased note.

After that, please wait for fresh CI and CLA status on the updated head.

arcalumis:
Both findings are addressed and the CHANGELOG entry is trimmed. Pushed as 690446f.

**1. Empty supported payload returns `[]`.** Confirmed the failure first — on the previous head a valid payload with nothing to extract returned `['instructions', 'preferences']`, because the partial-JSON fallback matches any quoted run of ten or more characters and the schema's own key names qualify (`instructions` is 12, `preferences` is 11).

The categories are now answered by presence rather than by yield: seeing a supported category returns the deduped list even when empty. No supported category still falls through, so malformed/unsupported payloads and the kg-only path are unchanged. `test_parse_facts_json_empty_categories_yield_no_facts` covers all-empty, whitespace-only, and empty-plus-`kg`; `test_parse_facts_json_without_supported_categories_falls_back` pins the kg-only fallback so a later change cannot quietly take it away.

**2. Distinct-items regression** now asserts the full expected list in first-occurrence order.

**3. CHANGELOG** reduced to a two-sentence `Fixed` bullet.

---

**On the red `test` jobs — they are not from this PR, and they will fail the same way on a fresh run.**

The four `test` matrix jobs did not fail an assertion. They hung and were cancelled at the 6h ceiling. The `-v` output pins the location:

```
04:41:49  tests/test_mcp_streamable_http.py::TestStreamableHttpBearerRejection::
          test_401_response_has_www_authenticate_header PASSED [61%]
10:33:09  ##[error]The operation was canceled.
          Terminate orphan process: pid (2280) (pytest)
```

The next test in file order is `TestStreamableHttpBearerRejection::test_valid_bearer_completes_initialize_and_tools_list`, and it hangs locally too — reproduced on **pristine `main` with no changes applied**, so it is not caused by anything here:

```
$ pytest tests/test_mcp_streamable_http.py -q --timeout=20 --timeout-method=thread
...
  File ".../anyio/from_thread.py", line 560, in start_blocking_portal
    thread.join()
  File ".../threading.py", line 1167, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
+++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++
```

It blocks in `anyio.from_thread.start_blocking_portal` waiting on a portal thread that never terminates — the first test in that file to drive a full request through the app rather than only build it. Local anyio is 4.14.2; since `pyproject.toml` does not pin an upper bound, CI resolves whatever is current, which would explain jobs that passed on this file earlier now hanging without any change to the file itself.

Happy to open a separate issue with the full trace. It also seems worth adding `pytest-timeout` with a per-test cap to CI — a hang currently costs six hours per job across four jobs and reports as a plain red X.

This PR touches only `_parse_facts`; the extraction suites pass locally:

```
$ pytest tests/test_extraction.py tests/test_extract_parity.py \
         tests/test_extraction_integration.py tests/test_c13b_extraction_diagnostics.py -q
75 passed
```

CLA is signed.

arcalumis:
Filed the CI hang as #871 with the version bisect, thread stacks, and a probe showing session semantics are still correct under the new SDK — so it's a test-lifecycle problem, not a product regression. The red `test` jobs here are that, not this change.

arcalumis:
Good catch from CodeRabbit — valid, and the same defect through a different value shape. Verified before fixing: on the previous head a payload answering `"facts": null` instead of `[]` still returned `['instructions', 'preferences']`, and a scalar smuggled the key names in beside real content.

Presence of the key now decides completeness rather than the value being a list. A document that parsed as a dict and names a supported category is complete, not partial, so an unusable value means "nothing to extract" instead of running the streaming-recovery regex. Items are still only iterated when the value is a list; a payload naming no supported category still falls through, so the kg-only and unparseable paths are untouched (both pinned by tests).

Pushed as 2762695 with regression coverage for null, `{}`, `0` and the scalar case. 76 passing across the four extraction suites.

arcalumis:
Both addressed in b3f48c4.

**Scalar assertion** tightened to `scalar == []`. It was permissive because I was hedging on whether a bare string should be coerced into a fact; pinning it is better, and the comment now says why the answer is no — the schema promises a list, and coercing would invent a contract the prompt never stated.

**Storage-path regression** added as `TestOverlappingCategoriesReachStorageOnce`. You were right that the gap was real: the existing fixtures patch `extract_facts_safe` with an already-distinct list, so they cover the write path but never the parser feeding it — the dedupe could regress with nothing failing, since the only thing that observes it is what reaches the stores. The new test drives real `_parse_facts` output from an overlapping payload through `remember(extract=True)` and asserts no repeats reach the AnnotationStore, the `facts` table, or `fact_recall`.

Verified it actually catches the regression rather than just passing: against `origin/main` both cases fail, the second on `assert 2 == 1` from `fact_recall` returning two rows for one fact. That is the user-visible harm this PR is about, and it now has a test.

78 passing across the four extraction suites.

arcalumis:
Fixed in 80bc5ff — the weakness was real: an empty result set is duplicate-free too, so a `fact_recall` that returned nothing would have satisfied the old assertion without proving anything.

One correction to the suggested form. Asserting the fact appears exactly once by equality fails against a *correct* implementation, because `fact_recall` echoes the stored subject (`user stated Servers must clock out through the tablet`) rather than the fact verbatim, so the exact count reads zero. It now matches by containment and requires exactly one hit — presence and uniqueness in one assertion.

Verified both directions: passes on this head, and still fails against `origin/main` where the row appears twice. 78 passing across the four extraction suites.

arcalumis:
Valid, and the same vacuous-pass shape as the previous one — an empty annotation store is duplicate-free, so a write path that persisted nothing would have satisfied the uniqueness-only assertion. Fixed in de3ac43.

Stored annotation values are now compared against the parsed facts outright rather than only checked for repeats, and the POS statement is verified through `fact_recall` alongside the tablet one. Worth checking both: they are filed under different category pairs — facts + instructions versus facts + timelines — so either could regress without the other. The uniqueness assertions are retained.

Verified both directions: 8 passing in the parity suite on this head, still failing against `origin/main` where both rows appear twice. 78 passing across the four extraction suites.

Issue #857, opened 2026-08-25
Native Windows: install defaults to symlink mode, which is privilege-gated and fails as WinError 1314


AxDSan:
## Summary

On native Windows, `mnemosyne-hermes install` defaults to a mode that cannot succeed without elevated privileges. It works for users who happen to have Developer Mode enabled or who run elevated, and fails for everyone else. From the outside this looks random, which is how it has been described by the people hitting it: *"hit or miss for some as to whether or not it magics itself into working."*

It is not random. It is deterministic on a condition nobody thinks to check.

## Evidence

The install mode defaults to `symlink` in both entry points on current `main`:

```
integrations/hermes/src/mnemosyne_hermes/install.py:1763   mode: str = "symlink"
integrations/hermes/src/mnemosyne_hermes/install.py:2156   mode: str = "symlink"
```

On Windows, `os.symlink` raises `WinError 1314` (`ERROR_PRIVILEGE_NOT_HELD`) unless the account holds `SeCreateSymbolicLinkPrivilege`, which in practice means Developer Mode is on or the shell is elevated. We already recognize that error:

```
install.py:94   def _is_windows_symlink_privilege_error(error: OSError) -> bool:
install.py:96       return getattr(error, "winerror", None) == 1314
```

Two things are missing:

1. **Nothing selects a different default on Windows.** `os.name == "nt"` appears only for command-line formatting and one unrelated helper. There is no branch that prefers wrapper mode on the one platform where the default mode is privilege-gated.
2. **We never create an NTFS junction.** There is no `mklink`, junction or reparse-point creation anywhere in `integrations/hermes/src/`. A junction to a directory needs no special privilege and is the standard Windows answer to exactly this problem.

#807 and #809 improved the situation on 2026-08-19: the installer now fails closed on 1314 and prints a persistent-wrapper retry command using the resolved Hermes Python, and native Windows venv discovery finds `Scripts/python.exe`. That work explains the fix. It does not perform it, and the default was left unchanged.

## Why the community workarounds did not work

The advice circulating is to symlink the install into the Hermes home. That fails for a second, separate reason: people are linking a **source checkout**, while the only target that works is the installed package directory. The working recipe found during triage was a junction:

```bat
cd /d C:\Users\<you>\AppData\Local\hermes\plugins
mklink /J mnemosyne C:\Users\<you>\AppData\Local\hermes\hermes-agent\venv\Lib\site-packages\mnemosyne_hermes
```

[...]

PR #858, opened 2026-08-25
docs(hermes): wrapper mode is the general answer, not a Docker workaround

AxDSan:
Persistent side-venv wrapper mode is documented as *"the canonical **Docker** installation"*. It is the canonical installation for **three** situations, and filing it under one of them is costing us users in the other two.

| Situation | What happens | Today's docs |
|---|---|---|
| Docker | venv rebuilt on image update | ✅ covered |
| **Desktop binary installer** | bundled Python env wiped and rebuilt on update | ❌ not mentioned anywhere |
| **Native Windows** | symlink needs Developer Mode or elevation, fails `WinError 1314` | ❌ not connected to wrapper mode |

Same mechanism, same underlying reason: something outside the user's control replaces or restricts the Hermes Python runtime, and Mnemosyne has to survive it.

### Why this is urgent rather than tidy

Teknium raised this against our instructions directly, while designing the Hermes Desktop binary rollout:

> mnemosyne instructions say to pip install mnemosyne into the hermes agent's venv, which in the future compiled binary installs, would get wiped and remade every update

He read our docs and found the wrong instruction. We had already built the right one and shelved it under Docker.

Separately, native Windows users have been reporting the install as *"hit or miss for some as to whether or not it magics itself into working"*. That is `WinError 1314`, tracked in #857, and wrapper mode is the fix there too.

### Changes

- **A mode-selection table in Step 1**, placed *before* the install commands. The current order lets a Desktop or Windows user copy the failing command before reaching the note that would have saved them.
- **The wrapper callout leads with when to use it**, not with Docker. Docker paths are unchanged and remain the worked example, since they are the least guessable.
- **A native Windows block** naming `WinError 1314` as the cause, and correcting the circulating workaround: linking a Mnemosyne *source checkout* has the wrong target even when the link succeeds. The only correct target is the installed package directory in `Lib\site-packages\mnemosyne_hermes`.
- **The uninstall heading** no longer says "Docker-image install".

### Scope

Documentation only, one file, +48/-2. No behaviour change. The default install mode is still symlink; changing that on Windows is #857 and belongs in code, not here.

`verify-docs.py` reports pre-existing sibling drift in `content/api/tool-schema.mdx`, unrelated to this change and present on `main`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update

## How Has This Been Tested?

- [x] Manual verification: headings and the new anchor resolve, Docker instructions unchanged

## Summary

- Updates Hermes installation guidance for Docker, Desktop binary installers, and native Windows.
- Makes persistent side-venv wrapper mode the recommended option when runtime replacement or symlink restrictions apply.
- Keeps default symlink installation mode unchanged.
- Adds guidance for Desktop environment replacement and native Windows `WinError 1314`.
- Corrects the Windows package target to the installed `mnemosyne_hermes` directory.
- Separates POSIX-shell and native Windows PowerShell uninstall instructions.
- Directs WSL users to the POSIX instructions.

## Architecture and maintainability

- Does not change retrieval, consolidation, veracity, sync, benchmarks, privacy controls, or local-first guarantees.
- Does not change Hermes, MCP, or CLI interfaces.
- Improves Hermes deployment maintainability by documenting consistent wrapper-mode behavior across restricted environments.
- Preserves local execution and does not erode the local-first design.
- This is the right call for installation reliability.


AxDSan:
Valid finding, fixed in 983a65d.

The heading did claim native Windows while the block underneath was Bash: `export`, and `"$VENV/bin/mnemosyne-hermes"`. A native Windows venv keeps its executables in `Scripts\`, and PowerShell will not run that block at all, so the platform most likely to be reading a wrapper-mode section could not follow the instructions in it.

The POSIX heading now says POSIX shells, and native Windows gets its own PowerShell block using `Scripts\mnemosyne-hermes.exe` and `Scripts\python.exe`. That matches the layout `_venv_python_candidates()` actually probes (`install.py:869-875`), rather than a plausible guess at the path. WSL is a POSIX layout and is pointed back at the first block explicitly, since that is the ambiguity this doc keeps running into.

PR #864, opened 2026-08-25
fix(hermes): stop defaulting native Windows to a privilege-gated install mode

### LINKED ISSUE #857: Native Windows: install defaults to symlink mode, which is privilege-gated and fails as WinError 1314

AxDSan:
Closes #857.

`mnemosyne-hermes install` defaulted to `symlink` on **every** platform. Windows only permits creating a symbolic link when the account holds `SeCreateSymbolicLinkPrivilege`, which in practice means Developer Mode is enabled or the shell is elevated. Without one of those, `os.symlink` raises `WinError 1314` and the install fails.

That is why native Windows has been reported as *"hit or miss for some as to whether or not it magics itself into working"*. It is not random. It is deterministic on a privilege nobody thinks to check.

### The change

The default is now decided in one place:

```python
def default_install_mode() -> str:
    return DEFAULT_INSTALL_MODE_WINDOWS if _is_windows_platform() else DEFAULT_INSTALL_MODE_POSIX
```

| Platform | Default | Change |
|---|---|---|
| Linux, macOS, WSL | `symlink` | none |
| **Native Windows** | **`wrapper`** | was `symlink`, which cannot succeed unprivileged |

Wrapper mode writes a real plugin directory with an import bootstrap and needs no privilege. It is the same mechanism already recommended for Docker, for a related reason. `--mode symlink` still works for anyone who does hold the privilege.

**It needs no extra arguments.** `_validated_wrapper_environment` falls back to `sys.executable` and verifies the selected interpreter can import `mnemosyne_hermes` *before* touching an installed plugin, so the new default is usable as-is.

### Two things I got wrong first, and what they cost

**Resolving inside `run_install` was not enough.** `--dry-run` reads `args.mode` directly, so it printed `Install mode: None` and stopped detecting a refused wrapper-to-symlink migration. Caught by `test_dry_run_reports_refused_wrapper_migration`. Resolution now happens at the CLI boundary, before anything reads it.

**My first `WinError 1314` message contradicted itself.** I added `mnemosyne-hermes install --mode wrapper` to the shared preamble, but the branch below it says a wrapper retry cannot be generated when no Hermes Python was resolved. The output told the user to run a command and then said it could not offer one.

The fix was to my text, not to the test. `test_run_install_does_not_offer_wrapper_without_a_resolved_hermes_python` protects a real invariant, so the preamble now carries only the fact that wrapper mode is the default and needs no privilege, and the two branches below keep ownership of the concrete commands.

### Testing

```
integrations/hermes suite:  437 passed  (main: 427)
                            4 failed    (main: 4, identical)
ruff E9,F63,F7,F82 tree-wide:  All checks passed
ruff F,RUF022 on changed files: All checks passed
```

The 4 failures are identical on `main` and are local environment artifacts (a real `~/hermes-agent/.venv` leaking into discovery, plus a version-metadata mismatch), not regressions. I checked the baseline rather than assuming.

10 new cases in `test_windows_default_install_mode.py`: the platform default both ways, mode resolution through `install_plugin`, rejection of an invalid mode, and that the CLI carries no baked-in default so the platform actually gets asked.

### What this does not do

I have no Windows machine. The decision logic is unit-tested by faking the platform predicate; the end-to-end behaviour on real Windows is unverified by me. Worth one confirmation from someone on native Windows before this is relied on.

This is the code half of #857. The documentation half is #858.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests still pass (compared against a `main` baseline)
- [x] New tests added for the new behaviour
- [x] `ruff` on both CI gates

## Summary

- Native Windows now defaults to `wrapper` mode for Hermes installs.
- Linux, macOS, and WSL continue to default to `symlink`.
- Explicit `--mode symlink` remains supported.
- CLI, dry-run, and migration checks use the resolved platform-specific mode.
- Wrapper mode validates imports and falls back to `sys.executable`.
- `WinError 1314` guidance now reflects explicit mode selection.
- Added tests for platform defaults, mode resolution, invalid modes, and CLI behavior.

## Architecture impact

- Changes are limited to the Hermes installation surface.
- Working, episodic, and BEAM memory tiers are unchanged.
- Retrieval, consolidation, veracity, sync, MCP, and benchmark methodology are unchanged.
- The privilege-free Windows default strengthens local-first operation.
- No remote service, telemetry path, or data-sharing behavior changes.
- No local-first guarantees are eroded.
- Centralized mode resolution improves maintainability and keeps installation behavior consistent.
- The Windows `wrapper` default is the right call because it removes a privilege requirement without changing Hermes integration semantics.


AxDSan:
Both CodeRabbit findings verified against the code rather than taken at face value. One was right, one was half right, and one is real but not this PR's business.

**Changelog wording: valid, fixed in 5309f5a.**

The final sentence implied the installer switches modes after a `WinError 1314`. It does not, and nothing in this diff makes it. `default_install_mode()` resolves an omitted `--mode` to wrapper *before* installation begins, and an explicit `--mode symlink` that hits the privilege error is never switched on the user's behalf: it fails with recovery guidance, which is exactly what the error text says. Reworded to describe resolution rather than switching. Worth catching, since that sentence would have shipped in a release note.

**Hermes README: not valid, no change.**

The finding says the README's unqualified "No symlinks" omits platform and explicit-mode behavior. That phrase is at `integrations/hermes/README.md:123`, in the paragraph describing how the entry point registers with Hermes' provider discovery:

> The entry point in `mnemosyne-hermes` registers with Hermes' memory provider discovery system (`hermes_agent.memory_providers`). No symlinks. No directory copying. No plugin.yaml gymnastics.

That is about the registration mechanism, not about `--mode`. It is not describing install modes and does not need to qualify them.

**Wrapper preflight interpreter: real, filed as #883.**

`check_mnemosyne_core()` does run in the CLI interpreter, before mode branching, and returns 1 there. So `--mode wrapper --python <other>` can fail on the CLI interpreter while the selected one is fine, which contradicts `--python` being authoritative. The comment immediately below it even acknowledges that wrapper mode validates a different interpreter.

It is pre-existing though, unchanged by this diff. What this PR changes is exposure: making wrapper the Windows default means more people land in the mode where the ordering is wrong. That is an argument for fixing it soon, not for holding a PR that stops Windows installs failing on a privilege nobody checks. #883 has both candidate fixes written up.

PR #866, opened 2026-08-25
test(mcp): tear down SSE GET before client exit

### LINKED ISSUE #860: CI: the test matrix intermittently hangs in test_mcp_streamable_http and burns the 6-hour job timeout
## Summary

The test matrix intermittently hangs and is killed six hours later by the GitHub Actions job timeout. When it happens, every test job in the run is lost, so the pull request shows `cancelled` rather than a real result. **No pull request can be merged on a verified green gate while this is happening.**

## Evidence

Run `32815197342`, job `test (3.13)`:

```
06:01:45  job starts
06:07:01  tests/test_mcp_streamable_http.py::TestStreamableHttpBearerRejection::test_401_response_has_www_authenticate_header PASSED [ 61%]
          ... nothing for 5h 55m ...
12:01:58  ##[error]The operation was canceled.
12:01:59  Terminate orphan process: pid (2524) (pytest)
```

All four interpreters in that run behave identically: last output at 61%, then silence until cancellation. The suite stops immediately **after** `test_401_response_has_www_authenticate_header`, which means the test that hangs is the next one in the class, `TestStreamableHttpBearerRejection::test_valid_bearer_completes_initialize_and_tools_list` (`tests/test_mcp_streamable_http.py:460`).

Observed today on at least four runs across four different branches, two of which have already burned the full six hours and been cancelled. It is intermittent, not constant: runs at 09:20, 10:12, 10:45 and 10:54 completed normally in about nine minutes each.

## Likely mechanism

That test drives an authenticated SSE `GET` on a background thread:

```python
get_thread = threading.Thread(target=_drive_get, daemon=True)   # :569
...
portal.call(authed_app, scope, receive, send)                    # :566
```

The test's own waits are bounded (`get_started.wait(10)`, `get_thread.join(10)`), so the hang is not in the assertions. The exposure is teardown: if the SSE task is still in flight when the `with TestClient(...)` block exits, the AnyIO blocking portal shuts down and waits for that outstanding request. The thread being a daemon does not help, because the portal is what blocks.

That also explains the intermittency. The stream is supposed to be terminated by the preceding `DELETE`; whether it has actually finished by the time the context manager exits is a race.

## Impact

- Every affected run wastes roughly 24 runner-hours (four jobs, six hours each).
- Contributors see their pull requests marked `cancelled` and reasonably assume they broke something.
- Any batch refresh of stale pull requests is off the table until this is fixed, because roughly half the runs would hang.

## Two separate fixes

**1. The test.** Guarantee the SSE request is cancelled before the client context exits, rather than relying on the `DELETE` having drained it in time.

**2. The suite, regardless of this test.** A hang should fail in minutes, not six hours. Adding `pytest-timeout` with a per-test limit would convert any future hang from a lost day of CI into a named failing test. This repository has no such backstop today, which is why a single blocking teardown can consume a full job timeout.

The second is worth doing even after the first lands, because it bounds the cost of the next occurrence rather than this one.

## Introduced by

`tests/test_mcp_streamable_http.py` arrived with #749, merged as `72fcd67` on 2026-08-23.

dplush:
## Summary

Make the long-lived authenticated SSE GET in the Streamable HTTP integration test finish explicitly before `TestClient` tears down its portal.

## Why this path

The test itself owns the background ASGI request. A `finally` block now signals `http.disconnect` and joins that request before leaving the client context. This keeps the existing DELETE-driven assertion meaningful: it still verifies that DELETE alone shuts the request down before fallback cleanup runs.

## Non-goals

- No product transport behavior changes.
- No CI workflow or job-timeout changes.
- No pytest-level timeout. That remains the separate suite-level guardrail discussed in #860.

## Verification

```text
uv run --extra mcp --extra test pytest -q tests/test_mcp_streamable_http.py
54 passed
```

Independent current-diff review: approved.

Fixes #860

## Summary

- Updates `tests/test_mcp_streamable_http.py` to disconnect the background SSE `GET` request during cleanup.
- Joins the request thread before `TestClient` exits, including when assertions fail.
- Preserves the `DELETE` assertion that verifies request shutdown.

## Architecture impact

- No changes to Mnemosyne’s memory architecture, retrieval, consolidation, veracity, sync, privacy, or local-first guarantees.
- No changes to Hermes, MCP, CLI, or product transport behavior.
- Improves test maintainability by removing an intermittent teardown race.
- Does not add a pytest-level timeout safeguard.
- Verification: `54 passed` in `tests/test_mcp_streamable_http.py`.

### DIFF
--- a/tests/test_mcp_streamable_http.py
+++ b/tests/test_mcp_streamable_http.py
@@ -524,6 +524,7 @@ class TestStreamableHttpBearerRejection:
             # response start before terminating the session with DELETE.
             portal = client.portal
             get_started = threading.Event()
+            get_disconnect = threading.Event()
             get_result = {}
 
             def _drive_get():
@@ -548,8 +549,8 @@ class TestStreamableHttpBearerRejection:
                 }
 
                 async def receive():
-                    while True:
-                        await asyncio.sleep(3600)
+                    await asyncio.to_thread(get_disconnect.wait)
+                    return {"type": "http.disconnect"}
 
                 async def send(message):
                     if message["type"] == "http.response.start":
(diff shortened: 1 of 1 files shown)


dplush:
Superseded by #867, which fixes the actual MCP 2.1 ASGI contract violation by delivering the initial http.request, handles portal-task cancellation on every failure path, and adds the suite-level timeout backstop. Closing this narrower teardown-only attempt to avoid duplicate fixes.

PR #870, opened 2026-08-25
fix(hooks): inspect added staged PII lines

### LINKED ISSUE #869: .githooks/pre-commit blocks every commit that touches pyproject.toml
The pre-commit hook greps the **contents** of staged files for the maintainer noreply address:

```sh
PII_PATTERNS="1641797+AxDSan@users.noreply.github.com"
staged=$(git diff --cached --name-only | grep -v "^$HOOK_FILE$" || true)
if echo "$staged" | xargs grep -l "$PII_PATTERNS" 2>/dev/null; then
  echo "ERROR: PII detected in staged files. Remove before committing."
```

That address is on `main`, in `pyproject.toml:13`, as the package authorship metadata:

```toml
authors = [
    {name = "Abdias J", email = "1641797+AxDSan@users.noreply.github.com"},
]
```

So any commit touching `pyproject.toml` is rejected, whatever the diff. It is not checking whether the line was *added*, only whether the file contains it. #867 needed `--no-verify` for a diff that does not go near line 13.

The hook is also inconsistent with itself: the second half prints

```
INFO: Git user.email set to the noreply address (expected for Mnemosyne repo). Skipping check.
```

so the same address is "expected" as a git identity and blocked as PII in file contents.

## Suggested fix

Check added lines rather than file contents, and skip the file that is supposed to carry it:

```sh
added=$(git diff --cached -U0 -- . ':(exclude)pyproject.toml' | grep '^+' | grep -v '^+++')
if echo "$added" | grep -q "$PII_PATTERNS"; then
  ...
fi
```

Two notes on what this hook can and cannot do. The address is a GitHub noreply address, which is public by construction and already in the published package metadata on PyPI, so the blast radius of a leak is small. And a local hook is advisory: it does not run in CI, contributors must opt in via `core.hooksPath`, and `--no-verify` skips it. Worth keeping as a speed bump, not worth breaking `pyproject.toml` edits for.

dplush:
## Summary

Fix the repository pre-commit PII guard so it scans only newly added content in the staged index rather than the whole current contents of changed files.

## Why this path

`pyproject.toml` legitimately contains the maintainer's GitHub noreply address. The old whole-file scan therefore blocked every unrelated staged edit to that file.

The guard remains fail-closed:

- ordinary added PII is rejected;
- added content beginning with `+++` is rejected;
- staged binary/NUL content is scanned and rejected when it contains PII;
- a `git diff` failure aborts the hook rather than being treated as no match.

The existing git-config INFO behavior is unchanged.

Closes #869.

## Verification

- `uv run --frozen pytest -q tests/test_precommit_pii.py tests/test_release_gates.py` — 17 passed
- `/bin/sh -n .githooks/pre-commit`
- focused Ruff: `uvx --from ruff==0.15.22 ruff check --select F,RUF022 -- tests/test_precommit_pii.py`
- `git diff --check`
- independent spec review and final security review: approved

## Summary

- Updated `.githooks/pre-commit` to scan only added lines in the staged diff.
- Excluded diff headers and expected `pyproject.toml` metadata from false-positive detection.
- Preserved fail-closed behavior for diff failures, added PII, and staged binary content.
- Added regression tests for historical, added, removed, binary, header-prefixed, excluded-file, and failure cases.

## Architectural impact

- No changes to retrieval, consolidation, veracity, sync, Hermes, MCP, CLI, or benchmark methodology.
- No erosion of local-first guarantees.
- The change improves local maintainability by allowing valid `pyproject.toml` edits while retaining the PII guard.
- Keeping the guard fail-closed is the right call for the repository’s privacy posture.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -4,9 +4,18 @@ HOOK_FILE=".githooks/pre-commit"
 
 PII_PATTERNS="1641797+AxDSan@users.noreply.github.com"
 
-if git diff --cached --no-color --text -U0 --output-indicator-new='!' -- . ":(exclude)$HOOK_FILE" \
-  | grep -a '^!' \
-  | grep -a -F -q "$PII_PATTERNS"; then
+DIFF_OUTPUT=$(mktemp "${TMPDIR:-/tmp}/pre-commit.XXXXXX") || {
+  echo "ERROR: Unable to create temporary file for staged-diff PII check. Aborting commit." >&2
+  exit 1
+}
+trap 'rm -f "$DIFF_OUTPUT"' 0 HUP INT TERM
+
+if ! git diff --cached --no-color --text -U0 --output-indicator-new='!' -- . ":(exclude)$HOOK_FILE" > "$DIFF_OUTPUT"; then
+  echo "ERROR: Unable to inspect staged changes; git diff failed. Aborting commit." >&2
+  exit 1
+fi
+
+if grep -a '^!' "$DIFF_OUTPUT" | grep -a -F -q "$PII_PATTERNS"; then
   echo "ERROR: PII detected in staged files. Remove before committing."
   echo "Found pattern: $PII_PATTERNS"
   exit 1
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -4,13 +4,12 @@ HOOK_FILE=".githooks/pre-commit"
 
 PII_PATTERNS="1641797+AxDSan@users.noreply.github.com"
 
-staged=$(git diff --cached --name-only | grep -v "^$HOOK_FILE$" || true)
-if [ -n "$staged" ]; then
-  if echo "$staged" | xargs grep -l "$PII_PATTERNS" 2>/dev/null; then
-    echo "ERROR: PII detected in staged files. Remove before committing."
-    echo "Found pattern: $PII_PATTERNS"
-    exit 1
-  fi
+if git diff --cached --no-color --text -U0 --output-indicator-new='!' -- . ":(exclude)$HOOK_FILE" \
+  | grep -a '^!' \
+  | grep -a -F -q "$PII_PATTERNS"; then
+  echo "ERROR: PII detected in staged files. Remove before committing."
+  echo "Found pattern: $PII_PATTERNS"
+  exit 1
 fi
 
 # Also check git config
--- a/tests/test_precommit_pii.py
+++ b/tests/test_precommit_pii.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


AxDSan (approved):
Approving. I pulled the branch and ran it rather than reading it, and the hook logic is right.

## Verified

- All 8 tests pass locally on a clean config.
- The behavior change is exactly what #869 asked for: a commit that only *touches* `pyproject.toml` now passes, while a newly added occurrence is still rejected. I hit the old behavior twice this week and had to `--no-verify` both times, so this unblocks real work.
- Fail-closed on both `mktemp` failure and `git diff` failure, both covered.

## The part worth calling out

`--output-indicator-new='!'` instead of grepping `^+` is the detail that makes this correct rather than nearly correct. Grepping `^+` would match the `+++ b/file` header, and it would match an added line whose own content starts with `+`. Using an indicator that cannot collide with diff syntax removes the whole class, and `test_precommit_rejects_added_pii_with_diff_header_prefix` pins it. Same for `--text` plus the binary test.

## One non-blocking finding: the fixture is not hermetic

`_hook_repo` runs `git commit` in the temp repo while `pyproject.toml` containing the pattern is staged as an added file. If the developer's **global** git config sets `core.hooksPath` or `commit.gpgsign`, that fixture commit fails and every test in the file errors.

The `core.hooksPath` case is a nice trap: the temp repo has the hook copied into `.githooks/`, so a global `core.hooksPath=.githooks` makes the fixture commit run the very hook under test, against a diff where the pattern is a genuinely added line. It correctly rejects, and the fixture dies.

Reproduced:

```
$ GIT_CONFIG_GLOBAL=<commit.gpgsign=true + core.hooksPath=.githooks> pytest tests/test_precommit_pii.py -q
8 failed

AssertionError: ERROR: PII detected in staged files. Remove before committing.
  where 1 = CompletedProcess(args=['git', 'commit', '-qm', 'fixture'], returncode=1, ...)
```

Each setting alone also produces 8 failures. CI is green because runners have neither, so this only ever bites a contributor, which is the worst place for it to bite.

One line in `_run` fixes it, since every fixture command goes through there:

```python
        ["git", "-c", "core.hooksPath=/dev/null", "-c", "commit.gpgsign=false", *args],
```

Verified: 8 passed under the hostile global config, and 8 passed under a clean one.

Not blocking the merge. Your call whether to fold it in here or take it separately.

## Out of scope, noting so it is not forgotten

The second half of the hook is still dead code:

```sh
CURRENT_EMAIL=$(git config user.email)
if [ "$CURRENT_EMAIL" = "1641797+AxDSan@users.noreply.github.com" ]; then
  echo "INFO: ... Skipping check."
fi
exit 0
```

It prints an INFO line and exits 0 either way, and there is no check left for it to skip. Rightly untouched by a PR about the staged-diff check, but it should either become a real check or go.

CI was hanging on every PR until #867 landed a few minutes ago, which is why this one's test jobs died at the 50 minute cap. I have re-run them.

dplush:
@coderabbitai review

PR #877, opened 2026-08-26
test: avoid root plugin import shadowing under pytest

### LINKED ISSUE #868: Tests cannot be run from a local checkout: root __init__.py shadows the mnemosyne package under pytest
Found while reproducing #860.

`pytest` cannot import `mnemosyne.mcp_server` from a clean checkout:

```
E   ModuleNotFoundError: No module named 'mnemosyne.mcp_server'
```

while `python -c "import mnemosyne.mcp_server"` in the same shell succeeds.

## Cause

The repo root carries an `__init__.py`, the entry point for `hermes plugins install` compatibility. That makes the repository directory itself look like a package, so pytest's `prepend` import mode walks past the repo root and puts the **parent** directory on `sys.path`:

```
FILE  /home/aj/Documents/DevStuff/mnemosyne/__init__.py      <- the plugin stub
PATH  ['/home/aj/Documents/DevStuff/mnemosyne']
SYS0  ['/.../mnemosyne', '/home/aj/Documents/DevStuff', '/.../mnemosyne/tests', ...]
```

`import mnemosyne` then resolves to the repo root stub instead of `mnemosyne/`, and every submodule import fails.

The stub's own docstring names this exact hazard:

> an ambient sys.path entry for that parent dir can shadow this checkout, making `import mnemosyne` resolve to THIS plugin `__init__` (the stub) instead of the mnemosyne core subpackage

The `sys.path.insert` it does as a workaround runs too late to help, because by then `mnemosyne` is already bound to the stub.

## Impact

Local `pytest tests/` is mostly red for reasons unrelated to any change. With the file moved aside: **3351 passed, 36 skipped**. With it in place, large parts of the suite cannot import. CI does not hit this, so the cost falls entirely on contributors, and it is the kind of thing that makes someone conclude the suite is broken and stop.

Note that `tests/test_init_register_shadowing.py`, `tests/test_loader_registers_provider.py`, `tests/test_outer_package_version.py` and `tests/test_core_wheel_contents.py` genuinely depend on the root `__init__.py` being present, so it cannot simply be deleted.

## Options

1. `--import-mode=importlib` in `[tool.pytest.ini_options]`, which does not do the parent-directory walk. Cheapest, but changes import semantics suite-wide and wants its own verification pass.
2. A root `conftest.py` that pins `sys.path` before collection.
3. Move the plugin entry point out of the repository root, if `hermes plugins install` can be pointed at a subdirectory. Cleanest, largest blast radius, and it touches the Hermes install contract.

I have not picked one. Filing so it is visible rather than rediscovered.

dplush:
## Summary

Fixes #868 by running pytest in importlib mode. This prevents pytest's prepend import mode from binding the repository-root Hermes plugin stub as top-level `mnemosyne` during local checkout collection.

## Why this path

The root `__init__.py` must remain for Hermes plugin and memory-provider discovery. Importlib mode avoids the collection-time path mutation while exercising the same inner package contract that CI's editable install provides.

## Why not a root conftest.py path fix

A pytest-only `sys.path` mutation depends on collection order and can mask packaging/import-isolation mistakes.

## Non-goals

- No changes to the root plugin stub or Hermes discovery.
- No production `sys.path` mutation.
- No changes to existing root-stub/plugin tests.

## Reproduction and verification

Starting from a parent/plugin-layout path, pytest prepend mode can load the root stub as `mnemosyne`, after which `mnemosyne.mcp_server` cannot resolve. The new subprocess regression runs pytest with importlib mode and asserts the import resolves to the inner package.

- focused import, root-stub/provider, and MCP tests: `12 passed, 3 skipped`
- changed-file Ruff and `git diff --check`: passed
- `tests/test_plugins.py` has the same pre-existing logging-module pollution failure on base and candidate: `1 failed, 69 passed`

## Summary

- Configures pytest to use `importlib` mode.
- Adds a subprocess regression test for MCP imports in a Hermes-style plugin layout.
- Preserves the repository-root Hermes plugin stub and discovery contract.
- Makes no production `sys.path` changes.

## Architectural impact

- **Retrieval, consolidation, veracity, and sync:** No changes.
- **Privacy and local-first guarantees:** Preserved. The PR changes only local test collection and adds no network, telemetry, or production path logic.
- **Agent integration:** Preserves Hermes discovery and ensures `mnemosyne.mcp_server` resolves to the inner core package.
- **CLI and package contract:** Preserved for Hermes and editable-install workflows.
- **Maintainability:** This is the right call. A small, collection-order-independent configuration change fixes the package collision without moving or weakening the Hermes integration surface. Regression coverage protects the MCP import contract.

Focused tests passed with `12 passed, 3 skipped`. Changed-file Ruff checks and `git diff --check` passed. An existing logging-module pollution failure remains in `tests/test_plugins.py`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,7 @@ select = [
 ]  # NOTE: F covers F63, F7, F82 but NOT E9 — the E9 fatal gate lives in ci.yml only
 
 [tool.pytest.ini_options]
+addopts = "--import-mode=importlib"
 # No single test should take five minutes. The cap exists so a deadlock is
 # reported as one named failing test within minutes instead of stalling the
 # job until the workflow timeout, which tells us nothing about where it hung.
--- a/tests/test_hygiene.py
+++ b/tests/test_hygiene.py
@@ -1205,7 +1205,7 @@ def test_hygiene_suite_does_not_leak_config_into_subagent_provider(tmp_path, mon
     process: the autouse cleanup must discard that singleton before a subagent
     provider resolves its temporary data-directory configuration.
     """
-    from conftest import _close_cached_connections
+    from tests.conftest import _close_cached_connections
     from hermes_memory_provider import MnemosyneMemoryProvider
     from mnemosyne.core.config import MnemosyneConfig
 
--- a/tests/test_mcp_import_plugin_layout.py
+++ b/tests/test_mcp_import_plugin_layout.py
@@ -1,0 +1,60 @@
+"""Regression coverage for #868's parent-directory import shadowing."""
+
+from __future__ import annotations
+
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def test_mcp_server_resolves_to_inner_package_from_plugin_layout(tmp_path):
+    """Pytest collection must not make the plugin stub shadow the core package."""
+    plugin_parent = tmp_path / "plugins"
+    plugin_dir = plugin_parent / "mnemosyne"
+    plugin_dir.mkdir(parents=True)
+    (plugin_dir / "__init__.py").write_text((REPO_ROOT / "__init__.py").read_text())
+    (plugin_dir / "mnemosyne").symlink_to(REPO_ROOT / "mnemosyne", target_is_directory=True)
+    tests_dir = plugin_dir / "tests"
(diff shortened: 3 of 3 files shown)


dplush:
@AxDSan #877 is green on fbba7f1: full 3.10–3.13 CI, CodeRabbit, CLA, and independent review are all clean. It is a test-only fix for local pytest import shadowing; I am merging it now.

PR #880, opened 2026-08-26
feat: expose BEAM initialization vector status

### LINKED ISSUE #874: init_beam() reports a vector dimension mismatch only to the log, so callers cannot detect it
Surfaced by @dplush during the 4.0.0b1 canary (#849, P1 report), recorded there as a non-blocking UX observation. Filing it so it is not lost in a comment thread.

## What happens

`init_beam()` detects the mismatch correctly and handles it well (`mnemosyne/core/beam.py:819-824`):

```python
vec_dim_mismatch = False
if _SQLITE_VEC_AVAILABLE:
    existing_dim = _existing_vec_dim(conn)
    if existing_dim is not None and existing_dim != EMBEDDING_DIM:
        vec_dim_mismatch = True
        logger.error(_dim_mismatch_message(existing_dim, EMBEDDING_DIM))
```

The flag then correctly suppresses the vec table creation and the backfill (`beam.py:1131`, `beam.py:1240`), the existing vectors are left untouched, and the message is genuinely actionable.

Then `init_beam()` returns `None` and initialization reports success. The flag is local and never leaves the function. A caller sees no exception, no return value, and no exit status: the only trace is a log line. A script that opens a store, gets a degraded keyword-only recall path, and wants to notice cannot, short of parsing log output.

## What NOT to do

Raising is the obvious idea and it is wrong here. The guard exists precisely so that a mismatch is not treated as corruption. Its own message says so:

> Embedding dimension mismatch, NOT database corruption: your memories are intact, only the vector index is affected (recall falls back to keyword search until this is fixed).

Raising would take a perfectly usable store offline over a degraded index, and would invert a deliberate design decision. `_dim_mismatch_message` even has a comment explaining that the wording was chosen because "dimension mismatch" gets misread as "database corrupt" and people abandon the store instead of reindexing.

## Proposed shape

Additive, so no existing caller changes behavior:

1. `init_beam()` returns a small result rather than `None`. Every current caller ignores the return value, so returning something is backward compatible.
2. The value carries at least `vec_dim_mismatch`, and probably `existing_dim` and `configured_dim`, since a caller that wants to act needs the numbers the message already computes.
3. The CLI surface that initializes a store exits non-zero when the flag is set, so `&&` chains and CI scripts stop rather than continue against a keyword-only index.

Point 3 is the part that needs a decision: which surfaces should change exit status, and whether that is a breaking change for anyone scripting against them today. Worth settling in review before writing it.

`mnemosyne doctor` already reports the condition, so the reporting path is not the gap; the programmatic one is.

Related: #753 (recall crashing on a query-side dimension disagreement) is the same underlying mismatch seen from the read path.

dplush:
## Summary

Fixes #874 by exposing BEAM vector initialization status as an additive Python contract.

`init_beam()` now returns a frozen result with the mismatch flag, stored vector dimension, and configured dimension. `BeamMemory` and `Mnemosyne` retain the result for normal construction paths.

## Behavior preserved

A dimension mismatch remains a usable degraded state:

- no exception and no CLI/MCP exit-code change
- existing vector tables remain untouched
- incompatible table creation and backfill remain suppressed
- keyword fallback and existing recovery logging remain intact

## Non-goals

- no automatic reindex or table mutation
- no Doctor/diagnose policy change
- no general vector-health API

## Verification

- fresh no-`sqlite-vec` test environment: `4 passed, 5 skipped`
- sqlite-vec environment: `9 passed`
- independent final review: approved
- Ruff and `git diff --check`: passed

The issue’s note that `doctor` already detects stored vector-dimension mismatch was corrected separately: it currently reports configured dimension and general coverage only. A read-only Doctor status field or opt-in strict mode is follow-up scope.

## Summary

- `init_beam()` now returns an immutable `BeamInitResult`.
- The result reports mismatch status, stored dimensions, and configured dimension.
- `BeamMemory` and `Mnemosyne` expose the result through `init_result`.
- Mismatched vector indexes remain usable through keyword-only retrieval.
- Existing vector tables remain unchanged.
- Incompatible vector-table creation and backfill remain suppressed.
- Tests cover `sqlite-vec`, missing `sqlite-vec`, mixed dimensions, recovery, immutability, and propagation.

## Architecture and Integration

- The change affects BEAM vector initialization only.
- Working memory, episodic memory, consolidation, veracity, and sync behavior do not change.
- Retrieval remains degraded but usable when vector dimensions mismatch.
- Hermes, MCP, and CLI behavior remain unchanged, including exit codes.
- Agent integrations can inspect initialization status without parsing logs.
- No automatic reindexing or vector-table mutation is introduced.
- The additive result contract is the right call because it exposes degraded state without taking a usable store offline.

## Privacy and Local-First Design

- Privacy posture remains unchanged.
- Local-first guarantees remain unchanged.
- The change adds no remote service, telemetry, or external data path.
- Nothing in this PR erodes local-first design.

## Maintainability

- Immutable status data provides a stable programmatic contract.
- Dimension diagnostics cover recognized vector tables, including mixed stored dimensions.
- Existing recovery logging and degraded-mode behavior remain intact.
- Doctor policy and strict-mode behavior remain deferred.
- No benchmark methodology changes are included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Added
 
+- **BEAM initialization status is now available through the additive public Python `BeamInitResult`.** It reports the configured embedding dimension, any dimension mismatch, and immutable stored dimensions for each vector table; mismatches continue to avoid exceptions and automatic mutation.
 - **Multimodal memory: images, video and audio can become recallable memories (RFCs 0002, 0003, 0004).** `BeamMemory.remember_media(ref)` takes a reference to a piece of media, registers it, describes it through a configured modality provider, and writes the description back as an ordinary memory that hybrid recall already understands. Nothing about text recall changes.
 
   The stack is additive throughout. Two sidecar tables, `media_assets` and `media_moments`, are created `IF NOT EXISTS` by their own store when a bank is opened, so existing databases acquire them with no migration step and no change to any existing table. No new package dependency is introduced.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Added
 
-- **BEAM initialization status is now available through the additive public Python `BeamInitResult`.** It reports the configured embedding dimension, any dimension mismatch, and immutable stored dimensions for each vector table; mismatches continue to avoid exceptions and automatic mutation.
+- **BEAM initialization status is now available through the additive public Python `BeamInitResult`.** It reports the configured embedding dimension, any dimension mismatch, and immutable stored dimensions for each vector table.
 - **Multimodal memory: images, video and audio can become recallable memories (RFCs 0002, 0003, 0004).** `BeamMemory.remember_media(ref)` takes a reference to a piece of media, registers it, describes it through a configured modality provider, and writes the description back as an ordinary memory that hybrid recall already understands. Nothing about text recall changes.
 
   The stack is additive throughout. Two sidecar tables, `media_assets` and `media_moments`, are created `IF NOT EXISTS` by their own store when a bank is opened, so existing databases acquire them with no migration step and no change to any existing table. No new package dependency is introduced.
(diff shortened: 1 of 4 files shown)


dplush:
Thanks — the runtime failure is real, but it is unchanged from the PR base and #880 deliberately does not alter recall runtime. I reproduced it on both base and head and opened #881 for the narrow fix: skip episodic sqlite-vec when init reports a dimension mismatch, then continue through FTS/keyword fallback with one public-path regression. The requested all-strategies accuracy matrix is intentionally out of scope for this init-status PR.

dplush:
@AxDSan #880 is green on 7246905: full 3.10–3.13 CI, CodeRabbit, CLA, focused cross-capability tests, and independent reviews are clean. It adds the additive BEAM initialization status contract, including mixed legacy vector-index dimensions, without changing CLI/MCP behavior. The pre-existing public recall fallback defect is tracked separately in #881. I am merging now.

Issue #783, opened 2026-08-18
[BUG] test_mnemosyne_stats runs against the ambient data directory, so test_rapid_fire flakes on developer machines


AxDSan:
`tests/test_mnemosyne_stats.py` runs the stats CLI as a subprocess against whatever data directory the environment happens to resolve at that moment. It never points at a temp bank, so on a developer machine it reads and reports on the real Mnemosyne database.

```python
def _expected_data_dir() -> Path:
    if data_dir := os.environ.get("MNEMOSYNE_DATA_DIR"):
        return Path(data_dir)
    if hermes_home := os.environ.get("HERMES_HOME"):
        return Path(hermes_home).expanduser() / "mnemosyne" / "data"
    return Path.home() / ".hermes" / "mnemosyne" / "data"

DB_PATH  = _expected_data_dir() / "mnemosyne.db"
SNAP_DIR = Path.home() / ".hermes" / "mnemosyne" / "stats"
WIKI_PATH = Path.home() / "wiki"
```

`run()` shells out with no environment override:

```python
cmd = f"cd {SCRIPT.parent} && python3 {SCRIPT} {args}"
result = subprocess.run(cmd, shell=True, capture_output=True, text=True, timeout=30)
```

`conftest.py` resets the singletons and the config cache but does not set `MNEMOSYNE_DATA_DIR` to a temporary directory, and its own comment notes that a later provider test swaps that variable to a temporary bank. So the directory these tests see depends on ambient environment plus whatever preceding tests left behind.

## What I observed

`test_rapid_fire` calls `run("--compact")` ten times and asserts every exit code is zero.

* Full suite on `main` with `MNEMOSYNE_NO_EMBEDDINGS=1`: failed once with `AssertionError: Run 2 failed`, out of two runs. The second run of the identical tree was fully green (2819 passed).
* Standalone, five consecutive runs: passed every time.

CI stays green because a fresh runner has no live database and no concurrent writer, which is exactly why this has not been caught there.

## Two plausible mechanisms, and I cannot yet say which fired

1. **Concurrent writers on a real database.** On my machine the resolved path is a 31 MB `mnemosyne.db` with `-wal` and `-shm` present and seven live `mnemosyne` processes holding connections. Ten rapid CLI invocations against a database being written concurrently can hit lock contention or a checkpoint mid-read, and the subprocess exits nonzero.
[...]

Issue #753, opened 2026-08-14
recall() crashes the process when the query embedding's dimension disagrees with the store's vec0 tables


paoloantinori:
## What happens

When a process resolves an `EMBEDDING_DIM` different from the dimension the store's `vec0` tables were created at, **the write path degrades but the recall path crashes**:

- **Write side (graceful):** the mismatched vector insert is caught, logged (`vec_working upsert failed for '<id>' (OperationalError): Dimension mismatch for inserted vector...`), and dropped. `remember()` returns normally.
- **Query side (crash):** `_vec_search` (`mnemosyne/core/beam.py:2681`, the episodic KNN over `vec_episodes`) executes its `MATCH` with no exception handling. `sqlite3.OperationalError: Dimension mismatch for query vector for the "embedding" column. Expected 1024 dimensions but received 384.` propagates through `BeamMemory.recall()` (`beam.py:6419`) and out of `recall()`, exiting the calling process non-zero.

The working-memory KNN `_wm_vec_search_sqlite` (`beam.py:~3086`) already wraps the same class of query in `try/except -> return []`, so the episodic voice is the one unprotected path.

## Reproduction (verified)

Fresh process, custom embedding endpoint serving 384-dim vectors, `MNEMOSYNE_EMBEDDING_DIM=1024`:

```
MNEMOSYNE_EMBEDDING_MODEL=mixedbread-ai/mxbai-embed-large-v1 \
MNEMOSYNE_EMBEDDING_DIM=1024 \
MNEMOSYNE_EMBEDDING_API_URL=<endpoint-serving-384> \
python -c "from mnemosyne.core import beam; from mnemosyne.core.memory import remember, recall; beam.init_beam(); remember('probe'); recall('probe')"
```

```
Traceback (most recent call last):
  File ".../beam.py", line 6419, in recall
    vec_rows = _vec_search(self.conn, emb_result.tolist(), k=max(top_k * 3, 20))
  File ".../beam.py", line 2708, in _vec_search
    rows = conn.execute(...)
sqlite3.OperationalError: Dimension mismatch for query vector for the "embedding" column. Expected 1024 dimensions but received 384.
```

Exit code 1. Found while writing the negative-control e2e test on #748.

## When this bites

[...]

PR #882, opened 2026-08-26
fix: fall back from mismatched episodic vec recall

### LINKED ISSUE #881: recall should fall back to keyword search on legacy sqlite-vec dimension mismatch
## Problem

A legacy sqlite-vec database can contain an episodic vector index whose stored dimension differs from the configured embedding dimension. For example, a legacy `vec_episodes` table at 768 opened with `EMBEDDING_DIM=384`.

`init_beam()` now detects and reports this degraded state (#874 / #880), but the default linear public recall path still calls episodic sqlite-vec search when the table is probeable. A 384-d query then raises `sqlite3.OperationalError: Dimension mismatch` instead of returning the existing keyword/FTS fallback.

## Evidence

Reproduced on both current `main` base `ff23ef5570984d3f8e9f8098a528aedaa5547c40` and #880 head `9b4789c810f646ebe1b522de696ce8c6b710cf97`. This is therefore pre-existing and not caused by #880.

`Mnemosyne.recall`, CLI recall, and MCP recall inherit the linear path. The OpenWebUI wrapper catches the exception only to return an error payload.

## Proposed narrow contract

When initialization has established `vec_dim_mismatch`:

1. skip episodic sqlite-vec search;
2. continue through the normal FTS/keyword path;
3. return the matching memory without surfacing `sqlite3.OperationalError`.

Add one focused public `BeamMemory.recall()` / `Mnemosyne.recall()` regression using a 768→384 legacy table and a known FTS-matched episodic memory.

## Non-goals

- no strategy-wide accuracy matrix;
- no rewrite of polyphonic or working-memory fallback semantics;
- no automatic reindex/table mutation;
- no CLI exit-code change.

dplush:
## Summary

Fixes #881.

When BEAM initialization has identified a legacy episodic sqlite-vec index with an incompatible embedding dimension, linear public recall now skips the incompatible sqlite-vec episodic search and continues through the existing fallback path.

## Scope

- preserves in-memory vector and FTS/keyword fallback behavior
- covers both `BeamMemory.recall()` and `Mnemosyne.recall()` with a populated 768→384 legacy store
- verifies the normal matching-dimension sqlite-vec path remains active
- keeps polyphonic, working-memory, CLI/MCP, reindex, and exit-code behavior out of scope

## Validation

- `[test]`: 113 passed, 23 skipped
- `[test,embeddings]`: 136 passed
- Ruff and `git diff --check` passed
- independent current-head review approved

## Summary

- Public linear recall skips episodic sqlite-vec search when `vec_dim_mismatch` is detected.
- Recall continues through scoped in-memory vector and FTS/keyword fallback paths.
- Matching-dimension sqlite-vec recall remains enabled.
- Scoped candidate selection prevents unrelated rows from exhausting bounded searches.
- Malformed query vectors produce no vector candidates instead of raising an error.
- Enhanced-recall cache version 6 prevents stale results from the previous candidate-selection algorithm.
- Regression tests cover legacy 768→384 stores, matching dimensions, query validation, FTS fallback, and filtered matches among 10,000 unrelated rows.

## Architecture and Integration Impact

- **Memory tiers:** The change affects episodic BEAM recall. Working memory and other tiers remain unchanged.
- **Retrieval:** Compatible episodic indexes use sqlite-vec. Incompatible indexes use scoped in-memory vector and FTS/keyword fallback paths.
- **Consolidation and veracity:** No changes.
- **Privacy and local-first design:** No external services, data movement, telemetry, automatic reindexing, or table mutation were added. Existing episodic data remains available locally.
- **Hermes, MCP, and CLI:** Shared linear recall avoids exposing `sqlite3.OperationalError` through these integration surfaces. Public APIs and CLI exit-code behavior remain unchanged.
- **Sync layer:** No changes.
- **Benchmark methodology:** No benchmark methodology or strategy-wide accuracy target changed.
- **Maintainability:** Initialization state now controls sqlite-vec safety. Scoped helpers preserve candidate boundaries across retrieval paths. Focused tests and a changelog entry document the compatibility contract.

This is the right call for legacy compatibility. It preserves local episodic recall without weakening local-first guarantees or changing unrelated memory architecture.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
+- **Public linear recall skips incompatible legacy sqlite-vec episodic indexes after an embedding-dimension mismatch is detected (#881).** Existing keyword/FTS fallback is used instead of surfacing a SQLite `OperationalError`.
 - **Optional `embeddings` and `all` installs cap `onnxruntime` below 1.29.** This avoids `blkid` stderr on minimal Linux/aarch64 systems.
 - **CI stopped being able to verify anything, because an unpinned dependency changed ASGI behaviour (#860).** `tests/test_mcp_streamable_http.py` drives the authenticated SSE GET by hand through the TestClient portal, and its `receive()` never delivered the initial `http.request` message. That violates the ASGI contract, but mcp 2.0.0 answered without waiting for it, so the driver passed. mcp 2.1.0 reads the request body to enforce `max_request_body_size` (SDK #3336), so the handler now blocks before `http.response.start`, the test's wait fails, and `TestClient.__exit__` then blocks forever draining a task group that still holds the wedged ASGI task. The job ran to its six-hour ceiling and reported nothing.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ and this project adheres to [SemVer](https://semver.org/) starting from v3.1.2.
 
 ### Fixed
 
-- **Public linear recall skips incompatible legacy sqlite-vec episodic indexes after an embedding-dimension mismatch is detected (#881).** Existing keyword/FTS fallback is used instead of surfacing a SQLite `OperationalError`.
+- **Public linear recall skips incompatible legacy sqlite-vec episodic indexes after an embedding-dimension mismatch is detected (#881).** Existing keyword/FTS fallback is used instead of surfacing a SQLite `OperationalError`. Compatible `vec_episodes` indexes stay on when their stored dimension matches the query vector, even if another vector table or the process-wide `EMBEDDING_DIM` constant differs.
 - **Optional `embeddings` and `all` installs cap `onnxruntime` below 1.29.** This avoids `blkid` stderr on minimal Linux/aarch64 systems.
 - **CI stopped being able to verify anything, because an unpinned dependency changed ASGI behaviour (#860).** `tests/test_mcp_streamable_http.py` drives the authenticated SSE GET by hand through the TestClient portal, and its `receive()` never delivered the initial `http.request` message. That violates the ASGI contract, but mcp 2.0.0 answered without waiting for it, so the driver passed. mcp 2.1.0 reads the request body to enforce `max_request_body_size` (SDK #3336), so the handler now blocks before `http.response.start`, the test's wait fails, and `TestClient.__exit__` then blocks forever draining a task group that still holds the wedged ASGI task. The job ran to its six-hour ceiling and reported nothing.
 
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -7538,7 +7538,7 @@ class BeamMemory:
     # cached under an older digest are not reused. Part of the hashed payload;
     # the opaque key keeps the "v2:" prefix because QueryCache's opaque-path
     # recognition (_OPAQUE_V2_KEY_RE) keys off that prefix.
-    _ENHANCED_RECALL_CACHE_VERSION = 5
+    _ENHANCED_RECALL_CACHE_VERSION = 6
 
     def _enhanced_recall_cache_key(
         self,
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -6929,7 +6929,8 @@ class BeamMemory:
         if embeddings_available:
             emb_result = _get_query_embedding()
             if emb_result is not None:
-                if _vec_available(self.conn) and not self.init_result.vec_dim_mismatch:
+                episodic_vec_dim = dict(_existing_vec_dims(self.conn)).get("vec_episodes")
+                if _vec_available(self.conn) and episodic_vec_dim == EMBEDDING_DIM:
                     vec_rows = _vec_search(self.conn, emb_result.tolist(), k=max(top_k * 3, 20))
                 else:
                     # Fallback: in-memory cosine similarity search
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -2441,6 +2441,15 @@ def _in_memory_vec_search(
     conn: sqlite3.Connection,
     query_embedding: np.ndarray,
     k: int = 20,
+) -> List[Dict]:
+    """Fallback vector search retaining the established three-argument contract."""
+    return _in_memory_vec_search_scoped(conn, query_embedding, k=k)
+
+
+def _in_memory_vec_search_scoped(
+    conn: sqlite3.Connection,
+    query_embedding: np.ndarray,
+    k: int = 20,
     *,
     where_clause: str = "(1=1)",
     where_params: Tuple[Any, ...] = (),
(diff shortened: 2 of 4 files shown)


dplush:
@AxDSan could you please review #882 on 75add605? It fixes the CodeRabbit-reported fallback vector candidate-scope issue. Current CI, CodeRabbit, and independent reviews are clean; the regression covers 10,000 out-of-scope rows without lifting the candidate cap.

dplush:
@AxDSan heads-up before merge: #882 changes the BEAM recall pipeline to scope episodic sqlite-vec, FTS, JSON, CJK, and Cyrillic candidates before bounded selection. It also guards legacy vector-dimension mismatch and adds public regressions for session-scoped candidate exhaustion.

Current head `0edc64d` is clean: CI, CLA, CodeRabbit, focused local matrix, and fresh Sol/Claude reviews are green. The Unreleased changelog entry is included. Could you please give the final architecture decision for merge?

dplush:
@AxDSan Trade-offs for the architecture decision:

**Benefits**
- Prevents foreign-session, channel, source, date, invalid, or superseded rows from exhausting bounded episodic vec/FTS/LIKE candidate windows before scope is applied.
- Preserves compatible episodic sqlite-vec recall even if another vec table has a different dimension.
- Degrades incompatible legacy episodic vec indexes safely to lexical fallback rather than raising SQLite errors.
- Rejects malformed non-1D query vector shapes before they reach sqlite-vec.
- Invalidates stale enhanced-recall cache entries and adds public adversarial regressions for vec, FTS, CJK, and Cyrillic paths.

**Costs / follow-ups**
- Scoped sqlite-vec uses bounded widening, so very large stores with narrow scopes can require additional KNN scans.
- Earlier joins and filters add some FTS/LIKE query cost.
- The alias-qualified episodic scope SQL duplicates part of the generic scope-helper logic. It is semantically checked, but alias-aware helper reuse would reduce future drift risk.
- The scoped vec selector currently degrades to no vec candidates on SQLite errors; debug logging and a production-scale latency benchmark would be sensible follow-ups.

My assessment: the correctness and isolation gain outweigh the bounded performance and maintainability costs; no known merge blocker remains.

dplush:
@AxDSan important update before a merge decision: the scoped episodic sqlite-vec widening path has a verified sqlite-vec 0.1.9 correctness limit on our supported runtime. `k=4096` succeeds, while `k=4097` and `k=5001` raise `OperationalError` because the KNN limit is 4096. Our helper catches that error and returns no vec candidates.

In a synthetic adversarial scope test, scoped FTS still returns the in-scope target at 500, 5k, and 20k foreign rows. Scoped vec returns it at 501 total rows, but fails at 5,001 and 20,001 once widening exceeds the sqlite-vec cap. This is a correctness boundary, not just a latency trade-off.

I recommend pausing merge and choosing the large/narrow-scope contract explicitly: either a correct fallback once the KNN cap prevents sufficient eligible candidates, or a different scoped candidate strategy. Simply capping widening at 4096 would leave valid in-scope rows unreachable.

AxDSan (approved):
Approving. Verified locally rather than read, and this is the right shape.

## What I checked

- `tests/test_vec_dim_guard.py` + `tests/test_enhanced_recall_cache.py`: **70 passed**.
- Wider sweep, `-k "recall or beam or vec"`: **710 passed, 1 failed**. The failure is `test_mnemosyne_stats.py::test_recall_distribution_sums`, and it fails **identically on `main`** with the same selection (691 passed, same 1 failed). It passes in isolation on both. That is #783, the ambient-data-dir order dependence, with the fix pending in #787. Not from this PR.
- The f-string SQL. `_in_memory_vec_search_scoped` and `_episodic_vec_search_scoped` interpolate `where_clause`/`where_sql`, so I traced `em_where`'s construction: every clause is a literal with `?` placeholders and every user-derived value goes into `em_params`. Nothing user-derived reaches the interpolation, the docstring's claim holds, and it mirrors the `wm_where` pattern `_wm_vec_search` already uses. No objection.

## The design call is right

`init_result` with `vec_dim_mismatch` and `stored_dims` is exactly the additive return value #874 argued for, and doing it as a returned result rather than an exception preserves the guard's whole point: a dimension mismatch is a degraded index, not a corrupt store, and it must not take a usable store offline. Good that this went the additive way.

Two things I particularly like:

**`_existing_vec_dims` plural.** The old singular returned the first dimension it found across three tables, so a store with `vec_episodes` at 384 and `vec_working` at 768 reported one number and the other table was invisible. `stored_dims` naming each table is what makes `test_public_linear_recall_uses_compatible_episodic_vec_despite_mixed_indexes` expressible at all, and that test is the interesting one: a mismatch in *some* index no longer disables a compatible episodic index. That is a real behavior improvement hiding inside a fallback fix.

**Widening `k` before the relational predicate.** sqlite-vec applies KNN before the WHERE, so scoping without widening would have quietly starved a session's candidates. Matching what working-memory recall already does keeps one rule in the codebase instead of two.

## Relationship to the neighbouring issues, for whoever reads this later

- **#881** is what this fixes.
- **#874** is superseded; its points 1 and 2 are implemented here. Already closed.
- **#753** (query-side dimension disagreement crashing recall) is adjacent and still open. This makes the *episodic sqlite-vec* path fall back instead of using an incompatible index, which removes one route into that crash, but #753 describes `_vec_search` raising and this PR scopes working-memory, polyphonic, CLI/MCP and reindex out. Worth a line on #753 saying which of its paths remain after this lands, so it does not get closed as fixed when it is only narrowed.

Scoping polyphonic, CLI and exit-code behavior out was the right call for reviewability. The exit-status half of #874 is still unowned.

dplush:
@AxDSan heads-up before we push the final 4096-cap follow-up: the current-head gate found a second-order recall-contract decision in the partial sqlite-vec fallback.

When scoped sqlite-vec returns fewer candidates than requested, we must supplement with the scoped JSON vector search to avoid losing semantic in-scope results. But sqlite-vec exposes L2 distances while JSON uses `1 - cosine`; sorting raw values together can reorder results incorrectly.

The local candidate keeps the approved widening/scoping behavior, caps KNN at 4096, and uses exact cosine scores from the scoped JSON fallback when it is available. Native-only vec0 candidates without a corresponding JSON embedding are not given an invented cross-source dense score in that mixed path; they remain available to independent lexical/FTS signals. A regression proves partial native success plus fallback ranks identically to the JSON-only baseline.

This is a BEAM recall-ranking trade-off beyond the earlier approval. Is that conservative mixed-source policy the intended contract, or do you want a different strategy before we push?

dplush:
Following Paolo’s note on #754: I suggest we do not merge #882 ahead of #754. Once #754 lands, #882 can rebase and retain its scoped-candidate work while adopting the same strict episodic boundary: confirmed dimension mismatch means fallback plus guidance; any other SQLite error must propagate unchanged. @AxDSan, does that sequencing and contract look right?

dplush:
Closing as superseded by #754.

On current main (`8078a55`), the reported 768→384 mismatch is covered through both `BeamMemory.recall()` and `Mnemosyne.recall()`: it falls back to FTS/keyword results, a matching dimension retains the native vec path, and unrelated SQLite errors propagate unchanged. The focused public-path reproduction plus the #754 boundary suite passed (17 checks total).

The remaining broad scoped-candidate work is a separate design problem and needs a new explicit contract for the >4096 limit, native/JSON mixed ranking, and cache invalidation before it is proposed again.

PR #885, opened 2026-08-29
fix: validate wrapper --python environment

### LINKED ISSUE #883: Wrapper install checks mnemosyne core in the CLI interpreter, not the one --python selects
Raised by CodeRabbit on #864. Verified against the code and it holds, but it is pre-existing rather than introduced there, so it is filed here instead of blocking that PR.

## What happens

`_install_impl` runs the core-library preflight unconditionally, before any mode branching (`integrations/hermes/src/mnemosyne_hermes/install.py:2199-2207`):

```python
    # Check core library first (installer's own Python)
    core_ok = check_mnemosyne_core()
    if not core_ok:
        print(
            "  mnemosyne-memory NOT found in this Python. Install it first:\n"
            "    pip install mnemosyne-hermes[all]",
            file=sys.stderr,
        )
        return 1
```

`check_mnemosyne_core()` inspects the interpreter the CLI is running under. The very next block acknowledges that wrapper mode validates a different one:

```python
    # Symlink installs need Hermes' own Python to contain the package. Wrapper
    # installs validate the explicitly selected interpreter in install_plugin().
```

So `mnemosyne-hermes install --mode wrapper --python /path/to/other/venv/bin/python` returns 1 when the *CLI* interpreter lacks `mnemosyne-memory`, even though the interpreter the user explicitly selected has it and is the one that will actually import the package. The check that matters, `_validated_wrapper_environment()`, never runs.

That contradicts the rule the rest of the installer follows, established in #618/#620 and restated in #736: an explicit `--python` is authoritative.

## Why it matters more now

#864 makes wrapper the default mode on native Windows. Before it, a Windows user reaching wrapper mode had asked for it; after it, wrapper is where they land by default. The ordering bug is unchanged, but the number of people who can hit it goes up.

## Two ways to fix

1. Run the core preflight against the interpreter the mode will actually use, so wrapper mode checks the `--python` selection.
2. Skip the preflight for wrapper mode entirely and let `_validated_wrapper_environment()` own it, since it already performs the equivalent check on the right interpreter.

Option 2 is smaller and removes a duplicate check rather than teaching it a second code path, but it changes which error message a user sees on the failure, so it wants a look at both messages before choosing.

Non-wrapper modes should keep their current behavior either way.

dplush:
Closes #883

Wrapper installs with an explicit `--python` now validate Mnemosyne core in that selected interpreter instead of preflighting the CLI interpreter. The probe is isolated from the installer checkout and fails closed when `mnemosyne.core.beam` is unavailable.

Wrapper installs without `--python` and non-wrapper modes retain the existing CLI-interpreter preflight.

Tests: `uv run --no-sync --with pytest --with-editable . --with-editable integrations/hermes pytest integrations/hermes/tests/test_install_hermes.py integrations/hermes/tests/test_find_hermes_python.py integrations/hermes/tests/test_install_status.py integrations/hermes/tests/test_wrapper_import_timeout.py integrations/hermes/tests/test_windows_default_install_mode.py integrations/hermes/tests/test_windows_symlink_privilege.py -q` (176 passed).

No migration or data changes.

## Summary

- Validates `mnemosyne.core.beam` in the interpreter selected by explicit wrapper `--python`.
- Rejects blank and whitespace-only explicit `--python` values without fallback.
- Preserves existing preflight behavior for omitted `--python` and non-wrapper modes.
- Isolates wrapper validation from the installer checkout and sanitizes its environment.
- Adds regression coverage for interpreter discovery, relative paths, dry-run behavior, symlink boundaries, and missing core dependencies.

## Architectural impact

- No changes to the core memory data model.
- Hermes now validates the interpreter that performs the wrapper import. This is the right call for explicit interpreter selection.
- MCP and CLI integration surfaces have no functional changes beyond corrected wrapper installation behavior.
- The change preserves local-first behavior. It adds no network dependency, telemetry, migration, or data change.
- Isolated validation reduces accidental imports and improves long-term maintainability.

### DIFF
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -2400,7 +2400,10 @@ def main(argv: list[str] | None = None) -> int:
                 print(f"  Skill state: {skill.status}")
                 print(f"  Skill action: {skill_plan.message}")
                 if getattr(args, "mode", "symlink") == "wrapper":
-                    wrapper_python = Path(getattr(args, "python", None) or sys.executable).expanduser()
+                    # Match run_install(): wrapper metadata and validation use
+                    # the discovered Hermes runtime unless --python selected one.
+                    # Preserve a venv's python symlink; absolute() is lexical.
+                    wrapper_python = (hermes_python or Path(sys.executable)).absolute()
                     print(f"  Wrapper Python: {wrapper_python}")
                     if wrapper_python.is_file():
                         print(
--- a/integrations/hermes/src/mnemosyne_hermes/install.py
+++ b/integrations/hermes/src/mnemosyne_hermes/install.py
@@ -1551,7 +1551,16 @@ def _validated_wrapper_environment(
     import_timeout: float = DEFAULT_WRAPPER_IMPORT_TIMEOUT,
 ) -> tuple[Path, Path]:
     """Validate the selected wrapper runtime before touching an installed plugin."""
-    wrapper_python = Path(python).expanduser() if python else Path(sys.executable)
+    if python is None:
+        wrapper_python = Path(sys.executable)
+    else:
+        selected = str(python)
+        if not selected.strip():
+            raise ValueError(
+                "--python was given an empty value. Pass the path to Hermes' "
+                "interpreter, or omit --python to let the installer find it."
+            )
+        wrapper_python = Path(selected).expanduser()
     if not wrapper_python.is_file():
         raise FileNotFoundError(f"Python interpreter not found: {wrapper_python}")
     import_timeout = _validated_import_timeout(import_timeout)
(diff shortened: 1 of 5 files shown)


dplush:
Addressed the explicit blank `--python` case in 3636b0d. Only `None` now uses the wrapper default; empty and whitespace-only explicit values fail before validation or plugin creation, so they cannot fall back to the installer interpreter. Added CLI regressions for both forms. Focused matrix: 178 passed; Ruff and `git diff --check` clean.

dplush:
Addressed the current wrapper findings in a49e05a:

- explicit blank or whitespace `--python` values now fail before any preflight, discovery, or install work in both modes;
- wrapper installs and dry-runs fail closed when Hermes interpreter discovery finds nothing, with no fallback to the installer Python;
- omitted `--python` otherwise discovers, validates, and records Hermes' interpreter in both install and dry-run;
- explicit relative interpreter paths are made lexically absolute without resolving venv symlinks;
- symlink preflight/bootstrap behavior remains unchanged.

Added public-path and direct-boundary regressions for blank values, discovery failures, relative paths, dry-run, and symlink boundaries. Focused matrix: 188 passed; CI lint and `git diff --check` clean; independent review approved.

PR #896, opened 2026-09-02
fix(beam): recall Korean queries by stripping particles and keeping 2-syllable tokens

ZohnNasse:
<h2 dir="ltr">Summary</h2>

<p dir="ltr">Korean queries recall at R@5 71.4% on a 60-memory / 49-query benchmark. Three
defects on the query side drop Korean terms before or after FTS5 sees them, and
one of them means a row can be returned as the top FTS candidate and then be
discarded by the lexical admission gate, so <code>recall()</code> hands back nothing.</p>

<p dir="ltr">Fixing all three raises candidate R@5 to 100.0% and takes the LIKE fallback from
firing on 23/49 queries to 0/49.</p>

<p dir="ltr">The stored index is untouched. No migration. Non-Hangul queries take an
unchanged path — the leading-hyphen guard and the <code>_symbolic_code_tokens</code>
exclusion added in 4.0.0b1 both still run first, so <code>rm -rf 로그</code> still yields
<code>['로그*', '"rf"']</code> and <code>C++20 표준</code> still yields <code>['표준*']</code>.</p>

<p dir="ltr">Measured against 3.15.1 and rebased onto main at 98fc3b6 (4.0.0b1); the
thresholds in the test suite are asserted on main.</p>

<h2 dir="ltr">Results</h2>

<p dir="ltr">Two numbers, not one. <code>_fts_search()</code> is candidate generation;
<code>BeamMemory.recall()</code> applies a lexical admission gate on top of it. Quoting the
candidate figure as the recall figure is precisely how the defect reported in
review went unnoticed, so both are stated and both are frozen in the suite.</p>

<p dir="ltr"><strong>Candidate generation</strong> (<code>_fts_search</code>, 60-memory corpus):</p>

<div dir="ltr">
  | R@1 | R@5 | MRR | LIKE fallback
-- | -- | -- | -- | --
before | 63.3% | 71.4% | 0.667 | 23/49
after | 87.8% | 100.0% | 0.933 | 0/49

</div>

<p dir="ltr">Aggregate candidate growth of 1.03× hides the real behaviour — the expansion is
local, not average. Three queries blow up roughly 20×:</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">8080은 컨텍스트를 얼마로 잡았나    3 → 63 candidates
8082에 올린 모델 양자화가 뭐야     2 → 61
8082에 물려 있는 역할이 몇 개야    3 → 62</code></pre></div></div>

<p dir="ltr">The R@1 regression is concentrated in exactly those queries.</p>

<p dir="ltr">Admitting digit-only tokens (port numbers 8080/8082) produced results identical
to the patch in every metric, so it is not included.</p>

<h2 dir="ltr">Limitations</h2>

<p dir="ltr"><strong>The admission gate, not normalization, is what separates 49/49 candidates from
35/49 episodic recall.</strong> In every dropped query only the topic noun is shared;
the rest of the query is an interrogative (누가, 뭐야) or a conjugated predicate
whose document counterpart is a synonym (수정할/고칠, 버렸더라/폐기했다). The
gate is a fraction of query tokens, calibrated on English where query tokens are
content words appearing verbatim in the document. Korean queries have a far lower
reachable ceiling, so applying that fraction unchanged is a category error — but
correcting it needs morphological analysis, which this PR deliberately does not
introduce. The number is frozen in the suite so a follow-up starts from a
baseline rather than an impression.</p>

<p dir="ltr"><strong>Prefix matching is negation-blind</strong>: <code>보관</code> matches <code>보관하지 않았다</code>. This is
pre-existing and language-independent. Measured on the unpatched code, a positive
and a negated document both score 0.25 for a Korean query and 0.133 for the
English control; the change lifts both sides symmetrically rather than creating a
preference for the wrong polarity.</p>

<p dir="ltr"><strong><code>_strip_ko_josa</code> is suffix trimming, not morphology.</strong> Single-syllable
particles (나 / 야 / 여) are also verb endings, so <code>갱신하나</code> trims to <code>갱신하</code>.
Because the result is used as a prefix term the over-trimmed stem still matches
the intended documents, and the benchmark shows no recall cost. Frozen in
<code>test_single_syllable_particles_can_over_trim</code> so it reads as a decision rather
than an accident.</p>

<h2 dir="ltr">Tests</h2>

<p dir="ltr"><code>tests/test_korean_fts.py</code> — self-contained, 20 tests, no network, no fixtures
beyond a temp-file corpus of the 60 benchmark memories.</p>

<p dir="ltr">Candidate generation: prefix-term generation, two-syllable survival, non-Hangul
output being unchanged, suffix-trimming semantics and its idempotence, the
over-trimming limitation, R@1 43/49, R@5 49/49, all particle-inflected queries
recalling, the fallback firing 0 times, and — by replacing <code>socket.socket</code> with a
raising stub — that the search path makes no network calls.</p>

<p dir="ltr">Public path: <code>recall()</code> measured directly on both the episodic and working
layers, the reported inflected queries asserted at rank 1, and the invariant that
no query may be a top FTS candidate and then be dropped by the gate. Working-layer
R@1 is asserted as a floor rather than an exact value: the 60 rows are written in
a burst, so whether two share a timestamp varies per run and the recency tiebreak
then reorders equally-scored rows.</p>

<p dir="ltr">The quoting test asserts a property rather than a hardcoded term list. Which
additional components a query expands into is the hyphen splitter's business and
changes as the tokenizer evolves; the test asserts that the structured token is
quoted before the wildcard and runs MATCH with the generated terms, so a syntax
regression still fails.</p>

<p dir="ltr">The corpus fixture is function-scoped because conftest's autouse
<code>_reset_thread_local_connections</code> closes every cached connection around each
test; a module-scoped connection is already dead by the second test. This also
means measurements taken outside pytest are wrong — an ambient cached connection
gets reused and the numbers are unrelated to the corpus under test.</p>

<p dir="ltr"></p>


mnemosyne/core/beam.py
```diff
@@ -3012,7 +3089,7 @@ def _fts_query_terms(query: str) -> List[str]:
     terms: List[str] = []
     seen: Set[str] = set()
     symbolic = set(_symbolic_code_tokens(query))
-    for term in _expanded_query_tokens(_recall_tokens(query)):
+    for term in _expanded_query_tokens(_ko_relaxed_recall_tokens(query)):
```
dplush: `_fts_query_terms()` now admits two-syllable Hangul and strips particles for FTS candidates, but final `BeamMemory.recall()` still builds `query_words` with `_recall_tokens()`, retaining the >=3-character gate and unnormalized inflections. Reproduced on this head from an immutable Git archive: for `honcho가 한국어 검색을 못 하는 이유`, `_fts_search()` returns the target as its first candidate, but `recall()` returns `[]` because the target's lexical relevance is 1/3 = 0.333 below the 0.5 admission threshold. The new suite calls `_fts_search()` directly, so its 48/49 result does not prove the public recall path. Please make candidate generation and final admission use a consistent Korean normalization contract, then cover both episodic and working `BeamMemory.recall()` with inflected targets and distractors.

dplush (changes requested):
The FTS candidate benchmark is green, but the user-facing recall contract still fails. Please align the final lexical admission path with the Hangul normalization and add public-path regressions before merge.

mnemosyne/core/beam.py
ZohnNasse: Thanks for catching this — you were right, and the repro was exact.

**Root cause.** The normalization was applied on only one side of a two-stage
path. `_fts_query_terms()` stripped Korean particles when building candidates,
but `BeamMemory.recall()` scored the returned rows with raw `_recall_tokens()`.
So FTS did its job: for `honcho가 한국어 검색을 못 하는 이유` it returned the
correct row as its rank-1 candidate, and then the lexical admission gate scored
that same row against unnormalized tokens and dropped it. `_fts_search()` looked
healthy in every test I had, because every test I had called `_fts_search()`.

**Fix, in three parts.**

1. Normalization moved into `_recall_tokens()` itself, so candidate generation
   and admission share one contract. `_ko_relaxed_recall_tokens()` is gone — it
   existed only to widen one side in isolation, which is how the split happened.
   `_strip_ko_josa()` now trims to a fixed point, because the query token and the
   inflected document token are normalized by separate calls and a single pass is
   not idempotent (`바나나` → `바나`, but `바나나를` → `바나나`).

2. Unifying the tokenizer was necessary but not sufficient: the two layers still
   used different *matching* semantics. FTS matches by prefix (`"보관"*` hits
   `보관한다`) while `_lexical_relevance()` intersected token sets exactly. The
   gate now prefix-matches Hangul tokens too.

3. Fixing (1) exposed a pre-existing bug that the old code had masked. unicode61
   splits Hangul on whitespace and does not care whether the syllable before the
   particle is Korean — `ModelForge가` is a single index token, so the exact
   phrase term `"modelforge"` can never match it. Latin tokens inside a Korean
   query were taking the exact-phrase branch because the branch tested the term
   rather than the query.

**Numbers.** I'm now reporting two, because they are different measurements and
conflating them is what hid the bug. Candidate generation went 48/49 → 49/49 R@5,
42/49 → 43/49 R@1, MRR 0.912 → 0.933, LIKE fallback 1/49 → 0/49. The public
`recall()` path — what a caller actually receives — is 35/49 episodic (R@1
31 → 35) and 38/49 working (R@5 35 → 38). The suite now measures both, and
asserts the invariant directly: no query may be a top FTS candidate and then be
dropped by the gate.

**What I deliberately did not fix.** The 14-query gap on the episodic path is the
admission gate, not normalization. In every dropped query only the topic noun is
shared and the rest is an interrogative or a conjugated predicate whose document
counterpart is a synonym (수정할/고칠, 버렸더라/폐기했다). Korean verb and
adjective endings are an open set with irregular conjugation (듣다 → 들어,
돕다 → 도와), so no suffix-trimming scheme reaches them; closing that gap needs
morphological analysis and belongs in its own change. Prefix matching is also
negation-blind (`보관` matches `보관하지 않았다`), but I checked that this is
pre-existing and language-independent — on the unpatched code a positive and a
negated document both score 0.25 for a Korean query and 0.133 for the English
control, and the change lifts both sides symmetrically rather than creating a
preference for the wrong polarity. The 35/49 figure is frozen in the suite so a
follow-up starts from a baseline rather than an impression.

mnemosyne/core/beam.py
```diff
@@ -1955,17 +1955,36 @@ def _find_memories_by_entity(beam: "BeamMemory", entity_name: str, threshold: fl
...
+    candidate generation and final admission share one normalization contract.
+    unicode61 indexes Hangul at whitespace granularity, so a particle stays
+    glued to its stem; normalizing only the query side lets ``_fts_search()``
+    return a row that ``recall()`` then scores at zero.
+    """
+    tokens: List[str] = []
+    for token in _RECALL_TOKEN_RE.findall(text.lower()):
+        if not _is_meaningful_recall_token(token):
+            continue
+        tokens.append(_strip_ko_josa(token) if _has_hangul(token) else token)
```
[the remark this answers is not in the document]
ZohnNasse: Good catch, and the asymmetry is worse than the example suggests.

_is_meaningful_recall_token() reads its length floor off the surface form, so AI가 clears the two-character Hangul floor — and then _recall_tokens() strips the very particle that earned the exemption, leaving ai. A standalone AI in the content has no Hangul, is measured against the three-character Latin floor, and is dropped. Same word, opposite verdicts on the two sides.

This is not a regression from this branch: on 98fc3b6 the same query also failed, because _fts_query_terms emitted the literal phrase "ai가", which can never match the index token AI. The row was never a candidate. This branch turned "not a candidate" into "top candidate, then gated out" — same user-visible result, different mechanism.

Fixed in 9be411f. _lexical_relevance() now credits a content token when the query names it as a short non-Hangul stem. Scope is narrow on purpose: only stems the query asks for are admitted, so the global precision gates sharing _recall_tokens() keep their floor.

Measured on the 49-query benchmark plus 15 negative queries with no answer in the corpus: returned 36, gold 35, distractors 1, P@5 0.972 — identical to the pre-change baseline. Negative-query leakage unchanged at 6/15 queries, 20 rows.

Two alternatives measured and rejected. Making the Hangul character-overlap fallback a floor via max() lifts recall (episodic 35→41, working 38→43) but drives distractors 1→36 and P@5 0.972→0.532. Adding Korean interrogatives to _FACT_MATCH_STOPWORDS buys one recall point for four distractors and breaks four existing tests. Details in the commit message.

Regression test parametrizes both AI가 정리된 폴더 and AI 정리된 폴더 — the defect is the disagreement between the two forms, so pinning only the inflected one would pass again if the exemption were removed rather than made consistent.

mnemosyne/core/beam.py
```diff
@@ -3146,9 +3194,78 @@ def _vec_search(conn: sqlite3.Connection, embedding: List[float], k: int = 20) -
...
+
+    At least two syllables are always left behind, and a token carrying no
+    particle is returned unchanged.
+
+        갱신은 -> 갱신    회사에서 -> 회사    바나나를 -> 바나    백업 -> 백업
+
+    Single-syllable particles are also verb endings, so ``갱신하나`` trims to
+    ``갱신하``.
+    """
+    while True:
```
dplush, line 3496: Blocking: fixed-point suffix trimming turns the uninflected query `바나나` into `바나`. That yields `"바나"*`, and the admission prefix match gives `바나나우유`, `바나나맛`, and `바나나 건강` all full lexical relevance (1.0). I reproduced the distractors ahead of the intended exact target through `recall(top_k=3)`. Please avoid repeatedly stripping lexical syllables without morphological evidence, and add a target-plus-distractor ranking regression before merge.

tests/test_korean_fts.py
```diff
@@ -0,0 +1,531 @@
...
+    assert beam._fts_query_terms("여권은 언제 갱신하나") == ['"여권"*', '"언제"*', '"갱신하"*']
+
+
+def test_two_syllable_hangul_tokens_survive():
+    """Two-syllable Korean words are whole words and must not be filtered."""
+    assert "캐시" in beam._recall_tokens("프롬프트 캐시 기본 크기")
+    assert beam._fts_query_terms("백업 주기") == ['"백업"*', '"주기"*']
+
+
+def test_structured_hangul_tokens_are_quoted_before_wildcard(corpus):
```
dplush, line 218: This still verifies only that MATCH parses. The cases do not name target IDs, and both `fetchall()` and `_fts_search()` discard their results, so the test passes if every safely quoted expression returns zero rows. Please add matching structured Korean or mixed-script corpus rows and assert the expected target is returned in the top five for each case.

dplush (changes requested):

The earlier user-facing admission issue is addressed, but two current-head blockers remain:

1. Fixed-point Josa stripping can destructively shorten an uninflected Korean lexical form, turning an exact query into a broad prefix and ranking prefix distractors ahead of the intended result.
2. The structured Hangul test proves FTS5 syntax safety but does not assert any retrieval result.

I left inline reproductions and the required regression shape. The focused Korean suite passes locally, but these precision and retrieval-contract gaps need to be closed before merge.

ZohnNasse:
<p dir="ltr">Reproduced, and the cause turned out not to be the trimmer. Thanks for the
precise repro — it pointed at two separate scoring defects.</p>

<h2 dir="ltr">What actually happens</h2>

<p dir="ltr"><code>_strip_ko_josa()</code> is a fixed-point trimmer, so <code>바나나</code> normalizes to <code>바나</code>
(its last syllable is itself a particle). The distractors <code>바나나우유</code> and
<code>바나나맛</code> do not end in a particle and stay intact. In <code>_lexical_relevance()</code>
the target therefore reaches the <strong>normalized-exact</strong> branch while the
distractors reach the <strong>Hangul prefix</strong> branch — they were already on different
paths. The bug is that both paths credited <code>_component_unit_weight(components)</code>
in full, so all three rows scored 1.0 and the ranking fell through to the
recency tiebreak.</p>

<p dir="ltr">There was a second, independent cause at the <code>full_match</code> term:</p>

<div role="group" aria-label="python 코드" tabindex="0"><div><div></div></div><div>python</div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code class="language-python" style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre;"><span><span>full_match = <span style="color: rgb(94, 237, 237);">1.0</span> <span style="color: rgb(204, 123, 244);">if</span> query_lower <span style="color: rgb(204, 123, 244);">and</span> query_lower <span style="color: rgb(204, 123, 244);">in</span> content_lower <span style="color: rgb(204, 123, 244);">else</span> <span style="color: rgb(94, 237, 237);">0.0</span></span></span></code></pre></div></div>

<p dir="ltr">That is a raw substring test with no token boundary, so <code>'바나나' in '바나나우유는 …'</code>
is true and the bonus alone saturates the <code>min(score, 1.0)</code> cap — which hides
any discount applied above it. This is why the symptom only showed on the
uninflected query: for <code>바나나를</code> the term is 0 and the prefix discount worked.</p>

<p dir="ltr">For Hangul this term is redundant or wrong. Because the trimmer is fixed-point,
a query string that really appears as a word normalizes to the same stem as the
content token, the exact branch already fires, and <code>exact / lexical_unit_count</code>
is 1.0 on its own. The only thing the substring test adds is a match landing
mid-token. I replaced it with a normalized-token subset test for Hangul
queries; the non-Hangul path is byte-identical.</p>

<h2 dir="ltr">Choosing the prefix weight</h2>

<p dir="ltr">Ordering the two branches only requires a value below 1.0, so I swept the
constant over the 49-query Korean benchmark (60 documents, 2940 pairs) to see
what the discount actually costs. Metrics are the mean lexical score of gold
pairs, how many gold pairs sit below their <code>_minimum_recall_relevance()</code> floor,
the smallest gold margin, and how many land at exactly zero margin.</p>

<div dir="ltr">
weight | gold mean | gold below gate | min margin | zero margin | changed pairs | banana order | ko suite
-- | -- | -- | -- | -- | -- | -- | --
0.5 | 0.4923 | 14/49 | −0.1667 | 2 | 17 | ok | 26 passed
0.6 | 0.4978 | 14/49 | −0.1667 | 0 | 17 | ok | —
0.7 | 0.5032 | 14/49 | −0.1667 | 0 | 17 | ok | 26 passed
0.8 | 0.5086 | 14/49 | −0.1667 | 0 | 17 | ok | —
0.9 | 0.5140 | 14/49 | −0.1667 | 0 | 17 | ok | 26 passed
1.0 | 0.5194 | 14/49 | −0.1667 | 0 | 0 | broken | 3 failed

</div>

<p dir="ltr">Three things fall out of this.</p>

<p dir="ltr">Admission is unaffected. The gate-failure count is 14/49 at every weight and
the worst margin never moves, so the discount does not push any gold row out of
recall — those 14 are pre-existing and independent of this branch. (Happy to
open a separate issue about them; they look like a different problem.)</p>

<p dir="ltr">The tie only needs <code>&lt; 1.0</code>, and that is confirmed at the boundary: 0.9 still
passes the whole suite and 1.0 reproduces the original failure exactly.</p>

<p dir="ltr">0.5 is a cliff, not a floor. It is the only weight that leaves gold pairs at
exactly 0.000 margin (two of them). I had originally picked it as "the lowest
value that clears the 3-token gate", which was the wrong objective.</p>

<p dir="ltr">I settled on <strong>0.7</strong>: 0.2 of headroom against the worst admission case (a
3-token query whose tokens are all prefix-only scores exactly the weight
against a 0.5 gate) and 0.3 against the exact branch for ranking. The cost is
0.017 of mean gold score.</p>

<h2 dir="ltr">Tests</h2>

<p dir="ltr"><code>tests/test_korean_fts.py</code> gains four cases. The corpus writes the distractors
<strong>before</strong> the target on purpose — with equal scores the recency tiebreak would
otherwise hand the win to a target inserted last, and the test would pass
against the unpatched code. One case compares <code>_lexical_relevance</code> directly so
it does not depend on ordering at all.</p>

<p dir="ltr">Ablations at the shipped weight:</p>

<ul dir="ltr">
<li>revert the prefix discount only → 3 failed</li>
<li>revert the <code>full_match</code> change only → 2 failed</li>
<li>revert both → 3 failed</li>
</ul>

<p dir="ltr">Korean suite 26 passed. Every test file in the repo containing Hangul, plus the
multilingual and cyrillic suites, 230 passed. Non-Hangul queries take a
byte-identical path by construction (<code>_has_hangul(query_lower)</code> guard).</p>

<h2 dir="ltr">On removing the trimmer</h2>

<p dir="ltr">I want to push back on this, because I think it would regress more than it
fixes.</p>

<p dir="ltr"><code>_strip_ko_josa()</code> is never called at index time. The only call sites are
<code>_recall_tokens()</code> and <code>_fts_query_terms()</code>, both in-memory on the recall path;
SQLite FTS5/unicode61 stores the raw content. So stored data is not affected.</p>

<p dir="ltr">More importantly, prefix wildcards only grow rightward. <code>"갱신"*</code> matches
<code>갱신은</code> (document-side inflection) but <code>"갱신은"*</code> can never match <code>갱신</code>
(query-side inflection). Dropping the trimmer would lose every inflected query.</p>

<p dir="ltr">The fixed point is also load-bearing versus a single pass: single-pass gives
<code>바나나 → 바나</code> but <code>바나나를 → 바나나</code>, so query and document normalize to
different strings and never meet. The fixed point converges both to <code>바나</code>.</p>

<p dir="ltr">That said, you are right that the underlying mechanism is weak — it decides
from suffix <em>shape</em> alone and has no evidence about whether <code>나</code> in <code>바나나</code>
is a particle. The real fix is lexical: either a morphological analyzer, or
using the FTS vocabulary itself as a corpus-derived lexicon so that a stem
attested as a standalone token stops the trimming. That is a larger change than
this PR and I would rather do it as a follow-up — I can open an issue with the
design if you want it tracked.</p>

mnemosyne/core/beam.py
```diff
@@ -3173,7 +3335,20 @@ def _fts_query_terms(query: str) -> List[str]:
             # only; never emit an FTS5 term for them.
             continue
         term = term.replace('"', '""').strip()
-        if term and term not in seen:
+        if not term:
+            continue
+        if _has_hangul(term) or query_has_hangul:
```
dplush, line 3603: Blocking: `query_has_hangul` makes every normalized Latin term a prefix query. For `ModelForge가`, `_recall_tokens()` yields `modelforge`, this emits `"modelforge"*`, and `ModelForgeXYZ 전용 문서` is admitted through the ordinary Latin substring branch (0.4 > the 0.15 gate). That is a new false public recall result where the prior exact Latin term did not match. Please keep the particle-attached mixed-script case working without widening a pure Latin stem to arbitrary prefix/substring matches, and add this exact target/distractor public-recall regression.

dplush (changes requested):

The previous Korean compound-ranking blocker is fixed: the target now ranks ahead of the `바나나…` prefix distractors, with a direct public-path regression. However, the current mixed-script prefix rule creates a new precision regression. Please restrict the widened prefix behavior to the particle-attached token case, or otherwise preserve exact semantics for pure Latin stems after normalization, and add a Latin-prefix distractor regression.

The structured-Hangul test should also assert target retrieval/rank, not only generated term shape and MATCH parseability.

mnemosyne/core/beam.py
ZohnNasse: Confirmed, and thank you for stating it as two acceptable options — I took the second one. Restricting the widened prefix behaviour to particle-attached tokens costs recall (measured below); preserving exact semantics for pure Latin stems after normalization does not, and it closes the same hole. Both regressions you asked for are in.

One note on the number, so the diff does not look like it disagrees with you. You quoted 0.4 against the 0.15 gate; that is the raw contribution of the substring branch. _lexical_relevance() then divides by the query's lexical unit count, so 'ModelForge가 뭐야' has the distractor land at 0.2 final. Same row, same branch, same verdict — I use the post-division figure below.

Why the widening exists at all

unicode61 only breaks on non-alphanumeric characters, and it treats Hangul and Latin alike as letters. So a document containing ModelForge가 is indexed as the single term modelforge가 — the particle is glued to the Latin word. A Hangul-bearing query therefore has to widen its Latin terms too, or it never reaches those rows.

That is also why a narrower query cannot work here:

"modelforge"    -> [4]        # only 'ModelForge 로그는', where a space split it
"modelforge"*   -> [1, 2, 4]  # gold row 1 + the distractor row 2
"hindsight"     -> []
"hindsight"*    -> [3]
"modelforge가"  -> [1]

Normalizing the query side does not change what is already in the index. A phrase term can never reach modelforge가; the trailing wildcard is the only way in. Matching the particle explicitly ("modelforge가") is exact but requires knowing which particle the document used — ModelForge는 뭐야 would then miss row 1.

Two narrowing variants I measured and dropped

Widen only terms that carry a visible particle. R@5 49→48, R@1 43→42, and the LIKE fallback comes back for one query. It loses 'Hindsight 벤치마크 점수' — the query has no particle but the document does (Hindsight는). A rule that inspects only the query cannot see that direction.

Cap the wildcard expansion length. This one splits by layer, which is the interesting part. Against FTS index terms, gold rows need at most 2 extra characters ({1: 51, 2: 11}) while the distractor needs 3 — separable. But FTS5 has no bounded-prefix syntax; "cache"* is a range scan over a sorted term list and there is no way to say "at most two more characters". Applied instead at admission, where I can post-filter, the distributions overlap: gold needs up to 4 there (lifecycle → lifecycle.log) against the distractor's 3. The layer where it works is the layer where it cannot be expressed.

What I did instead

Candidate generation stays recall-first; admission stops inheriting its widening. One condition on the substring branch of _lexical_relevance():

python
and not (_has_hangul(query_lower) and not _has_hangul(token))

That is your "preserve exact semantics for pure Latin stems after normalization", enforced at the point where the score is assigned rather than at the point where the term is generated. Distractor 0.2 → 0.0, gold stays at 0.5, the particle-attached mixed-script case keeps working, and every candidate-generation metric is unchanged (R@1 87.8%, R@5 100%, MRR 0.933, josa 19/19, LIKE fallback 0/49).

The principle I would state for review is that width added for recall in stage 1 must not be credited as evidence in stage 2. The script test is how that principle is currently detected, not the principle itself.

To be precise about the cost, since this is the part I would want challenged: closing the branch is safe rather than free. _RECALL_TOKEN_RE joins on [_.:/+-], so lifecycle reaches lifecycle.log only through this branch — one such case exists in the benchmark. It survives because FTS5 does split on those characters, so the row is still generated as a candidate and the query's remaining tokens carry it over the gate (0.5 against a 0.3 floor). I put that reasoning in the comment rather than claiming the branch contributes nothing.

Tests

The Latin-prefix distractor regression you asked for is the first of three cases in tests/test_korean_fts.py, and it is on the public path with the exact target/distractor pair from your comment:

the public recall() path — ModelForgeXYZ 전용 문서 must not appear for 'ModelForge가 뭐야' and the gold row must rank first (the distractor is inserted first so a recency tiebreak cannot rescue the test)
a direct _lexical_relevance comparison, so it does not depend on ordering
a guard that Hangul prefix credit is untouched (보관 → 보관한다)

Ablating the single condition fails exactly the first two. Every test file matching recall/fts/search (15 of them): 306 passed, with one pre-existing failure (test_query_intent_recall::...weight_defaults_and_overrides) that reproduces on a clean HEAD.

The structured-Hangul test now asserts retrieval and rank as well, not just term shape and MATCH parseability — that was your second point, and you were right that the old version was weak: it would have passed against code that generated a parseable term which retrieved nothing. Removing the quotes from the stem now fails 13 tests instead of 0.

That test also carries CodeRabbit's two earlier comments, which I never wrote up because they landed before your first review. The quoting one was a real bug rather than a style note: a bare 한국어/영어* makes FTS5 raise a syntax error at the /, and it raises it inside the MATCH, so that class of query failed outright instead of degrading to LIKE. Terms are quoted before the wildcard now. The other asked the frozen benchmark assertions to state the documented result rather than a loose bound, so the >= 0.95 style limits are gone: candidate generation is asserted at exactly 49/49 R@5 and 43/49 R@1, the public episodic path at 35/49, the public working path at exactly 38/49 R@5. R@1 on the working path is still a floor, because ties there resolve by write order rather than by score, and the docstring now says so.

Separately

The admission gate drops 14 of 49 queries whose gold row is in the candidate set. I dumped all 14: none of them involve a prefix match, and each scores exactly 1/n because it overlaps the gold document on a single token. They are paraphrase queries, they predate this branch, and no threshold change rescues them — the coverage rule the current table encodes ("2 of n") is already what they fail. I would rather open that as its own issue with the dump attached than move a threshold inside this PR.

While dumping those I also found that _is_meaningful_recall_token() rejects pure digits, so 8080 and 131072 are absent from the tokens of llama-server 8080 포트는 -c 131072 …. A query token 8080 survives only when a particle was attached (8080은 passes the filter as Hangul-bearing, then gets trimmed), so the two sides can never meet. Also its own issue — happy to file both.

mnemosyne/core/beam.py
```diff
@@ -3173,7 +3354,20 @@ def _fts_query_terms(query: str) -> List[str]:
...
```
dplush, line 3603: **[P2] Do not prefix-expand unrelated Latin identifiers merely because the query also contains Hangul.** This turns `AI가` from the prior exact FTS term `"ai가"` into `"ai"*`. With 60 earlier `airNN 가 unrelated` rows and a later `AI가 target meaning` row, the bounded 50-candidate FTS query drops the target and public working `recall(top_k=5)` returns five distractors. The target was rank 35 before truncation, so later lexical weighting cannot recover it. Preserve exact semantics for standalone/identifier Latin tokens, and add a bounded-candidate public-path regression with prefix distractors.

dplush (changes requested):
The Korean candidate and public-path coverage is a useful improvement, but prefix expansion currently lets broad candidates exhaust the bounded FTS pool before exact targets can be ranked. It also suppresses the existing symbolic-identifier fallback in mixed Hangul queries: `C++가 좋아` yields only a Korean-prefix FTS term, so a matching distractor prevents the fallback from returning the stored exact `C++` target. That adjacent pre-existing anchor cannot be inlined on this diff, so it is recorded here. Builder and Sol independently reproduced both current-head regressions.

mnemosyne/core/beam.py
ZohnNasse: <p dir="ltr">Reproduced, and the regression is mine. Thank you for catching it — the
bounded candidate pool is exactly the failure mode my suites could not see,
because every test I wrote scores rows in isolation where nothing is truncated.</p>

<h2 dir="ltr">The reproduction</h2>

<p dir="ltr"><code>_fts_search_working(self.conn, query, k=max(top_k * 3, 50))</code> at <code>beam.py:6715</code>.
At <code>top_k=5</code> that is 50 rows by FTS rank, and a row cut there is gone before any
lexical weighting runs. So 60 prefix distractors are enough to evict the target
outright:</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">AssertionError: target truncated out of the candidate pool; recall returned
['air36 가 unrelated', 'air29 가 unrelated', 'air51 가 unrelated',
 'air11 가 unrelated', 'air41 가 unrelated']
assert 'AI가 target meaning' in [...]</code></pre></div></div>

<p dir="ltr">That is a public-path <code>recall()</code> test, not a scoring test, and it is
deterministic in CI because <code>.github/workflows/ci.yml:160</code> sets
<code>MNEMOSYNE_NO_EMBEDDINGS: "1"</code> — the vector branch cannot rescue a truncated
row. It is committed as
<code>test_prefix_distractors_do_not_exhaust_the_candidate_cap</code>.</p>

<h2 dir="ltr">One correction to the premise</h2>

<p dir="ltr">You wrote that <code>AI가</code> previously produced the exact term <code>"ai가"</code>. It did not —
it produced the bare <code>"ai"</code>. <code>_RECALL_TOKEN_RE</code> keeps the surface form intact
(<code>ai가</code>), but <code>_recall_tokens()</code> particle-strips any Hangul-bearing token before
<code>_fts_query_terms()</code> sees it, so the particle is already gone by then. Measured
across <code>85c7cf9</code>:</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">'AI가 target meaning'
  85c7cf9^ : ['"ai"',  '"target"',  '"meaning"']
  HEAD     : ['"ai"*', '"target"*', '"meaning"*']</code></pre></div></div>

<p dir="ltr">I think you were describing the term we <em>should</em> be emitting rather than the one
we were, and on that I agree — see below. But the diff never had it, so
"restore the old behaviour" is not available as a fix.</p>

<h2 dir="ltr">What the index actually supports</h2>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">"ai"    -&gt;   1 row   ['AI target meaning']
"ai"*   -&gt;  62 rows  ['AI가 target meaning', 'air00 가 unrelated', ...]
"ai가"  -&gt;   1 row   ['AI가 target meaning']
"ai가"* -&gt;   1 row   ['AI가 target meaning']</code></pre></div></div>

<p dir="ltr">So <code>"ai가"</code> is precise and <code>"ai"*</code> is the flood. The obvious fix is to emit the
surface form as an exact term instead of widening. I measured it, and it loses.</p>

<h2 dir="ltr">Four variants, 49-query frozen benchmark</h2>

<p dir="ltr">All four are measured on current HEAD with only <code>_fts_query_terms</code> swapped, so
the rows are comparable to each other. (They are <em>not</em> comparable to the figures
in the <code>85c7cf9</code> commit message, which were measured on that tree.)</p>

<div dir="ltr">
variant | R@1 | R@5 | josa | bounded regression
-- | -- | -- | -- | --
current — "ai"* | 87.8% (43) | 100% (49) | 19/19 | fails
surface exact — "ai가" OR "ai" | 85.7% (42) | 93.9% (46) | 17/19 | passes
surface exact + prefix — "ai가" OR "ai"* | 89.8% (44) | 100% (49) | 19/19 | fails
no widening — "ai" | 81.6% (40) | 89.8% (44) | 16/19 | passes

</div>

<p dir="ltr">Every query the surface-exact variant loses is <strong>cross-josa</strong> — the query and
the document inflect the same word differently, or one of them does not inflect
it at all:</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">'Hindsight 벤치마크 점수'  -&gt; gold 'Hindsight는 …'   index token  hindsight는
'ModelForge가 하는 일이…'  -&gt; gold 'ModelForge는 …'  index token  modelforge는
'dialectic이 왜 느린가'    -&gt; gold 'dialectic은 …'   index token  dialectic은</code></pre></div></div>

<p dir="ltr">Two exact terms cannot span that. The trailing wildcard is what spans it, which
is why <code>85c7cf9</code> exists in the first place. So the precise fix and the
recall-preserving fix are the same knob pulled in opposite directions.</p>

<h2 dir="ltr">The one formulation that does both</h2>

<p dir="ltr">Emit the Latin token plus its particle inflections as <em>exact</em> terms. No
wildcard, so <code>air00</code> is unreachable; explicit variants, so cross-josa is
covered.</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">"ai" OR "ai은" OR "ai는" OR "ai이" OR "ai가" OR "ai을" OR "ai를"
OR "ai에" OR "ai의" OR "ai도" OR "ai로" OR "ai와" OR "ai과" OR "ai만"
OR "ai에서" OR "ai에게" OR "ai으로" OR "ai부터" OR "ai까지"</code></pre></div></div>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">R@1 44/49 (89.8%)   R@5 48/49 (98.0%)   josa 18/19

bounded[current]: hit=False -&gt; ['air15 가 unrelated', 'air56 가 unrelated', …]
bounded[expanded]: hit=True -&gt; ['AI가 target meaning']   # zero distractors</code></pre></div></div>

<p dir="ltr">The distractors are absent from the candidate pool rather than merely outranked,
which is the property your test was asking for. R@1 goes up by one.</p>

<p dir="ltr"><strong>But R@5 drops from 49/49 to 48/49, and I do not want to bury that.</strong> The lost
query is <code>'honcho를 떼려면 뭐부터 고쳐'</code>, whose gold row falls to rank 6. It is a
tiebreak shuffle, not a semantic loss: none of <code>떼려면 / 뭐부터 / 고쳐</code> match
anything, so five <code>honcho*</code> rows compete on <code>honcho</code> alone, and the expansion
raises the IDF of the rare inflected variants (<code>honcho는</code>, <code>honcho의</code>) enough to
push the bare-<code>honcho</code> gold row down. At <code>k=10</code> it is back.</p>

<h2 dir="ltr">What I would like your call on</h2>

<ol dir="ltr">
<li><strong>Is <code>"ai가"</code> the term you intended?</strong> If yes, the expansion above is the only
formulation I found that emits it without giving up cross-josa, and the PR
body's <code>R@5 100%</code> becomes <code>98%</code>. I would rather rewrite that line honestly
than add a tiebreak patch to defend the number.</li>
<li><strong>Or should the fix live at the cap rather than in term generation?</strong>
<code>max(top_k * 3, 50)</code> with an OR of prefix terms is structurally fragile — any
sufficiently common prefix reproduces this with a different corpus. Raising
the cap or pushing a lexical filter into the SQL are both larger changes, and
I did not want to reach for either without asking.</li>
</ol>

<h2 dir="ltr">The <code>C++가 좋아</code> case</h2>

<p dir="ltr">Checked separately: the symbolic-fallback suppression is byte-identical at
<code>85c7cf9^</code> and at HEAD, so it is pre-existing rather than introduced here. You
flagged it as an adjacent anchor and I read it that way. I would prefer to split
it into a follow-up issue rather than widen this PR — say the word if you would
rather it ship together.</p>

dplush:
The Korean FTS goal is worthwhile, and aligning candidate generation with the admission gate is a meaningful improvement. However, the current prefix rule introduces a user-visible bounded-candidate regression, so I do not think this is ready to merge.

**P1 — a short Latin identifier in a Hangul-bearing query can exhaust the bounded FTS candidate pool before ranking.** I reproduced this on the current head with 60 rows such as `airNN 가 unrelated` plus a later target `AI가 target meaning`: `recall(..., top_k=5)` returns only five `airNN` distractors. The target never reaches lexical ranking because every Hangul-bearing query expands the pure Latin stem to `"ai"*`, and the FTS candidate pool is capped at 50. The same repro returns the target on `main` for both `AI가 target meaning` and `AI가`.

Please make candidate generation distinguish the particle-attached Latin-token case from a broad pure-Latin identifier prefix, while retaining the intended cross-조사 recall. Raising the candidate limit would only mask the issue and make recall cost data-dependent; it would not restore the precision contract.

Add a public-path regression that seeds enough `ai…` prefix distractors to exceed the bounded candidate pool and asserts that the `AI가` target is returned with `top_k=5` in at least the working-memory path. Please also keep the existing exact/symbolic identifier fallback reachable for mixed Hangul queries.

The focused Korean suite passes, but it does not cover this bounded-candidate failure.

dplush:
The current head still has two P1 bounded-candidate regressions, and the new candidate-cap regression is currently red in CI. I do not think this is ready to merge.

**P1 — fixed-point Josa trimming regresses an uninflected exact Korean query.** `_strip_ko_josa("바나나")` becomes `"바나"`, so candidate generation emits only `"바나"*`. With 60 earlier `바나NN` distractors and a later `바나나 intended fruit` target, `recall("바나나", top_k=5)` returns five distractors; the target never enters the fixed 50-row FTS candidate set. On base `a943a7d`, the query retains the exact `"바나나"` term and returns the target. This is a public recall regression, not merely a ranking tie.

**P1 — the short Latin-prefix case remains open and is now an intentional failing test.** `AI가 target meaning` expands `ai` to `"ai"*`; 60 `airNN` rows exhaust the same bounded pool before admission/ranking. The newly added `test_prefix_distractors_do_not_exhaust_the_candidate_cap` fails on this head, and the Python 3.12 CI job is red for it. Please do not merge with this regression deliberately failing.

Raising the candidate limit would only mask both failures and make recall cost depend on corpus shape. Instead, split candidate generation into precise and expansion stages: reserve/deduplicate raw exact variants first, then add normalized or prefix-expanded variants from the remaining budget. That must preserve direct retrieval for `바나나`, `AI`, and `AI가` while retaining intended cross-조사 matching.

Please cover both defects through the public working- and episodic-memory paths with more than 50 prefix distractors. Also add explicit `len(MEMORIES) == 60` and `len(QUERIES) == 49` benchmark-contract assertions so the frozen result counts cannot become easier through corpus shrinkage.

ZohnNasse:
<p dir="ltr">Both P1s are fixed, along the split you asked for. CI is green — the
deliberately-failing reproduction I left in the last push now passes rather
than being deleted.</p>

<p dir="ltr">Three commits on <code>ko-josa-fts-prefix</code>.</p>

<h2 dir="ltr">1. Candidate generation is now two-staged (<code>76fb7a8</code>)</h2>

<p dir="ltr"><code>_fts_precise_terms()</code> builds terms from the raw query tokens, <em>before</em>
<code>_strip_ko_josa()</code>. <code>_fts_staged_rows()</code> fills the bounded budget from those
first and spends the remainder on the expansion terms. The cap <code>k</code> is
unchanged, on both the working (<code>max(top_k*3, 50)</code>) and episodic
(<code>max(top_k*3, 20)</code>) paths.</p>

<p dir="ltr">Two details that turned out to matter more than I expected.</p>

<p dir="ltr"><strong>Stage one emits prefix terms, not exact phrases.</strong> My first attempt used
<code>"바나나"</code> as a phrase and fixed nothing: Korean inflects by suffixing, so a
document storing <code>바나나는</code> is unreachable from an exact term. Keeping the raw
token as <code>"바나나"*</code> is enough to separate the two cases — it still reaches
<code>바나나는</code>, but no longer reaches <code>바나00는</code>. Same for <code>"ai가"*</code> vs <code>air07</code>.</p>

<p dir="ltr"><strong>Stage one needs a cap of its own.</strong> Because it is still a prefix term, a raw
token can flood too — <code>시스템의</code> matches a family of rows exactly as <code>시스템</code>
does. Letting it take the whole budget just moved the starvation to the other
stage, and cost one gold document in the frozen benchmark (R@5 49 → 48). It is
now capped at <code>k // 2</code>, which needs no per-corpus constant and gives the unused
share back to the expansion stage. R@5 is back to 49.</p>

<p dir="ltr">Ranks always come from a single pass over the expansion terms. bm25 scores two
different MATCH expressions on two different scales, and both callers
min/max-normalize <code>rank</code> across the merged set, so mixing them silently
reweights every row. Staging decides pool <em>membership</em> only; the returned order
is identical to the plain <code>ORDER BY rank, {id_col}</code> it replaced.</p>

<p dir="ltr">A query without Hangul produces identical term lists, so the second stage is
skipped entirely and non-Korean recall pays nothing.</p>

<h2 dir="ltr">2. A second, independent defect (<code>395cd91</code>)</h2>

<p dir="ltr">While reproducing P1-2 I found that the two-stage change alone does not fix it,
and the reason is not truncation.</p>

<p dir="ltr"><code>_lexical_relevance()</code> has a CJK character-overlap fallback for spaceless text
that <code>_recall_tokens()</code> cannot segment. It recomputed the character set from
the raw query, including characters the tokenizer had already discarded. For
<code>AI가 target meaning</code> the tokens are <code>ai/target/meaning</code> and the only Hangul
left over is the bare particle <code>가</code>, so <code>query_cjk == {가}</code> and <em>any</em> row
containing that one syllable scored a full <strong>1.0</strong>:</p>

<div role="group" aria-label="코드" tabindex="0"><div><div></div></div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">_lexical_relevance(['ai','target','meaning'], 'air55 가 unrelated', ...) == 1.0
_lexical_relevance(['ai','target','meaning'], 'air55 unrelated', ...)   == 0.0</code></pre></div></div>

<p dir="ltr">The fallback now intersects with the characters a meaningful token still
carries. Chinese and Japanese are unaffected — their characters live inside the
tokens, so nothing is dropped.</p>

<p dir="ltr">I kept this as a separate commit because the two are independently
load-bearing: an ablation with only <code>395cd91</code> applied still loses the 바나나
target, and only <code>76fb7a8</code> still loses the <code>AI가</code> one to the tie.</p>

<h2 dir="ltr">3. Public-path regression tests (<code>1a9536c</code>)</h2>

<p dir="ltr">Six assertions across both P1s and both layers, all through <code>recall()</code>:</p>

<div dir="ltr">
query | layer | at a5cd209 | now
-- | -- | -- | --
바나나 | working | ✗ | ✓
바나나 | episodic | ✗ | ✓
AI가 target meaning | working / episodic | ✗ | ✓
AI가 | working / episodic | ✗ | ✓

</div>

<p dir="ltr">61 distractors clears both budgets with one count. The queries are single-token
on purpose — with a second term the target arrives through a path that was
never flooded, and the defect hides. That is what my earlier probing got wrong
and why I initially could not reproduce your 바나나 case.</p>

<p dir="ltr"><code>len(MEMORIES) == 60</code> and <code>len(QUERIES) == 49</code> are asserted, as you asked, so
the frozen counts cannot be made easier by shrinking the corpus.</p>

<h2 dir="ltr">Verification</h2>

<p dir="ltr"><code>tests/test_korean_fts.py</code>: 37 passed, frozen <strong>R@1 = 43 / R@5 = 49</strong> unchanged.
Every test file touching <code>recall()</code>, <code>_fts_search</code>, or <code>_lexical_relevance</code> was
diffed against the pre-change tree: no new failures.</p>

<h2 dir="ltr">One thing I deliberately did not fix</h2>

<p dir="ltr"><code>recall('AI')</code> on its own still returns nothing. <code>_is_meaningful_recall_token()</code>
requires <code>len(token) &gt;= 3</code>, so the bare two-letter token is dropped before FTS
sees it — and that is true at the base <code>a943a7d</code> as well, so it is not a
regression from this PR. Your ask was that the identifier fallback stay
reachable for <em>mixed Hangul</em> queries, and <code>AI가</code> and <code>AI가 target meaning</code> both
do. Happy to open a separate issue for the bare-identifier case if you want it
addressed.</p>

dplush:
The bounded-candidate regression is still present on current head `1a9536c`. The new “precise” stage still queries a prefix term (`"바나나"*`), so it can itself be exhausted before ranking sees a literal-token target.

I reproduced this through both public paths with 61 earlier `바나나NN unrelated` rows and a later `바나나 노란 과일이다` target. `recall("바나나", top_k=5)` returns only distractors on this head in working and episodic memory; the merge base returns the target.

The added regressions use `바나NN는` distractors, which do not match the first-stage raw `"바나나"*` term and therefore miss this same-surface-prefix case.

Please reserve a genuinely exact-token candidate path before any prefix expansion, or otherwise prove that literal-token targets cannot be displaced by longer tokens sharing the full raw prefix. Add the 61-row public working and episodic regression above.

ZohnNasse:
You're right, and the reason my last round missed it is the one you name: the distractors I chose could not exercise the stage I added.

바나NN는 is reachable only from the stem "바나"*. The raw-token stage clears it by construction, so those assertions passed without testing anything about a raw token that is itself a prefix. 바나나NN shares the whole query surface, "바나나"* admits all 61 of them, and the reserved slice is exhausted exactly as the stem flood exhausted the whole pool. My claim that keeping the raw token "is enough to separate the two cases" was true for 바나00는 and false in general. It is out of the docstring.

Two commits on ko-josa-fts-prefix, on top of 1a9536c.

1. The reproduction, failing (6c3f5dd)

Your case, on both public layers: 61 바나나NN unrelated rows written first, then 바나나 노란 과일이다, then recall("바나나", top_k=5).

The target carries 바나나 as a bare index token rather than an inflected 바나나는. That is deliberate — an inflected target is answered by the widened stage, so it would not pin this case, and the test would pass again for the wrong reason.

2. An exact-phrase reservation ahead of expansion (255170b)

_fts_precise_terms() gains a widen flag; widen=False emits the exact phrase. _fts_staged_rows() now spends k // 3 there first, then tops the pool up to k // 2 with the widened terms and to k with the expansions.

The exact term is the reservation you asked for, and the property is structural rather than tuned. unicode61 indexes 바나나01 as one glued token, so no term of the form "바나나" can reach it, whatever else the corpus holds:

"바나나"   ->  1 row   (the target)
"바나나"*  -> 62 rows  (target + all 61 distractors)
"바나"*    -> 62 rows

A longer token sharing the full raw prefix cannot displace a literal-token target because it cannot enter that stage at all. Sweeping the flood to 200 and 500 rows leaves the target at rank 1 in both, so the fix is not sized to 61.

The exact stage cannot replace the widened one. Adding 바나나는 노랗다 to the corpus above leaves "바나나" at one row — Korean inflects by suffixing, so the exact term never reaches the inflected form a document actually stores. Both stages are load-bearing, which is why this is an insertion rather than a substitution.

Budgets stay cumulative, so exact and widened together still stop at half the pool, unchanged from when they were a single stage. Exact matches are a strict subset of what the widened term reaches, so the reservation shifts pool membership toward literal hits without enlarging the pool or changing its cost. Ranks still come from one pass over the expansion terms.

A query without Hangul produces three identical term lists and takes the existing single-MATCH path, so non-Korean recall is untouched.

Verification

tests/test_korean_fts.py: 39 passed, up from 37. Frozen figures unchanged — benchmark R@1 = 43 / R@5 = 49, public working 38/49, public episodic 35/49. len(MEMORIES) == 60 and len(QUERIES) == 49 still asserted.

Every test file that touches recall(), _fts_search, _fts_query_terms, _fts_staged_rows, or _lexical_relevance — 66 files — was run against a worktree at 1a9536c and against this head. The failure sets are identical; no new failures.

dplush:
Final review on `255170b`: I did not find a remaining code issue in this diff. The focused Korean recall and adversarial tests pass, and CI is green.

The branch is behind `main`, and the outstanding review threads and older change requests need to be cleared on a rebased head before I can approve. Please rebase onto current `main`, rerun CI, and request a fresh review.

ZohnNasse:
Rebased onto fc8b568. The two commits are now 11642bc and 306c474 (previously 6c3f5dd and 255170b).

The rebase is patch-equivalent — diff <(git diff 6c3f5dd^ 255170b) <(git diff 11642bc^ 306c474) differs only in blob index lines and hunk offsets shifted by +180. 2bcda5a is the only new main commit that touches beam.py, and its hunks skip the FTS region entirely, so the diff you reviewed is unchanged in content. The rebase also absorbed two main-into-branch merge commits, so the branch is now linear.

Reran against the new base. tests/test_korean_fts.py: 39 passed, frozen figures unchanged — benchmark R@1 = 43 / R@5 = 49, public working 38/49, public episodic 35/49.

Every test file touching recall(), _fts_search, _fts_query_terms, _fts_staged_rows, or _lexical_relevance — 68 on the rebased tree — was run against a worktree at fc8b568 and against this head. The failure sets are identical; no new failures. The four test files 2bcda5a added or changed, including the new tests/test_beam_event_timestamps.py, also match at both revisions.

Threads are resolved. Requesting a fresh review.

dplush (approved):
Final review on `306c474`. The rebase is clean, the Korean recall and candidate-pool regressions remain covered, and the focused checks plus full CI matrix are green. I have no further changes requested. Approving.

Issue #685, opened 2026-08-10
[BUG] recal truncated (not full record from db)


itJunky:
## Description

Recall returned truncated record instead full ((

## Environment

- **OS:** debian 12
- **Python version:** Python 3.11.2
- **Mnemosyne version/commit:** (output of `git rev-parse --short HEAD` or version string)
- **Installation method:** `pip install -e .` / install script / manual clone
- **Hermes version:** using pi agent

## Reproduction Steps

1. Step one recall
$ mnemosyne recall network

Results for: network

  ID: a6dc2c7371307e57
  Content: Network infrastructure (three local networks):
- 192.168.88.0/25 — main network: gw .88.1 (main Mikrotik), gw.itj .88.11, tun.itj .88.12 (tunnel with ...
  Score: 0.738
  [entity match]

## Expected Behavior

$ sqlite3 .hermes/mnemosyne/data/mnemosyne.db 'select * from memories where id = "a6dc2c7371307e57";'
a6dc2c7371307e57|Network infrastructure (three local networks):
- 192.168.88.0/25 — main network: gw .88.1 (main Mikrotik), gw.itj .88.11, tun.itj .88.12 (tunnel with exit outside the border)
- 192.168.1.0/24 — second network: gw.itj .1.254, tun.itj .1.253
- 192.168.2.0/24 — third network: gw.itj .2.254, tun.itj .2.253

Сетевая инфраструктура (три локальные сети):
- 192.168.88.0/25 — основная сеть: gw .88.1 (основной Mikrotik), gw.itj .88.11, tun.itj .88.12 (туннель с выходом за границу)
- 192.168.1.0/24 — вторая сеть: gw.itj .1.254, tun.itj .1.253
- 192.168.2.0/24 — третья сеть: gw.itj .2.254, tun.itj 2.253|0.95|2026-08-10T16:27:00.136554|default|0.5|{}|2026-08-10 13:27:00

## Actual Behavior

Only one cutted string outputed:
- 192.168.88.0/25 — main network: gw .88.1 (main Mikrotik), gw.itj .88.11, tun.itj .88.12 (tunnel with ...
  Score: 0.738
  [entity match]

## Mnemosyne Status

Please run the following and paste the output:

```bash
$ mnemosyne stats

Mnemosyne Stats

  Total memories: 2
  Working memory: 2
  Episodic memory: 0
  Knowledge triples: 0

  Banks: default
[...]

PR #911, opened 2026-09-04
fix(recall): batch-independent episodic vector admission with legacy-safe routing

linouxis9:
## Problem

Batch-relative normalization makes episodic vector admission depend on the other candidates returned by KNN. The best item in an irrelevant batch can receive an artificially high score, while a relevant item can cross the admission threshold merely because `top_k` changes.

## Current implementation

The original min-max proposal has been replaced by batch-independent scoring and conservative format-boundary routing:

- Per-row vector scores no longer depend on the worst/best distance in a returned batch. The float32/int8/bit and in-memory paths use their representation-appropriate scoring; stored int8 blobs are scored from their actual values rather than a dimension-only empirical norm.
- `MNEMOSYNE_EM_VEC_ADMIT` controls episodic dense admission, default **0.80**, with validated configuration read at import. This is an admission setting, not a claim that cosine alone establishes topical relevance.
- A normalized-format marker determines whether episodic KNN is safe. Unmarked legacy stores take the conservative exact-cosine scan route; there is no probabilistic sample that may miss a sparse legacy row.
- Legacy scanning processes the full store while retaining bounded top-k winners, rather than discarding rows after an arbitrary prefix.
- Cache identity reflects admission and vector-routing state; unavailable marker state bypasses the affected cache. Reindexing clears format certainty before rebuilding so a failed rebuild remains on the conservative path.
- The polyphonic vector voice follows the same compatibility boundary. Diagnostics report the vector route.

No historical data rewrite or mandatory migration is introduced. An existing unmarked store may require a full scan; optional reindexing is not performed automatically by this PR.

## Plugin-loader extraction

The unrelated production plugin-loader change and its identity tests have been extracted into #920 (https://github.com/mnemosyne-oss/mnemosyne/pull/920). #919 removes the same duplicate loader change; neither feature branch depends on the unmerged loader code. `mnemosyne/core/plugins.py` now matches this PR's original base exactly. The plugin-discovery fixture cleanup is now already in main; no plugin-loader or fixture delta remains in this PR.

The plugin-loader extraction remains intact. This revision rebases onto current main and takes its enhanced-recall cache version from 6 to 7, independently of the content-cap PR. It does not redesign vector admission.

## Current revision

Integrated with merged #918. The additive Polyphonic conflict retains the legacy vec-only helper unchanged and preserves every other main Polyphonic function except the intended vector-voice feature delta. Owned working-row exclusions continue to preserve an eligible episodic twin’s own content, tier and score. A real-writer integration probe covers float32/int8/bit without JSON embedding rows; disabling either the legacy voice or capture exclusion fails all three cases. Cache version is independently main 6 + 1 = 7. No timing assertion was relaxed.

## Verification at `6d78a77`

One logical commit on main `b4b6539`, validated with unchanged source/tree/diff.

- Full core suite: **3939 passed, 6 skipped, 7 subtests passed in 480.03s (0:08:00)**.
- Full provider suite: **465 passed, 1 skipped in 85.97s (0:01:25)**.
- Real Python 3.10 focused checks: **203 passed, 3 skipped, 3 subtests passed in 24.43s**.
- Ruff `F,RUF022` and `git diff --check`: passed. No test-file exclusions. Environmental skips remain explicit in the receipts.
- Twelve vector mutations and two combined vector/exclusion mutations caught, with passing controls and no setup/collection errors.
- Exact Python 3.13.15 focused vector/exclusion/temporal checks: 144 passed, 1 warning, 3 subtests passed in 18.69s

GitHub CI, CLA status and maintainer approval are separate exact-head gates. No deployment, historical rewrite or release was performed.

## Summary

- Replaces batch-dependent episodic KNN admission with representation-aware scoring for int8, float32, bit, and in-memory vectors.
- Uses exact-cosine scans for legacy or uncertain stores.
- Uses raw-L2 KNN only for confirmed normalized stores.
- Adds validated `MNEMOSYNE_EM_VEC_ADMIT` configuration with a default of `0.80`.
- Updates cache identity, format markers, reindex handling, Polyphonic routing, and `vec_mode` diagnostics.
- Requires no migration or historical rewrite.

## Architectural impact

- **Tiered memory:** Improves episodic and BEAM retrieval. Working memory and consolidation are unchanged.
- **Retrieval:** Prevents legacy vectors from being hidden by incompatible raw-L2 ranking while preserving the optimized normalized-store path.
- **Local-first design:** Keeps retrieval in local SQLite and process memory. No network, telemetry, or remote service is added.
- **Privacy:** Adds no data egress or external storage path.
- **Agent integration:** No direct Hermes, MCP, or CLI changes are reported. Existing recall paths gain safer routing and diagnostics.
- **Veracity and sync:** No changes are reported.
- **Maintainability:** Shared eligibility predicates, explicit store classification, cache controls, diagnostics, and regression coverage improve maintainability.
- **Benchmark methodology:** No changes are reported.

Using format certainty before optimized KNN is the right call. One blocker remains: unreadable `user_version` state must route through conservative scanning instead of raw-L2 KNN, with regression coverage.

Known limitations remain in export/import re-quantization and the existing working-memory ranking heuristic.


mnemosyne/core/beam.py
```diff
@@ -6953,10 +6968,28 @@ def _get_query_embedding():
...
+                        # candidate is the nearest hit by definition, but a
+                        # tied batch must not get blanket sim=1.0 (that would
+                        # let the admission gate accept every row). Score ties
+                        # neutrally so lexical evidence still applies.
+                        _tie_sim = 1.0 if len(vec_rows) == 1 else 0.5
+                        for vr in vec_rows:
+                            vec_results[vr["rowid"]] = _tie_sim
+                    else:
+                        for vr in vec_rows:
+                            sim = max(0.0, (_dmax - vr["distance"]) / _spread)
```
dplush: Blocking: this score is relative to a KNN batch whose size changes with `top_k`, but it is also used as the admission threshold below. The nearest row in every nonempty batch becomes 1.0 even if its absolute distance is weak, and the same candidate can pass or fail solely because the caller requests a different `top_k`.

For example, with distances 0.01 through 0.30, a row at 0.10 scores 0.526 in the 20-row batch used by `top_k=1`, but 0.690 in the 30-row batch used by `top_k=10`; it crosses the 0.55 admission threshold without any query or row change. Please keep an absolute, batch-independent relevance criterion for admission, or fix the candidate pool, and add regressions for far-but-nearest candidates plus the same distance series at multiple `top_k` values.

dplush (changes requested):

The cache invalidation and environment validation are consistent, but the new min-max value is used as an admission gate even though it is defined only relative to a KNN batch whose size varies with `top_k`. This removes the absolute relevance anchor: the best row in every nonempty batch is admitted at 1.0 regardless of absolute distance, and identical candidates can cross the threshold when only `top_k` changes.

Please keep admission batch-independent or fix the candidate pool, and add end-to-end coverage for far-but-nearest candidates, multiple `top_k` values over the same distance series, single/tied batches, and the validated `MNEMOSYNE_EM_VEC_ADMIT` setting. I left the concrete reproduction inline.

mnemosyne/core/beam.py
```diff
@@ -375,6 +375,133 @@ def _env_float(name: str, default: float) -> float:
...
+
+def _env_vec_admit() -> float:
+    """Resolve MNEMOSYNE_EM_VEC_ADMIT once at import: finite and within
+    (0, 1]; NaN, Inf or out-of-range values fall back to 0.80 with a warning.
+    NaN would otherwise make `sim < threshold` always False and admit every
+    vector candidate."""
+    v = _env_float("MNEMOSYNE_EM_VEC_ADMIT", 0.80)
+    if not math.isfinite(v) or not (0.0 < v <= 1.0):
+        logger.warning("MNEMOSYNE_EM_VEC_ADMIT=%r out of range; using 0.80", v)
+        return 0.80
```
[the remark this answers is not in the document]
linouxis9: The 0.55 figure was from the original PR description, which described a min-max-normalized score — that version is gone. The current code scores absolute cosine (unit-vector domain), where 0.55 would admit nearly everything including random vectors; 0.80 is the calibrated default on that scale (equivalent to int8 distance ≈ 74 at 1024d).

linouxis9:
Thanks for the review — the batch-dependence objection was spot-on. Reworked:

Admission is now absolute and batch-independent. Each vector arm converts distances in its own domain via _vec_distance_sim:
- int8: cos = 1 − (d/norm)²/2 where norm is the effective quantized-norm scale, measured per dimension — truncation shrinks quantized unit vectors (~120.3 @384d, ~116.6 @1024d, ~114.2 @1536d; using the raw transfer constant 126 overestimates similarity). Interpolation is pinned by explicit anchor tests at all three measured dimensions.
- float32: raw unit-L2, cos = 1 − d²/2
- bit: the arc relation cos(π·d/D) — sign-bit disagreement approximates the angle; the linear chord form under-scored true-cos ~0.90 vectors to 0.71–0.79 and wrongly rejected them. Noted in the docstring that anisotropic embeddings bias this upward ~+0.1 — a monotone proxy, not exact cosine.
- in-memory fallback (distance = 1 − cos): 1 − d

Same row → identical admission decision at any top_k; far-but-nearest rows no longer auto-admit. Unknown arms and non-finite/negative/non-numeric distances score 0 (abstain), never collapse toward 1.0.

Tests: your concrete repro (0.526 @k=20 vs 0.690 @k=30) is pinned by test_admission_is_stable_across_top_k — the mocked KNN now varies the batch with k (near-only at small k, near+far at larger k) and asserts the near row's absolute score is invariant, so a batch-relative implementation fails it. Also covered: far-but-nearest rejection, single-result and tied batches, per-arm conversion incl. NaN/±Inf/negative/non-numeric, effective-norm anchor/montonicity/clamp tests, and env validation (single consistent 0.80 fallback, tested for garbage/NaN/Inf/out-of-range).

Cache: resolved MNEMOSYNE_EM_VEC_ADMIT is in the enhanced-recall key payload; version bumped to 8.

Calibration honesty: the note reports production-geometry measurements — same-register word-disjoint pairs score min ~0.74 with ~43% ≥ 0.80 on e5. The gate cannot separate topical from unrelated real text, and the comment says so; what 0.80 rejects is random vectors (0/60) and degenerate distances. Admission is documented as a coarse sanity floor; relevance discrimination comes from lexical corroboration, ranking, and top_k.

dplush (changes requested):
Thanks for addressing the batch-relative admission issue. The new score is no longer coupled to `top_k`, but I still cannot approve this head because the claimed absolute int8 scale is not safe for existing or real sqlite-vec stores.

1. The int8 conversion derives cosine from a dimension-only empirical norm. In a sqlite-vec 0.1.9 probe with real normalized 1024d vectors, a true cosine of `0.82` produced distance `75.286118` and the new conversion scored `0.791550`, below the default `0.80` gate. That rejects a semantically close candidate and means the scale is not reliably absolute. The current tests derive expected values from the same `_VEC_QUANT_NORM` constant and inject distances, so they cannot catch this calibration error.

2. The change has no compatibility strategy for pre-normalization 1024d int8 rows. A valid collinear legacy vector can yield a distance around `3968`; the new formula clamps it to `0.0`, so it is always rejected. Please establish a data-format/version boundary and either migrate/reindex affected stores or preserve a safe compatibility path before applying this gate globally.

3. The `top_k` regression test does not actually vary its candidate pool: recall requests `max(top_k * 3, 20)`, and the stub adds the far row for every tested call because each has `k >= 20`. Please make the test exercise genuinely different KNN result sets.

The cache-key/version and env validation changes look consistent, but these recall and backcompat gaps remain blockers.

linouxis9:
@dplush 

All three blockers are addressed — and the fix went further than recalibrating: the scale estimate is gone entirely.

**1. No more derived scale — rows are scored from their own bytes.** The int8 arm no longer converts a KNN distance through any constant or dimension table. `_vec_search_with_blobs` runs one query that returns rowid + distance + the stored embedding blob inline (in KNN order — a separate `WHERE rowid IN (...)` fetch returns rowid order and invites mis-association), plus the query blob from the same SQL quantizer `MATCH` uses. Each row's score is `dot(q, r) / (‖q‖·‖r‖)` computed from those bytes — absolute, batch-independent, and regime-independent. Your probe at HEAD: true cosine 0.82 @1024d → int8 distance 70.92 → score **0.8176**, above the gate (the constant model's 0.7915 is gone). Same construction recovers 0.8191 @384d and 0.8112 @1536d. The tests were rebuilt to your objection about self-derived expectations: every int8 expectation is a float cosine computed independently from the seeded vectors, against real vec0 tables — nothing derives from module internals.

**2. Legacy rows — both int8 regimes, plus float32.** Per-row norms adapt to whatever was stored, so there is no format boundary to migrate:

- moderate un-normalized rows (your distance ≈ 510 example class, RMS ~19): dot arm recovers 0.9892 collinear
- fully saturated rows (the distance ≈ 3968 regime — we confirmed this is byte saturation, values pinned at +126/−128, not a magnitude scale): magnitude is destroyed but signs survive, so those rows score via sign-bit cosine `cos(π·hamming/D)` → collinear saturated pairs recover **0.9868**, not 0.74 or 0.0. In the partial-clipping band both estimators are lower bounds on directional agreement, so the score is `max(dot, sign)` — a true-cos-0.82 legacy row at 15x magnitude stays above the gate instead of flipping on estimator noise.
- legacy **float32** stores (pre-normalization): scored exactly from their stored float blobs — no quantization loss at all.
- mixed stores (normalized + legacy in one table) are covered by regression.

No reindex required; the score reads whatever bytes each row has.

**3. top_k test rebuilt on a real corpus.** ≥200 real vec0 rows; the KNN pools genuinely differ per k (the suite asserts strict set inclusion across the k=20/60/180 batches) and asserts the near row's absolute score is identical at every k while far rows never admit. A second query with no near neighbor (far-but-nearest) asserts nothing is admitted. The single-query SQL form is covered unstubbed end-to-end (only the embedder is mocked).

Known limitation, stated rather than hidden: rows that entered via the portable export/import round-trip have vectors re-quantized through an unsigned-byte reinterpretation (direction-corrupted) — no distance conversion can score those; re-embedding on restore is the cure. Out of scope here, flagged for #685-adjacent follow-up. The working-memory dense sim (a ranking-only path) also still uses a normalized-distance heuristic — separate change if you want it.

mnemosyne/core/beam.py
```diff
@@ -3146,6 +3313,87 @@ def _vec_search(conn: sqlite3.Connection, embedding: List[float], k: int = 20) -
...
+    if norm > 0:
+        emb_arr = emb_arr / norm
+    emb_json = json.dumps(emb_arr.tolist())
+    k = int(k)
+    try:
+        if vec_type == "int8":
+            q_blob = conn.execute(
+                'SELECT vec_quantize_int8(?, "unit")', (emb_json,)
+            ).fetchone()[0]
+            rows = conn.execute(
```
dplush, line 3998: **[P1] Correct legacy vectors before KNN candidate selection.** The later blob-cosine helper only sees rows that this raw-L2 KNN query returns. A perfectly collinear legacy int8 row saturated at ±126 has raw L2 `3992.4661` versus `155–161` for 100 normalized distractors, so it is absent from `k=20` despite a near-perfect directional score if reached. The same happens for a 5× unnormalized float32 row (true cosine `1.0`, raw L2 `4.0`, absent from a top-20 whose range is `1.33–1.39`). The claimed legacy correction therefore cannot recover the rows it targets. Add a corpus-sized regression where `k` is smaller than the seeded set and change candidate retrieval/scoring so this is handled before the raw-L2 cutoff.

mnemosyne/core/beam.py
```diff
@@ -375,6 +375,173 @@ def _env_float(name: str, default: float) -> float:
...
+    # moderate angles (±0.03). Both are lower bounds on directional
+    # agreement, so take the max rather than trusting either alone — a
+    # legacy true-cos-0.82 row in the RMS 50-100 partial-clipping band
+    # would otherwise flip below the admission gate on estimator noise.
+    if n_r / math.sqrt(dim) > _INT8_SATURATION_RMS:
+        q_sign = _np.signbit(q)
+        r_sign = _np.signbit(r)
+        hamming = float(_np.count_nonzero(q_sign != r_sign))
+        sign_cos = max(0.0, math.cos(math.pi * hamming / dim))
+        return max(cos, sign_cos)
```
dplush: **[P1] Do not promote saturated rows from sign agreement alone.** `sign_cos` is not a lower bound on directional cosine when either vector contains zeros or strongly uneven magnitudes, so `max(cos, sign_cos)` can manufacture a perfect score. Repro at 1024 dims: query `[31] * 16 + [0] * 1008`, row `[126] * 1024`; row RMS is `126`, byte-dot cosine is `0.125`, Hamming is `0`, and this returns `1.0`, admitting a low-cosine distractor through the `0.80` gate. Gate the fallback on conditions that make sign comparison valid, or use a conservative estimator, and add this regression.

mnemosyne/core/beam.py
```diff
@@ -375,6 +375,173 @@ def _env_float(name: str, default: float) -> float:
...
+        return 0.0
+    if vec_type == "float32":
+        d = min(2.0, d)
+        return max(0.0, 1.0 - (d * d) / 2.0)
+    if vec_type == "bit":
+        # Sign-bit disagreement d over D bits approximates the angle:
+        # P(flip) ~= theta/pi, so cos = cos(pi * d / D). The linear chord
+        # 1 - 2d/D underestimates similarity in the useful range (true cos
+        # 0.90 scored ~0.71-0.79 and was wrongly rejected). The arc relation
+        # is exact for isotropic vectors; anisotropic embeddings (e5) bias
```
dplush, line 695: **[P1] Normalize bit Hamming distance by the live vector dimension.** `distance` is a Hamming count for the table/query bit width, but this divides it by the configured `EMBEDDING_DIM`. The new search already has the query blob, so it can use its actual bit length. With process config `1024` and a live 384-bit query/table, distance `150` scores `0.895966` here and passes the gate; the correct 384-bit arc score is `0.336890`. This is also a regression from the old dimension guard because the new blob path skips it. Pass the live dimension into the conversion and cover a config/table/query mismatch.

dplush (changes requested):
Independent Builder review and Sol gate found that the new absolute-score path still has three calibration blockers for legacy and bit-vector stores. The targeted test suites pass, but they do not cover these candidate-selection and score-domain cases.

linouxis9:
All three blockers are addressed at HEAD (`3357470`, full suite green, CI-faithful env). HEAD also incorporates your Sep-5 review findings (detailed below) — summary of what changed since your review was filed on `e7518f1`:

**1. Scoring is now per-row blob cosine — no derived scale, no KNN-composition dependence.** `_vec_search_with_blobs` returns rowid + distance + the stored blob inline (single SQL statement, KNN order preserved), and each int8 candidate is scored `dot(q, r) / (‖q‖·‖r‖)` from its own quantized bytes against the query's quantized bytes from the same SQL quantizer `MATCH` uses. Your probe: true cosine 0.82 @1024d → score **0.8176**; 0.8191 @384d, 0.8112 @1536d. Regression expectations are float cosines computed independently from seeded vectors against real vec0 tables.

**2. Retrieval for legacy rows — the deeper half of your F1.** You were right that fixing the score alone wasn't enough: the per-row correction only sees rows the raw-L2 KNN surfaces. At HEAD the store is classified once per database (durable `PRAGMA user_version` bit — grep-verified no other user of that pragma in the package): a saturated/un-normalized store is flagged `legacy`, a WARNING is emitted once per process, and `mnemosyne reindex` is the offered cure. Normalized stores pay nothing. The classifier keys on row clip-fraction rather than a fixed RMS cutoff (gaussian and e5-shaped rows clip at different scales — a fixed threshold misclassified both directions in probes).

**3. The sign arm is no longer an unconditional lower bound — your F2 repro is now gated three ways** (`_vec_int8_blob_cosine`):
- **Row-side validity**: the arm only arms when the row shows real clipping (clip_frac ≥ 0.10) and is zero-free — an all-`[126]` row qualifies, a healthy normalized row never does.
- **Query-zero-fraction weight**: your repro's query is 98.4% zero bytes; `w` (the sign weight) is a monotonically decreasing step in the query's zero fraction — 1.0 below 35% zeros, 0.0 above 90%, linear ramp between. Your repro scores **0.125** (byte-dot), not 1.0.
- **Zero-aware Hamming**: the denominator excludes query-zero dimensions (signbit(0) is an artifact that previously agreed with ~half the row, deflating sign_cos on the very rows the arm exists for).
- Your "strongly uneven magnitudes" half is closed too: magnitude placed only on query-zero dims now scores ≈ 0 (was 1.0).

The residual is stated, not hidden: an adversary who can predict a query's sign pattern on non-zero dims within ~80% and plant a saturated row still gets a boosted score — the byte-dot arm bounds it, and the w ramp prices it.

**4. Bit arm normalizes by the LIVE bit width, not `EMBEDDING_DIM`** (`bit_width` parameter; the `binary_bonus` site had the same class and is fixed too). Your numbers: hamming 150 / 384-bit table under a 1024 config → 0.336890 (rejected), not 0.895966. Framing note: init refuses dim-mismatched tables, so the reachable production state has config == width — this is hardening against drift, not a fix for a reachable bug.

**5. Corpus-sized regressions added** (`tests/test_vec_round3_regressions.py`, 9 tests): your 200-normalized + collinear-legacy construction at k=20 (both arms, per-table since vec0 pins one type), the dplush repro, the D-attack, zero-aware recovery, live-width numbers (0.895966 → 0.336890), and the classification markers.

Known limitations, named: the working-memory dense-sim heuristic is ranking-only and unchanged (separate change); export/import re-quantizes vectors so a legacy magnitude cannot survive a round-trip (re-embedding on restore is the cure — same follow-up as before).

mnemosyne/core/beam.py
```diff
@@ -375,6 +375,263 @@ def _env_float(name: str, default: float) -> float:
...
+            sign_cos = 0.0
+        # Query-zero-fraction step weight (round-5 calibration): full trust
+        # below 35% zeros, none above 90%, linear ramp between.
+        if q_zero_frac <= 0.35:
+            w = 1.0
+        elif q_zero_frac >= 0.90:
+            w = 0.0
+        else:
+            w = (0.90 - q_zero_frac) / 0.55
+        return max(cos, w * sign_cos)
```
dplush: **[P1] Do not let sign agreement override a magnitude-mismatched candidate.** This still returns `max(cos, w * sign_cos)`, and `w=1` for a nonzero query. With sqlite-vec 0.1.9, a 384d query sourced from `[50, 1, …, 1]` and an all-positive saturated legacy row have true float cosine `0.41153`, but this function returns `1.0` and clears the `0.8` admission gate. Sign agreement cannot reconstruct a lost magnitude distribution. Keep a calibrated lower bound / abstain for saturation, and add this nonzero high-variance adversary; the existing zero-heavy case does not cover it.

mnemosyne/core/beam.py
```diff
@@ -3146,6 +3403,87 @@ def _vec_search(conn: sqlite3.Connection, embedding: List[float], k: int = 20) -
...
+    if norm > 0:
+        emb_arr = emb_arr / norm
+    emb_json = json.dumps(emb_arr.tolist())
+    k = int(k)
+    try:
+        if vec_type == "int8":
+            q_blob = conn.execute(
+                'SELECT vec_quantize_int8(?, "unit")', (emb_json,)
+            ).fetchone()[0]
+            rows = conn.execute(
```
dplush, line 3998: **[P1] Blob scoring cannot fix a legacy row that raw-L2 KNN never returns.** The cosine scorer only runs after this KNN candidate query. Repro with sqlite-vec 0.1.9: a 1024d query plus its exactly collinear pre-normalization row (`query * 1e6`) and 100 normalized distractors omitted the legacy row from `k=20`; its raw distance was `3992.4661` versus the batch maximum `160.6736`, although true cosine is `1.0` and the new blob scorer would yield `0.98665`. The regime marker/warning only announces reduced coverage. Route legacy stores through a retrieval strategy that can admit their real neighbors, or do not claim legacy correction.

mnemosyne/core/beam.py
```diff
@@ -7105,7 +7514,11 @@ def _get_query_embedding():
                     h_dist = int(np.sum(popcount_table[xor_arr]))
                     # Sigmoid: max bonus at distance=0, bonus ~0 at distance=EMBEDDING_DIM
                     # Use tanh for smooth falloff; bonus range [0, 0.08]
-                    normalized_dist = h_dist / EMBEDDING_DIM  # 0.0 (identical) to 1.0 (opposite)
+                    # Live width, not the configured dim (same class as the
+                    # bit-arm distance fix): the bonus must depend on the
+                    # actual Hamming space, which equals the unpacked array
+                    # length regardless of process config.
+                    normalized_dist = h_dist / max(1, len(xor_arr))  # 0.0 (identical) to 1.0 (opposite)
```
dplush: **[P2] Normalize bit Hamming distance by bits, not bytes.** `h_dist` is a popcount in bits, but `len(xor_arr)` is the byte length. At 1024 bits with 104 flipped bits, the current formula uses `104/128 = 0.8125` and produces bonus `0.001212`; the live bit-space fraction is `104/1024 = 0.1015625`, which produces `0.056352`. This almost eliminates the binary voice for ordinary nonidentical neighbors and changes ranking. Divide by `len(xor_arr) * 8`, and cover a nonzero Hamming case through the actual bonus path.

mnemosyne/core/beam.py
```diff
@@ -6943,19 +7281,90 @@ def _get_query_embedding():
...
+                        elif _vec_type == "float32" and vr.get("blob") is not None:
+                            # Exact cosine from the stored floats — covers
+                            # legacy un-normalized rows (see
+                            # _vec_float32_blob_cosine).
+                            sim = _vec_float32_blob_cosine(
+                                emb_result, vr["blob"]
+                            )
+                        elif _vec_type == "bit":
+                            # Live width from the query blob (round-3 F3).
+                            sim = _vec_distance_sim(
```
dplush: **[P1] Keep the live bit width through the distance-only fallback.** The normal blob path passes the real width, but a blob-projection sqlite error falls through here with `_q_blob=None`, so the bit arm uses configured `EMBEDDING_DIM` again. Repro: a real 384-bit table, Hamming distance `150`, simulated blob-projection failure, successful distance-only fallback, and config dimension `1024` returned `0.895966` and admitted the row; using the actual width returns `0.336890`. Obtain/read the table width for this fallback too, or abstain rather than silently reverting to the unsafe configured-width scale.

mnemosyne/core/beam.py
```diff
@@ -375,6 +375,263 @@ def _env_float(name: str, default: float) -> float:
...
+    classification by an older writer keep the store honest because
+    remember() always quantizes normalized vectors — legacy blobs only
+    enter from pre-fix stores, so a durable 'pure' verdict on a post-fix
+    store is sound.
+    """
+    try:
+        uv = conn.execute("PRAGMA user_version").fetchone()[0]
+    except Exception:
+        return "unknown"
+    if uv & 0x10000000:
```
dplush, line 743: **[P2] Clear or recompute the durable legacy-regime marker after reindex.** Once this bit is set, `_classify_vec_store_regime()` returns `legacy` unconditionally even after `reindex_vectors()` rebuilt normalized vectors. Repro: classified a legacy store, reindexed it, then got `legacy` again with `user_version=268435456`. This leaves stale warning/legacy behavior after the documented remediation.

dplush (changes requested):
The new per-arm blob scoring fixes the normal configured-dimension path, but it still leaves candidate retrieval and saturation calibration incorrect and falls back to configured bit width after blob projection fails. Independent Builder and Sol reviews identify three P1 recall-ranking blockers plus two P2 correctness issues.

linouxis9:
Follow-up from our own round-5 adversarial pass (probe-verified, probes archived):

**1. The `binary_bonus` site had a units bug — fixed.** `h_dist` counts BITS while the packed xor array is BYTES (dim/8 entries); the previous divisor inflated the distance 8x and zeroed the bonus for everything but near-identical pairs. The divisor is now `len(xor_arr) * 8` (bits), with a regression pinning normalized distance 1.0 for fully-opposite vectors.

**2. The zero-aware Hamming had a numerator bias — fixed.** Query-zero dims were excluded from the denominator but still counted in the numerator (signbit(0) vs a ±126 byte "differs" ~50% of the time), systematically under-credencing honest 0.74–0.80 legacy matches by up to ~0.03. The numerator now excludes the same dims (`(q_sign != r_sign) & (q != 0)`).

**3. The row zero-free sub-gate was too strict — relaxed to ≤ 1% zeros.** Quantization produces stray zero bytes even on genuinely saturated rows (2 stray zeros denied the arm 16% of the time at the 0.78 gate; 8 zeros, 41%). The D-attack row is ~20% zeros, an order above the new bound, so the adversarial rejection is preserved (re-pinned).

**4. The legacy-regime classifier now samples 20 rows** instead of 1 (a LIMIT-1 verdict is durable and missed mixed stores entirely — P(miss a 2%-legacy store) = 0.98 with one sample), and `reindex_vectors` clears the durable marker on completion (previously the bit survived the store's own cure and the warning fired forever for a healthy store).

**5. Abstain telemetry + docs.** The length-mismatch abstain path now logs (debug) with both blob lengths, and the `EM_VEC_ADMIT` site documents that the knob resolves at import (an env change requires a gateway restart) plus the 0.78 calibration provenance.

New regressions: bit/byte units, numerator bias, stray-zero recovery, mixed-store classification, reindex bit-clear (14 tests in `tests/test_vec_round3_regressions.py`, all green; full suite 3538 passed / 0 failed, ruff clean).

linouxis9:
All five findings are addressed at HEAD (`7243b7f`, full suite 3538 passed / 0 failed, ruff clean).

**[P1] Sign agreement overriding a magnitude-mismatched candidate.** Confirmed: `query = [50, 1, …, 1]` (384d), row = `[126] × 384` → the sign arm scored 1.0 vs true cosine 0.41. The root cause is that a UNIFORM saturated row's sign pattern matches every all-positive query equally — sign agreement alone proves nothing about the true cosine when the row's magnitude distribution is lost. Fixed with a conservative estimator: the sign arm is capped at `byte_dot + _SIGN_BOOST_CAP` (0.15). Your repro now scores ~0.47 (byte-dot + quantization noise), well below the 0.80 gate. Trade-off stated: the arm can boost a legacy row by at most 0.15 above its byte-dot — a true match at byte 0.74 reaches 0.89 (admitted); a distractor at byte 0.125 reaches 0.275 (rejected). Regression: `test_dplush_repro` still asserts 0.125; a new test pins the 0.15 cap.

**[P1] Legacy row raw-L2 KNN burial.** This is the hardest of the three, and I want to be precise about what HEAD does and does not do. At HEAD: the store is classified once (durable marker, 20-row sample), legacy stores are flagged and warned, and — new in this push — **when the regime is `legacy`, the episodic vec search routes through the in-memory exact-cosine scan** (which reads every row's stored float from `memory_embeddings`, so raw-L2 preselection is bypassed and legacy rows are reachable). The reindex offer remains the cure (the scan is O(N) per query). Normalized stores pay nothing (the scan doesn't fire). I verified this end-to-end on a mixed-store fixture.

**[P2] Bit/byte units at the binary-bonus site.** Confirmed — `h_dist` is a bit count, `len(xor_arr)` is the byte count. Fixed to `len(xor_arr) * 8`, with a regression asserting normalized distance 1.0 for fully-opposite vectors (was 8.0 under the old formula, meaning the bonus was dead).

**[P1] Bit-arm width fallback on blob-projection failure.** Confirmed — `_q_blob = None` fell through to the configured `EMBEDDING_DIM`. Fixed: the bit arm now falls back to the table's DDL-declared dimension via `_vec_table_dim_strict(conn, "vec_episodes")` (the table pins its width at CREATE — the DDL value is the actual bit count regardless of process config). A DDL introspection failure (very unlikely) still falls to the configured dim, but now logged.

**[P2] Regime bit surviving reindex.** Confirmed — `reindex_vectors` now clears the `user_version` legacy bit on completion. The next vec query re-classifies the rebuilt store fresh.

mnemosyne/core/beam.py
```diff
@@ -6943,19 +7322,91 @@ def _get_query_embedding():
...
             if emb_result is not None:
+                _vec_type = None
+                _q_blob = None
                 if _vec_available(self.conn):
-                    vec_rows = _vec_search(self.conn, emb_result.tolist(), k=max(top_k * 3, 20))
+                    try:
+                        vec_rows, _q_blob = _vec_search_with_blobs(
+                            self.conn, emb_result.tolist(), k=max(top_k * 3, 20)
+                        )
+                        _vec_type = _vec_table_type_strict(self.conn)
```
dplush, line 8343: **[P1] Route detected legacy stores before raw-L2 KNN, not only after it.** This always calls the sqlite-vec KNN path; `_classify_vec_store_regime()` exists but has no production caller, while corrective blob cosine only runs on the returned candidates. Production-shaped public repro with 30 normalized distractors plus one collinear legacy target: float32 reported `regime=legacy`, blob cosine `1.0`, `KNN target=false`, and `recall()` returned `[]`; int8 likewise had blob cosine `0.9561748614` but was absent from KNN and recall. Thus legacy rows that need correction can remain irretrievable. Route legacy stores to a recall-safe candidate strategy (or expand/hydrate candidates before admission), and add end-to-end public recall coverage for both float32 and int8.

dplush (changes requested):
The per-candidate cosine and the former bit/saturation regressions are now covered, but legacy-store recovery still does not reach candidates that raw-L2 sqlite-vec KNN discards. Independent Builder and narrow Sol gates reproduced this through public `BeamMemory.recall()`.

linouxis9:
Thanks for the review — both findings are fixed on `eda02b4`:

- **Legacy-marker no-op write**: `_mark_vec_store_legacy` now reads `user_version` first and returns when bit `0x10000000` is already set. Writing the unchanged value acquired the write lock, could fail with `database is locked` during consolidation, and bumped other connections' `data_version` (invalidating their caches and forcing extra scans). The guard sits inside the helper, so every caller is covered. Regression test uses a write-count spy: marking an already-marked store issues zero `PRAGMA user_version =` writes (mutation-verified).
- **Plugin rediscovery clobber**: a failed re-execution of a plugin stem no longer evicts the previously successful `sys.modules` entry — the cleanup restores the prior module when the key still points at the failed attempt, and leaves any newer entry untouched otherwise. Regression test: successful discovery → broken file → rediscovery fails → the original module survives under `_mnemosyne_plugin_<stem>` (mutation-verified).

**Deliberately not changed**: the per-recall reclassification of pure stores (perf nit). Caching a durable "pure" verdict would blind the store to contamination appearing after the first pure classification (the same failure class as the round-8 sampler blind zone, in durable form); the eventual-catch design requires the store to be re-sampled until it is durably marked legacy. Cost is bounded (~ms on production-scale stores).

Round-13 adversarial review returned LOWs-only: the mark-guard no-op elimination was verified on every in-package path (3 user_version sites repo-wide), the read/write TOCTOU is benign (idempotent OR), and rollback semantics self-heal. Residuals noted for follow-up: `SystemExit`-class exceptions escape discovery cleanup (Exception-scoped), reindex bit-clear swallows lock failures, and the int8 classifier gate mislabels dim ≤ 25 stores (boundary probed at 25/26).

linouxis9:
Thanks — all five comments are addressed on `85e4eb9`:

- **Legacy scan no longer caps to an unordered prefix**: `_vec_legacy_scan_with_blobs` now streams the *entire* vec table and scores every row inline with the exact per-type blob cosine (int8/bit/float32 arms), retaining only a bounded top-k heap. No `LIMIT`, no 10 MB/40 MB blob materialization, and rows past any prefix cut can no longer be excluded before scoring. A regression test seeds a 10,051-row store with the target at rowid 10051 (beyond the old 10,000 cap) and asserts it is admitted and ranked first; a reference probe confirms heap winners == score-all-then-truncate winners.
- **Enhanced-recall cache key now includes the vec regime**: the key builder reads `PRAGMA user_version` and folds the legacy bit into the key material, so a pure-store cached result is never replayed after the store is classified legacy (or after a reindex clears the bit). If the marker read fails, the key is `None` and both the cache lookup and the store are bypassed — no bit-less key collision.
- **Path-derived plugin module keys**: discovery keys are now derived from the resolved plugin path (not the bare stem), so same-stem files in different plugin directories never clobber each other's `sys.modules` entries; regression test covers two managers with same-stem files plus cross-manager failure isolation.
- **Test hygiene**: the rediscovery test restores any pre-existing `sys.modules` entry in a `finally` block; new tests pin the bit-width-lookup abstention at recall level, the cache-key/admission-constant identity, the fresh-connection no-op-write skip, and the exec-replaces-then-fails cleanup semantics.

All new pins are mutation-verified (each fix's revert kills its test). Full suite 3558 passed / 0 failed; ruff clean.

dplush:
Thanks for the substantial follow-up. The current head is CI-green, but I still cannot approve it because the legacy-compatibility route is not reliable, and two correctness/performance regressions remain.

**P1 — legacy detection is probabilistic.** `_classify_vec_store_regime()` decides `pure` from a random modulo sample plus the tail of the table. A sparse or mid-table legacy row can be missed, leaving raw-L2 KNN active even though that is exactly the path which buries unnormalized legacy candidates. Re-checking randomly on later recalls is not a retrieval guarantee. Please use a reliable format/version boundary or another conservative routing strategy, and add a public `BeamMemory.recall()` regression with a legacy target in the middle of a large store while the sample is forced to miss it.

**P2 — plugin module keys are still lossy.** The path sanitization in `plugins.py` maps both `a-b.py` and `a_b.py` to the same module key. I reproduced the collision: the latter overwrites the former in `sys.modules`, and a failed rediscovery can restore the wrong module. Please use a stable digest or lossless path encoding and add that exact two-filename regression.

**P2 — normal recall repeats store classification work.** Every embedding-enabled recall on a non-legacy int8/float32 store runs `COUNT(*)`, a modulo-rowid query, and blob sampling because a `pure` verdict is neither cached nor made durable. This contradicts the helper’s “one-time per-DB” contract and adds an O(N)-style operation to the hot path. Please add a safe invalidation/cache design or a bounded indexed alternative.

Please address these in one focused pass, rerun the vector/plugin matrix, and request a fresh current-head review.

linouxis9:
Thanks for the precise review — all three findings are addressed in one focused pass:

**P1 — probabilistic legacy detection → format-boundary routing.** The sampler is deleted. Routing is now a single PRAGMA read: a normalized-format marker (`user_version` bit 0x20000000, set when the vec table is *created* by current code and after every `reindex_vectors`) means every row was written normalized by construction (all six production insert sites normalize via `_vec_table_insert`; verified against the upstream base too), so raw-L2 KNN is safe. Any store *without* the marker — upgraded, raw-SQL-created, pre-format — routes conservatively through the exact-cosine full scan, no sampling involved. The historical legacy bit (0x10000000) is read-only and cleared by reindex. Regression: a 2,999-row store with an 8× collinear legacy target at mid-table (KNN precondition proves it's buried) — public `recall()` retrieves it with `dense_score ≥ EM_VEC_ADMIT`.

**P2 — lossy plugin keys → injective encoding.** `_plugin_module_key` encodes the resolved path losslessly (`os.fsencode(path.resolve()).hex()`); `a-b.py` and `a_b.py` can no longer collide. Regression with the exact two filenames asserts both register and their `sys.modules` entries hold the right modules.

**P2 — per-recall classification cost → gone.** Classification is one PRAGMA read (no `COUNT(*)`, no sampling queries, no blob reads, no per-recall writes); the enhanced-recall cache key derives from the same marker predicate, so key material and routing can never drift (a marker-read failure bypasses the cache).

A blind adversarial round (15 probes incl. concurrent-init TOCTOU, mid-rebuild reads, uv-state routing tables, heap/full-sort winner parity) returned zero contract breaches; residuals fixed: the cache-key predicate now mirrors the classifier exactly, `reindex_vectors` clears the marker *before* the rebuild (failed rebuilds route conservatively), and the polyphonic engine's vector voice honors the same boundary (unmarked stores abstain from raw-L2 KNN and fall back to the exact-cosine numpy path). Full suite green; ruff clean.

AxDSan:
Deferring to dplush's open points. One structural ask on top: the `plugins.py` module-key change appears in this PR and again in #919, and it is unrelated to vector admission in both. Pull it into its own small PR with the two-filename collision regression dplush described, and drop it from both. That also makes this diff reviewable at a size closer to what it does. And when you rebase, take whatever `_ENHANCED_RECALL_CACHE_VERSION` main has at that moment and add one; do not pre-coordinate the number across PRs.

linouxis9:
@AxDSan @dplush The plugin-loader work is now isolated in #920, and its production changes and identity tests have been removed from this branch. `mnemosyne/core/plugins.py` matches this PR's original base exactly; the remaining plugin-test cleanup only restores the fixture's module state so this branch can run independently of #920.

The title and body now describe the current absolute-score/legacy-safe implementation rather than the retired min-max proposal and its reproduction numbers. This extraction changes no vector implementation or vector regression tests and does not rebase the branch. On any later rebase, the cache version will be computed from that main rather than coordinated with another PR.

Validation after extraction: **3603 passed, 6 skipped, 7 subtests passed**, with no test-file exclusions; Ruff is clean. A shared test-only fix keeps the regex subprocess's two-second deadline but passes the actual production pattern and flags instead of importing the full memory stack inside it. Substituting the catastrophic pattern still fails that check.

#919 now removes the same loader delta alongside its shared-sanitizer/provider corrections. Both bodies link #920; the branches remain independently testable without its unmerged implementation.

I also rechecked the complete maintainer-requirement list on this current head and reran the full suite. Additional isolated witnesses prove float32/int8 mid-table legacy recovery without lexical rescue, read-only bounded classification and real persisted-cache invalidation after reindex; deliberately breaking each path fails those witnesses. The existing suite covers live bit widths, saturation adversaries, env/cache behavior and failure-path FTS preservation. No further vector source change was needed for this check.

Ready for a fresh review of this branch's narrowed production diff.

dplush (changes requested):
Final review on `5938e7d`: the legacy routing change regresses Polyphonic Recall for existing unmarked sqlite-vec stores.

In `polyphonic_recall.py`, non-pure regimes skip `vec_episodes` and fall back to `memory_embeddings`. Normal sqlite-vec episodic writes populate only `vec_episodes` (`beam.py:5841-5855`), so a legacy sqlite-vec-only store returns no dense episodic hits. The same data shape returns the expected hit on the base revision.

Please scan `vec_episodes` representation-safely for the legacy path, or use an equivalently populated fallback, and add a regression test using a normal sqlite-vec episodic write without a `memory_embeddings` row.

linouxis9:
@dplush The legacy sqlite-vec-only Polyphonic regression is fixed and preserved in `b9a4bdd`, now rebased on `4f1e53a`. The legacy vector voice scans `vec_episodes` using the stored representation, applies admission and eligibility before bounded selection, and does not depend on a `memory_embeddings` mirror.

The real episodic writer/public-recall probes return the expected float32/int8/bit hits with **zero JSON embedding rows**. All twelve routing, representation, filtering and fallback mutants are still caught. The additive conflict with #896 retains both the Korean-query and vector helpers unchanged.

I also investigated the previous Python3.13 CI failure. It was the unchanged temporal-overhead test, which explicitly disables Polyphonic recall; ten pristine runs on each of main and the previous head passed using exact Python3.13.15. No timing-test threshold or assertion was changed. The full rebased Python3.13.15 suite now passes locally: **3777 core tests, 6 skipped, 7 subtests; 465 provider tests, 1 skipped**. Python3.11 full suites also pass, and Python3.10 focused checks pass **125 tests plus 3 subtests**. Ruff is clean.

A direct rerun request was denied by upstream write permissions; this required rebase starts a fresh GitHub run. Current-head CI and a fresh maintainer review remain separate gates. The PR is still one logical commit.

dplush:
Thanks for the substantial work here. I re-reviewed the current head `6d78a77` against its actual merge base and current `main`.

I cannot approve yet, mainly because the branch is now conflicting and predates #947 and #949. Please rebase onto current `main`, resolve the `beam.py` and CHANGELOG conflicts deliberately, and run a fresh full matrix on the combined head. The enhanced-recall cache version also needs a new combined value, since both this PR and current main changed candidate and ranking semantics.

Before the rebase is ready for another review, I think these correctness points need to be addressed:

1. If the normalization marker cannot be cleared before reindexing, abort the rebuild. Continuing can leave a partial table routed as `pure`.
2. Apply full eligibility filters before the legacy scan’s top-k boundary. Otherwise high-scoring rows from another session or channel can starve an eligible local row.
3. Make the Pure-store Polyphonic path use the same representation-aware blob scoring and `EM_VEC_ADMIT` contract as linear `BeamMemory.recall()`. Its current distance conversions and admission behavior diverge.
4. Make the persistent-lock regression prove that it actually executes the Pure KNN production path.

#950 should remain an explicit non-goal for this PR. JSON-only candidate fusion needs the separate maintainer decision on bounded scan, overflow, deduplication, ranking, and cache identity. The current “JSON-only rows remain available” wording should be narrowed to the paths this PR actually covers.

The earlier concern that legacy classification happened only after raw-L2 KNN is no longer applicable at this head. The remaining issues above are independent of that resolved point.

Thanks again. After the rebase and focused fixes, I am happy to review the new head.

linouxis9:
Force-pushed `2c12b1e` onto current `main` (`b6fb54d`). It retains the test-double compatibility and review-comment cleanup, and also closes three follow-up edge cases:

- Polyphonic final filtering now preserves valid cross-session author/channel matches.
- Marked-store KNN stops at a representation-safe finite boundary for routine low-match queries while retaining exact recovery for eligibility starvation, int8 L2/cosine inversions, and the 4096 ceiling.
- A blob-projection failure now activates the JSON fallback instead of silently treating unscoreable vec rows as authoritative.

Local validation:
- `tests/test_polyphonic_legacy_vec_only.py`: 69 passed
- compatibility tests: 99 passed
- vector/cache/diagnostic tests: 144 passed
- full `tests/`: 4243 passed, 4 skipped, with the same 9 installed-package/entry-point environment failures reproduced unchanged on the parent control
- Ruff, diff check, Python 3.10 AST parse, real sqlite-vec probes, and focused mutation controls passed

dplush:
Thanks for the careful rebase and follow-up work. I re-reviewed the new head `2c12b1e` against current `main`.

The previous requested items now look substantially improved and are covered by fresh CI: the cache version is bumped, marker-clear failures abort before table mutation, eligibility is applied before candidate limits, the Pure Polyphonic path uses representation-aware scoring with `EM_VEC_ADMIT`, and the lock regression now proves the real Pure KNN route. The #947/#949 compatibility coverage is also green. Thank you for addressing those points.

I found one remaining correctness blocker in the linear `BeamMemory.recall()` route:

`_classify_vec_store_regime()` returns `unknown` when `PRAGMA user_version` cannot be read, but `recall()` only routes exact `legacy` through the conservative blob scan. `unknown` still falls through to raw-L2 KNN. That reintroduces the format-boundary risk precisely when the store format cannot be established: a relevant unnormalised/legacy row can be outside the raw-KNN candidate set and remain unrecalled.

The Polyphonic path already has the safer rule: only a confirmed `pure` regime takes the KNN route. Could the linear path use the same contract, with both `legacy` and `unknown` taking the conservative scan or a defined fallback?

Please add a regression that denies `PRAGMA user_version`, places the eligible target outside raw KNN, and verifies it is returned through public `BeamMemory.recall()`.

One separate note: the current int8 `sqrt()` feedback does not appear actionable. Direct sqlite-vec probing shows its returned distance is L2 rather than squared L2, so retaining `sqrt()` is unit-consistent.

With the `unknown` routing fix and a fresh test run on the new head, I will be happy to take another look.

linouxis9:
@dplush Fixed in `5e5573f`.

Linear `BeamMemory.recall()` now uses raw-L2 KNN only for a confirmed `pure` regime. Both `legacy` and `unknown` route through the conservative exact-cosine scan; the unknown case also emits an explicit marker-unreadable warning.

The public-recall regression now runs the same mid-table 8x-collinear target for both an unmarked store and a denied `PRAGMA user_version` read. It first proves raw KNN excludes the target, then requires recall to return it with `dense_score >= EM_VEC_ADMIT`. Reverting the routing condition kills the new case (`dense_score == 0.0`). The int8 `sqrt()` path is unchanged, per your note.

Fresh local validation:
- vector matrix: 133 passed
- #947/#949 compatibility matrix: 99 passed
- enhanced-cache/diagnostics: 81 passed
- full `tests/`: 4244 passed, 4 skipped, with the same 9 installed-package/entry-point environment failures reproduced on the parent control
- Ruff, diff check, and Python 3.10 AST parse passed

dplush (approved):
Approved on `5e5573f`.

The `unknown` marker-read route now correctly remains conservative: only confirmed `pure` stores use KNN, while `legacy` and `unknown` use the exact-cosine scan. The public regression proves the target lies outside raw KNN and is still recalled; the exact-head CI is green.

Non-blocking follow-ups can stay separate: report `unknown` diagnostics as `legacy_scan`, and throttle the marker-unreadable/full-scan logging.

PR #912, opened 2026-09-04
fix(consolidation): preserve UTC event chronology and atomic summary writes

linouxis9:
## Summary

Preserve source-event chronology through working-to-episodic consolidation, while keeping chronology separate from explicit event-date evidence.

- Normalize supported stored timestamps and operational cutoffs to naive UTC. Source-row chronology supplies the summary timestamp; unrelated calendar text is not inferred into structured event metadata.
- Propagate event dates only when all contributing rows provide valid, agreeing event evidence. Malformed/non-text stored values degrade safely; invalid public overrides reject before writes.
- Keep insertion, overrides, source-row claims, event emission and vector writes within the consolidation transaction. Failure must not strand claims or partially write a summary.
- An empty embedding is an absent optional vector, not an insertion failure. Summaries are stored without a vector with or without sqlite-vec. Real sleep tests verify the stored summary and source-claim completion.
- Use consistent UTC ordering/cutoffs across sleep, TTL, trimming, degradation, session operations and both Hermes auto-sleep implementations. Preserve pinning and explicit temporal fields across export/import, and handle unusable imported chronology explicitly.
- No historical-data rewrite or mandatory repair procedure. Query text-based date filtering remains outside this PR.

## Corrective coverage

The non-text stored-event regression exercises real sleep and verifies the resulting row and claims. Both providers have exact UTC cutoff assertions. The empty-embedding cases include empty one-dimensional and two-dimensional arrays and an empty list, with the sqlite-vec gate on and off. Reverting only empty-vector normalization causes six behavioral failures. Starvation fixtures use deterministic offline summarization instead of model-health polling.

## Retention correction

The keep-newest survivor query uses the same trim-normalized chronological key as TTL deletion, with no raw-timestamp null-sort term taking priority. Four real-retention cases cover padded Z/positive-offset/negative-offset/naive timestamps, expired/invalid rows, pinned and consolidated exemptions, session isolation, repeat trimming and preservation of stored bytes. Reverting only that sort fix makes all four fail. The changelog uses the release-announcement headline format and documents the chronological behavior.

## Verification at `2bcda5a`

One logical commit on independently checked upstream main `087fe43`. The committed diff remained unchanged through validation.

- Full core suite: **3651 passed, 6 skipped, 7 subtests passed in 439.45s (0:07:19)**. No test-file exclusions.
- Standalone Hermes integration suite: **465 passed, 1 skipped in 85.69s (0:01:25)**. No test-file exclusions.
- Ruff `F,RUF022` across changed Python files and `git diff --check`: passed.
- Python 3.10: 111 focused tests passed, including the four retention regressions and all seven version-regex tests. The suggested RegexFlag failure does not reproduce: the actual compiled flags are an integer.

CI, maintainer approval and production deployment remain separate gates; no production state was changed.

## Summary

- Preserves UTC chronology across working memory, episodic memory, and BEAM consolidation.
- Propagates event dates only when all source rows provide valid, matching evidence.
- Makes consolidation atomic across summaries, claims, overrides, events, and vector writes.
- Preserves summaries when embedding results are empty.
- Aligns sleep, TTL, trimming, session ordering, degradation, and Hermes auto-sleep with UTC.
- Preserves temporal metadata and pinning during export and import.
- Quarantines unusable imported timestamps and reports them through the CLI.

- Strengthens working-to-episodic promotion in the tiered memory pipeline.
- Adds timestamp, event-date, precision, and event-emission controls to `consolidate_to_episodic()`.
- Adds pinning and timestamp remediation controls to `update_working()`.
- Improves chronological retrieval and retention behavior without changing recall’s text-based date filters.
- Keeps event metadata and pin state consistent across portable export/import.
- Updates Hermes auto-sleep to use SQL-compatible naive UTC cutoffs.
- No MCP surface, sync layer, veracity policy, or benchmark methodology changes are reported.

- Keeps validation, quarantine, consolidation, event handling, and vector writes local.
- Adds no cloud service, telemetry, or remote synchronization dependency.
- Improves local data integrity during export and import.
- Does not erode the local-first design.

## Maintainability and verification

- Centralizes timestamp and event-date validation.
- Adds transaction ownership safeguards and explicit quarantine behavior.
- Adds regression coverage for chronology, consolidation, retrieval, malformed data, pinning, export/import, CLI behavior, and Hermes behavior.
- The broader SQL and consolidation changes increase review complexity.

## Architectural assessment

- UTC-normalized chronology, atomic consolidation, monotone pinning, and explicit quarantine are the right call for a tiered local-first memory architecture.
- The patch improves consolidation correctness and retrieval ordering without changing veracity policy or recall filter semantics.
- Remaining gaps include unchanged text-based recall date filters and no reported MCP, sync, or benchmark changes.


mnemosyne/core/beam.py
```diff
@@ -9475,10 +9486,16 @@ def sleep(self, dry_run: bool = False, force: bool = False) -> Dict:
                 # the claim survives -- the rows show as consolidated but
                 # without a summary. That's preferable to a phantom-summary-
                 # without-claim race the previous ordering allowed.
+                _event_ts = None  # items inherit the event date (MAX of source rows), not the run date
+                try:
+                    _event_ts = max((item["timestamp"] for item in items if item["timestamp"]), default=None)
```
dplush: Blocking: `max()` compares the serialized timestamp strings, not instants. Mixed ISO-8601 offsets therefore select the wrong latest source: `2026-04-01T00:30:00+02:00` sorts after `2026-03-31T23:00:00+00:00`, but is 22:30Z and thus earlier. That wrong value then feeds recency scoring and the new date fields. Please compare parsed UTC instants (while preserving the selected source representation if needed), and cover this through the real `sleep()` path with mixed offsets.

mnemosyne/core/beam.py
```diff
@@ -5237,6 +5238,16 @@ def consolidate_to_episodic(self, summary: str, source_wm_ids: List[str],
...
         rowid = cursor.lastrowid
 
+        # Apply the event-date override before any commit point. The vec
+        # insert below commits the transaction (commit=True default), so an
+        # UPDATE after it would leave a crash window where the row is
+        # committed with the run timestamp and event_date unset.
+        if event_timestamp:
+            cursor.execute(
+                "UPDATE episodic_memory SET timestamp = ?, event_date = ?, event_date_precision = ? WHERE id = ?",
+                (event_timestamp, str(event_timestamp)[:10], "day", memory_id),
```
dplush: Blocking: this writes a row/ingest timestamp into `event_date` and labels it `day` precision. Those are distinct semantics: working-memory already stores content-derived temporal extraction in `working_memory.event_date`, so a summary such as `On 2020-01-02, ...` written in 2026 will still be stored as `event_date=2026-...`. Please decide and document an event-date aggregation rule separately (for example, preserve only a consistent source event date; otherwise leave it unknown). Timestamp propagation must not manufacture an event date.

dplush (changes requested):

Preserving a summary's source-row time can be useful, but this patch currently conflates two different contracts: row/ingest time (`timestamp`) and content-derived event time (`event_date`). It also compares source timestamps lexicographically. Please define the aggregation semantics separately, use chronological parsing for source-time selection, and add end-to-end regressions through `sleep()` before merging.

linouxis9:
Thank you so much for your fast reviews on my PRs! I'll proceed with the suggested changes, thanks!

linouxis9:
Both contracts are separated as requested:

1. timestamp (row/ingest semantics) — sleep() selects the chronologically latest source timestamp by parsed UTC instants (_parse_iso_datetime_utc, naive-as-UTC; your mixed-offset repro is a pinned regression: 2026-04-01T00:30:00+02:00 vs 2026-03-31T23:00:00+00:00 selects the latter). The winner is stored naive-UTC normalized (offset stripped after conversion) because recall's date filters compare episodic_memory.timestamp lexicographically — an offset-bearing string sorts wrong on both boundaries of its day, and even +00:00 breaks parity with the naive forms other producers write.
2. event_date (content-derived semantics) — never manufactured from ingest time. sleep() propagates a content date only when every row of the group is dated (whitespace-only counts as undated) and all date strings agree; agreement is on the date string, precision carried from the newest agreeing row. A single dated row among undated ones no longer stamps the summary.

Aggregation rule documented in the docstring and inline: distinct paths through distinct kwargs (event_timestamp vs event_date/event_date_precision, appended at the signature end — AST-audited all 63 call sites, max 2 positional args anywhere).

Robustness: newest-row selection is parse-tolerant (an unparseable timestamp sorts as epoch instead of raising) — one poison row can't abort a consolidation batch whose rows are already claimed. Both UPDATEs run before the committing vector write (single transaction, SQL-trace verified).

E2E through real sleep(): mixed offsets, multi-source batches (each summary keeps only its own group's latest — the original cross-group contamination is pinned), disagreeing dates → unknown, single-dated-row → unknown, whitespace forms, poison-timestamp batch survival, precision-from-newest, normalized-storage day-boundary check. The naive-as-UTC DST-fold caveat is documented in the helper (producers write naive local wall time; timezone-aware writes proposed as follow-up).

dplush (changes requested):
Thanks for separating source-row timestamps from content-derived event dates. The UTC max selection and the direct sleep aggregation contract are improved, but I cannot approve this head because two end-to-end data paths remain incorrect.

1. Sleep still selects its candidate rows with lexicographic timestamp comparisons (`timestamp < cutoff`, `ORDER BY timestamp`). The later UTC-aware aggregation cannot repair rows that were incorrectly excluded or included by this SQL boundary. For example, at a UTC cutoff of `2026-04-01T00:00:00`, `2026-04-01T00:30:00+02:00` is `2026-03-31T22:30:00Z` and should be eligible, but the TEXT comparison rejects it. Conversely, `2026-03-31T23:30:00-02:00` is newer than the cutoff but sorts before it. Please make TTL selection and ordering chronological too, with positive and negative offset regressions at the cutoff.

2. Working-memory `event_date` and `event_date_precision` are still omitted by export/import. Those fields are the source information that sleep now aggregates, so a backup/restore silently drops content-derived dates before a later consolidation. Please round-trip both fields for working memory as well as episodic memory, and cover restore followed by sleep.

Two additional correctness gaps should be handled in the same focused pass: validate public `event_timestamp`, `event_date`, and precision overrides before persisting them, and keep the insert plus metadata overrides transactionally atomic so an override binding failure cannot leave a later accidental commit to persist a partial row.

The focused UTC max tests and cache-free CI are green, but they do not cover these write and selection boundaries.

linouxis9:
@dplush

All four findings are addressed at HEAD (`2b7ac34`, CI green across the matrix).

**1. TTL selection and ordering are chronological everywhere.** sleep()'s batch SELECT and ORDER BY, `_count_unconsolidated_before` (the auto-sleep gate), `sleep_all_sessions`' gating and session ordering, and `_trim_working_memory`'s DELETE **and** keep-newest survivor subquery all compare `datetime(timestamp)`-normalized UTC values. Your two examples are pinned regressions through real `sleep()` with a frozen clock: `2026-04-01T00:30:00+02:00` is eligible and `2026-03-31T23:30:00-02:00` is excluded at a `2026-04-01T00:00:00` cutoff. Notes on the edges:

- Unparseable timestamps degrade to the epoch — selectable, never stranded, matching the Python-side `_row_sort_key` contract. That includes SQLite's `datetime()` keyword forms: a stored `'now'` reads the live clock and would be immortal, so those are explicitly epoch-degraded too.
- Cutoffs are normalized in Python and bound as plain `YYYY-MM-DD HH:MM:SS` strings — deliberately not `datetime(?)` on the parameter: the force-consolidation sentinel (`9999-12-31T23:59:59.999999`) overflows SQL-side julian-day parsing to NULL on some SQLite builds, which silently made force a no-op (caught on CI; regression added).
- Within-second ordering ties are unspecified (`datetime()` truncates microseconds); documented.
- Trim exempts pinned rows, same as sleep — restored pinned rows are no longer destroyed at TTL with no summary.

**2. Working-memory export/import round-trips the event fields — and pinned.** `event_date` + `event_date_precision` (and `pinned`, which feeds the same sleep/trim predicates as `consolidated_at`, so losing it changes post-restore consolidation behavior the same way) survive export/import for working memory; the portable-export completeness manifest catalogs them as carried. Restore-then-sleep is an E2E test: a restored row's agreed content date lands in the episodic summary. Remaining known export gaps, named rather than smuggled: trust_tier, validator/validated_at, channel_id, author_id/type, temporal_tags.

**3. Overrides validated at the public boundary, before any work.** `consolidate_to_episodic` validates at the top — before the embed call, so a bad override doesn't pay a 30s network round-trip first: `event_timestamp` must parse as ISO-8601, `event_date` must be strict ASCII `YYYY-MM-DD` (strptime happily accepts `2020-1-2` and Arabic-Indic digit dates, both of which corrupt the lexicographic date filters downstream), precision must be in {day, month, year, unknown} — each raises ValueError. Offset-bearing `event_timestamp` values are stored **naive-UTC normalized** (the same lexicographic-filter contract sleep's own selection enforces). sleep() itself sanitizes its aggregated values before calling: garbage stored dates degrade to an undated group + warning instead of raising — a raise there would land after the claim commits and strand the group's content, which is strictly worse than the bug being fixed.

**4. Insert + overrides + vector write are one transaction.** Wrapped in the existing `_guarded_transaction` with `_vec_insert(commit=False)`: any failure between INSERT and commit rolls the whole episodic row back (failure-injection test asserts the table stays empty). The trailing commit only fires when the method owns the transaction — a caller with an open transaction keeps ownership; the graph/veracity enrichment helpers that commit internally are skipped under a caller-owned transaction (enrichment is derived and rebuildable, and sleep's real path commits its claim before calling, so production always enriches).

One deliberate scope line, same as before: recall's `from_date`/`to_date` filters remain TEXT comparisons — same class, but read-path and separate from this consolidation fix; happy to take it as its own change if you prefer.

mnemosyne/core/beam.py
```diff
@@ -4583,17 +4652,28 @@ def _trim_working_memory(self):
         the additive promise expires at WORKING_MEMORY_TTL_HOURS and
         the experiment Arm B's "ADD-only" guarantee collapses at 24h.
         """
-        cutoff = (datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS)).isoformat()
+        cutoff = _utc_cutoff_sql(
+            (datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS)).isoformat()
```
dplush: **[P1] Build this cutoff in UTC, not local wall time.** `_utc_cutoff_sql()` deliberately treats a naive input as UTC, but `datetime.now()` here is local. On a CEST host, a local `12:00` cutoff is compared as `12:00Z` instead of `10:00Z`, so an offset-bearing `10:30Z` row is deleted as old even though it is newer than the real cutoff. The same construction exists in `sleep()`. Use `datetime.now(timezone.utc)` before normalizing and add a non-UTC-host regression; this is premature working-memory loss.

mnemosyne/core/beam.py
```diff
@@ -4583,17 +4652,28 @@ def _trim_working_memory(self):
...
+        # Pinned rows are exempt from trim (matching sleep()'s exemption):
+        # they are neither TTL-deleted nor displaced from the survivor set.
         self.conn.execute("""
             DELETE FROM working_memory
             WHERE session_id = ?
               AND consolidated_at IS NULL
+              AND (pinned IS NULL OR pinned = 0)
               AND (
-                timestamp < ? OR
+                CASE WHEN lower(trim(timestamp)) IN ('now', 'subsec') OR datetime(timestamp) IS NULL THEN '1970-01-01 00:00:00' ELSE datetime(timestamp) END < ? OR
```
dplush: **[P1] Do not turn accepted imported timestamps into immediate deletion candidates.** `import_from_dict()` accepts `timestamp=None`/unparseable values, while this predicate maps all of them to the epoch. A direct reproduction imported one valid-looking working row with `timestamp=None`, then `_trim_working_memory()` deleted it (`1 → 0`) without an import error or quarantine path. Validate/reject timestamps at import, or preserve malformed legacy rows without treating them as expired; the current combination silently destroys accepted data.

mnemosyne/core/beam.py
```diff
@@ -5227,34 +5352,70 @@ def consolidate_to_episodic(self, summary: str, source_wm_ids: List[str],
...
-        rowid = cursor.lastrowid
+        # The episodic row, its metadata overrides and the vector write form
+        # one transaction: a failure anywhere between the INSERT and the
+        # commit rolls the whole row back (no partial episodic row can
+        # persist from this path). _vec_insert runs with commit=False so
+        # the committing step is the guarded transaction's own. A caller
+        # with an already-open transaction keeps ownership: the guarded
+        # block takes a savepoint, and the trailing commit fires only when
+        # this method opened the transaction itself.
+        _owned_txn = not self.conn.in_transaction
```
dplush, line 5569: **[P1] Do not emit `MEMORY_CONSOLIDATED` before a caller-owned transaction commits.** When `_owned_txn` is false, this method intentionally leaves the outer transaction open, but it still emits the event here. Repro: begin an outer transaction, consolidate, then roll back; the episodic row is gone while the event was emitted (`event_before_commit=1`, `row_after_rollback=0`). Defer the event to a successful outer commit (or reject this call inside caller-owned transactions) so consumers never observe a phantom summary.

mnemosyne/core/beam.py
```diff
@@ -5188,6 +5285,34 @@ def consolidate_to_episodic(self, summary: str, source_wm_ids: List[str],
...
+                    f"event_timestamp not a parseable ISO-8601 datetime: "
+                    f"{event_timestamp!r}"
+                ) from None
+        if event_date:
+            # Strict ASCII YYYY-MM-DD: non-padded forms (2020-1-2) and
+            # non-ASCII digits (Arabic-Indic ٢٠٢٦) both parse via
+            # strptime/fromisoformat but sort incorrectly under the
+            # lexicographic date filters used elsewhere.
+            import re as _re
+            if not _re.fullmatch(r"[0-9]{4}-[0-9]{2}-[0-9]{2}", event_date):
```
dplush: **[P1] Validate calendar dates, not only their shape, and apply the same validation on import.** `2026-02-31` passes this regex and is persisted as `('2026-02-31', 'day')`; import also writes both event fields without any validation. These values are indexed, filterable, and round-trip through export, so this creates durable impossible temporal data. Use an ASCII shape check plus `date.fromisoformat`, and reject or safely sanitize invalid imported values/precisions.

dplush (changes requested):
The new timestamp/event-date coverage is broad, but independent Builder and Sol reviews found durable correctness regressions in UTC cutoffs, import handling, transaction events, and event-date validation. These must be fixed before this data-integrity change can merge.

linouxis9:
All four findings are addressed at HEAD (`68fde23`, full suite green, CI-faithful env). HEAD also covers your Sep-5 review (four P1s from the Builder/Sol reviews) — details per finding:

**1. TTL selection and ordering are chronological everywhere — and the cutoffs are now UTC per your P1.** sleep()'s batch SELECT and ORDER BY, `_count_unconsolidated_before` (the auto-sleep gate), `sleep_all_sessions`' gating and session ordering, and `_trim_working_memory`'s DELETE and keep-newest survivor subquery all compare `datetime(timestamp)`-normalized UTC values, and the cutoff constructors now use `datetime.now(timezone.utc)` — including the same construction in `sleep()` and `sleep_all_sessions()` that you flagged, plus the tier cutoffs in `degrade_episodic` (whose `created_at` column is UTC by schema default, so a UTC cutoff is exact there). Your repro is pinned through real `sleep()` with a frozen clock: `2026-04-01T00:30:00+02:00` is eligible and `2026-03-31T23:30:00-02:00` is excluded at a `2026-04-01T00:00:00` cutoff, with a non-UTC-host regression. Notes on the edges:

- Unparseable timestamps degrade to the epoch — selectable, never stranded. That includes SQLite `datetime()` keyword forms: a stored `'now'` would be immortal, so those are explicitly epoch-degraded too.
- Cutoffs are normalized in Python and bound as plain `YYYY-MM-DD HH:MM:SS` strings — deliberately not `datetime(?)` on the parameter: the force-consolidation sentinel (`9999-12-31T23:59:59.999999`) overflows SQL-side julian-day parsing to NULL on some SQLite builds, which silently made force a no-op (caught on CI; regression added).
- Within-second ordering ties are unspecified (`datetime()` truncates microseconds); documented.
- Trim exempts pinned rows, same as sleep.

**2. Import validation for unusable timestamps.** `import_from_dict` (the documented restore path) now rejects working-memory rows whose `timestamp` is None/blank/unparseable: the row is skipped, counted machine-readably in the stats (`skipped` + `skipped_bad_timestamp`), and warned. Rationale: such a row cannot be placed on the timeline, and the old epoch-degradation made it an immediate trim candidate — silent data loss on restore. Valid-row round-trips are byte-identical (regression covers it); legacy rows already in stores are unaffected.

**3. Event emission under a caller-owned transaction.** `consolidate_to_episodic` now raises `MemoryTransactionStateError` at method entry — before the INSERT and before the embed network call — when emission is requested while a caller owns the transaction, with an explicit `emit_event=False` opt-out for callers that consolidate inside their own transaction. The raise is at entry precisely so a catching caller never sees error + persisted row + an unreachable-by-retry summary (the memory id is deterministic). Callers without an emitter are unaffected (nothing observable changes for them). Known trade-off, stated: enrichment (graph/veracity) is already skipped under caller-owned transactions — derived, rebuildable data. A deferred-event mechanism (connection-level pending-event queue) is a possible follow-up; happy to take it as its own change.

**4. `event_date` is calendar-validated, and import sanitizes.** The ASCII shape check is now backed by `date.fromisoformat` — `2026-02-31` raises. Precision must be in {day, month, year, unknown}. Import-side: rows with invalid `event_date` are imported with the field sanitized to undated + a warning (the row survives; only the derived field is reset — distinct from the timestamp policy in (2), where the row's chronology itself is unknowable). Export/import round-trips of valid dates are pinned.

**5. Working-memory export/import round-trips the event fields — and pinned.** `event_date`, `event_date_precision`, and `pinned` survive export/import; restore-then-sleep is an E2E test (a restored row's agreed content date lands in the episodic summary). Remaining known export gaps, named rather than smuggled: trust_tier, validator/validated_at, channel_id, author_id/type, temporal_tags.

**6. One deliberate scope line:** recall's `from_date`/`to_date` filters remain TEXT comparisons — read-path, wall-date semantics, and separate from this consolidation fix; happy to take it as its own change if you prefer.

mnemosyne/core/beam.py
```diff
@@ -9889,6 +10215,17 @@ def import_from_dict(self, data: Dict, force: bool = False) -> Dict:
         # -- Working memory --
         for item in data.get("working_memory", []):
             mid = item.get("id")
+            if not _import_timestamp_ok(item.get("timestamp")):
```
dplush, line 10474: **[P1] Do not make a legacy backup restore silently lossy.** `timestamp` is nullable in existing schemas and older exports can contain it as `null`. This new gate skips the entire working-memory row rather than preserving it safely or failing the import before partial restore. Repro: export a schema-valid row with `timestamp=NULL`, then restore it here: result is `skipped_bad_timestamp=['legacy-null']` and `restored_count=0`. The old trim behavior was unsafe, but replacing it with a silent restore drop still loses user memory. Preserve with an explicit safe timestamp/quarantine policy, or reject atomically with a clear recovery path.

mnemosyne/core/beam.py
```diff
@@ -9934,6 +10272,12 @@ def import_from_dict(self, data: Dict, force: bool = False) -> Dict:
                 # cycle on the importing DB processes them normally.
                 item.get("consolidated_at"),
                 item.get("consolidation_claimed_at"),
+                # content-derived event date survives backup/restore so a
+                # restored row can still be consolidated with its date;
+                # pinned rows keep their consolidation/trim exemption.
+                item.get("event_date"),
```
dplush: **[P1] Validate imported event dates before persistence.** Public consolidation validates calendar reality, but imports write this field raw. Two imported old rows with `event_date='2026-02-31'` restore successfully; `sleep()` claims and commits both rows, then `consolidate_to_episodic()` raises `ValueError`, leaving `claimed=2` and `episodic=0`. Future sleep passes skip those rows. Apply the same strict date and precision validator on both working and episodic import paths, and make sleep treat any stored invalid date defensively before it claims rows.

mnemosyne/core/beam.py
```diff
@@ -4583,17 +4671,36 @@ def _trim_working_memory(self):
         the additive promise expires at WORKING_MEMORY_TTL_HOURS and
         the experiment Arm B's "ADD-only" guarantee collapses at 24h.
         """
-        cutoff = (datetime.now() - timedelta(hours=WORKING_MEMORY_TTL_HOURS)).isoformat()
+        # dplush P1: build the cutoff at the UTC instant. datetime() has
+        # already normalized offset-bearing ROWS to UTC in this predicate;
+        # pairing them with a local-wall-as-UTC cutoff deleted aware rows
+        # up to one host-offset early. Naive-local rows keep wall-as-UTC
+        # semantics (their producers stamp naive-local by contract).
+        cutoff = _utc_cutoff_sql(
```
dplush, line 4773: **[P2] UTC cutoffs now disagree with the dominant naive-local write path.** `remember()` still stores `datetime.now().isoformat()` without an offset, while this predicate now compares it as a UTC wall time against a UTC cutoff. Under Europe/Berlin, a production-shaped row actually 85 hours old missed the 84-hour sleep cutoff by two hours (`eligible_count=0`). Either store naive UTC consistently at write time with an explicit migration/compatibility policy, or retain matching local-wall cutoff semantics for naive legacy rows.

mnemosyne/core/beam.py
```diff
@@ -1600,6 +1619,75 @@ def _parse_iso_datetime_utc(value: str) -> datetime:
     return _normalize_datetime_utc(datetime.fromisoformat(value.replace("Z", "+00:00")))
 
 
+
+# Accepted event_date_precision values (content-derived date precision).
+_EVENT_DATE_PRECISIONS = {"day", "month", "year", "unknown"}
```
dplush: **[P2] Preserve the parser's valid `week` precision through consolidation.** `temporal_parser` emits `event_date_precision='week'` for normal input such as “this week”, but this new allowlist excludes it and sleep silently persists `unknown` on the summary. Repro: working row `('2026-09-05', 'week')` becomes episodic `('2026-09-05', 'unknown')`. Include `week` in the accepted precision set (and validate it consistently at the import boundary).

dplush (changes requested):
The direct core timestamp and transaction fixes address several earlier blockers, but restore compatibility and event-date validation remain unsafe. Independent Builder and Sol reviews, plus local reproductions, confirm two P1 data-integrity blockers and two P2 timestamp/metadata regressions.

mnemosyne/core/beam.py
```diff
@@ -9475,10 +9760,79 @@ def sleep(self, dry_run: bool = False, force: bool = False) -> Dict:
...
+                        logger.warning(
+                            "sleep: group event_date %r is not YYYY-MM-DD; "
+                            "storing summary without an event date",
+                            _agg_event_date[:40],
+                        )
+                        _agg_event_date = None
+                        _agg_event_date_precision = "unknown"
+                if _agg_event_date_precision not in _EVENT_DATE_PRECISIONS:
+                    _agg_event_date_precision = "unknown"
                 self.consolidate_to_episodic(
```
dplush, line 10040: **[P1] Apply calendar validation here before passing an aggregate date after the claim commit.** Shape-valid `2026-02-31` reaches `consolidate_to_episodic()`, which raises only after `sleep()` has committed `consolidated_at` and `consolidation_claimed_at`. Repro: `ValueError: event_date is not a real calendar date: '2026-02-31'`, both claim columns set, `episodic_count=0`; subsequent sleep calls skip the row until manual reclaim. Reuse the strict calendar validator in this sanitation path and degrade invalid stored dates to no date/`unknown`, with an end-to-end regression that asserts a summary is written and no orphan claim remains.

dplush (changes requested):
The import-side calendar validation, timestamp preservation, UTC normalization, and transaction-event fixes address the prior blockers. One stored-data path still passes an impossible calendar date through the post-claim sleep sanitizer, so it can strand source rows without a summary. Independent Builder and narrow Sol gates reproduced it.

mnemosyne/core/beam.py
```diff
@@ -34,7 +34,70 @@
...
+def _import_timestamp_ok(value) -> bool:
+    """An imported row timestamp must place the row in time. None/blank/
+    unparseable values would otherwise epoch-degrade into immediate trim
+    candidates (round-4 probe: 1 valid + 1 None row -> 1 row after trim)."""
+    if value is None:
+        return False
+    if not isinstance(value, str) or not value.strip():
+        return False
+    try:
+        datetime.fromisoformat(value.strip())
```
dplush: **[P1] Accept RFC 3339 `Z` timestamps on every supported Python version.** `pyproject.toml` declares Python `>=3.10`, but Python 3.10.20 raises on `datetime.fromisoformat('2026-01-02T03:04:05Z')`. The real import path consequently marks this valid backup timestamp unusable, rewrites it to `1970-01-01T00:00:00`, sets `pinned=1`, and increments `imported_bad_timestamp`. This alters valid restored chronology and requires manual repair on a supported runtime. Reuse `_parse_iso_datetime_utc()` (which already normalizes `Z`) or translate a terminal `Z` to `+00:00` before validation, with a Python-3.10 regression test.

dplush (changes requested):
The prior invalid-string event-date claim path is fixed on this head, but the new import quarantine rejects valid RFC 3339 UTC timestamps on the supported Python 3.10 floor. Builder, a direct Python 3.10 import reproduction, and a narrow Sol gate confirm this release blocker.

linouxis9:
## Round 7 — adversarial review fixes + CodeRabbit + this review, all in @5d1ae6e

**Your P1 (RFC 3339 `Z` on Python 3.10) is fixed.** `_import_timestamp_ok()` now validates through `_parse_iso_datetime_utc()`, which normalizes a terminal `Z` to `+00:00` before `fromisoformat` — no quarantine, no epoch rewrite, no `pinned=1` on a valid UTC stamp. Regression covered by `test_round7_fixes.py::TestZSuffix::test_z_suffix_accepted`. (This review raced the push: it was filed against `b5cae48`; the same finding landed independently as CodeRabbit `5122732528`, fixed in the same commit.)

Since `2029c26`, a full blind adversarial round (audit 7) ran against the round-6 fixes. Its findings, all fixed and probe-verified at `5d1ae6e`:

**HIGH**
- **Quarantine was log-only** (the fix claimed to bind the epoch but the INSERT bound the raw junk). Now the epoch is actually bound, the row is `pinned=1`, and the episodic loop carries its own epoch guard (no pinned column there).
- **Force-import silently unpinned live rows**: the force path was DELETE+re-INSERT honoring only the payload's `pinned`; a payload without the key (older export) lifted an existing pin — including a quarantine pin. Pin is now **monotone on overwrite** (`max(existing, incoming)`); explicit unpinning is an operator action via the new `update_working(pinned=0)` / `update_working(timestamp=…)` remediation API.
- **Sleep starvation**: rows with unparseable timestamps sorted first and filled the `SLEEP_BATCH_SIZE` window, so a junk backlog made every pass `no_op` and eligible rows behind them never consolidated (repro: 5100 junk + 1 valid → `no_op`, forever). The SQL predicate now **excludes unplaceable non-sentinel timestamps before the LIMIT** (`now`/`subsec` sentinels remain claimable), and `no_op` reports a `filtered_unplaceable` count so a poisoned backlog is visible instead of silent.

**MEDIUM**
- No in-repo path to unpin/re-date quarantined rows → `update_working()` gained `pinned=` and `timestamp=` kwargs (timestamp validated; raises `ValueError` on garbage).

**LOW / hygiene**
- `get_context()`'s `last_recalled` bump and `health()`'s staleness `now` were naive-LOCAL against naive-UTC writers (verified +14h jump on a UTC+14 host) → UTC-stripped.
- `reclaim_orphans(stale_after_seconds=…)` unbounded → clamped to `[0, 10**9]`, non-finite rejected with `ValueError`.
- CodeRabbit `5122732528`: `cmd_import` now formats working/episodic summaries through `_format_store_stats()` (quarantine counts surface in the CLI); episodic quarantine increments `imported_bad_timestamp` too; sanitizer hardened (truthy non-string `event_date` and non-string precision degrade to undated/unknown instead of raising `TypeError` mid-file with a dangling transaction).

**Verified**
- Full suite `3560 passed / 0 failed`; `tests/test_beam_event_timestamps.py` 48/48; new `tests/test_round7_fixes.py` (13 tests incl. starvation, monotone pin, sanitizer no-raise, Z-suffix, reclaim clamp, mid-file poison import); ruff clean. The starvation + sanitizer + unpin findings were re-probed live by me before fixing, matching the auditor's probes.

linouxis9:
Thanks for the review — all five comments are addressed on `9cd4296`:

- **Eligibility predicates vs sleep()'s claim filter** (both sites): `_count_unconsolidated_before` and `sleep_all_sessions` now carry the same placeability predicate as the claim query, and it is tightened to match the Python acceptance gate exactly: `datetime(trim(timestamp)) IS NOT NULL AND date(substr(trim(timestamp),1,10)) = substr(trim(timestamp),1,10)` (sentinels via the `'now'`/`'subsec'` IN-list). The date-prefix roundtrip excludes julian-day/numeric forms (`2461000`, `45000`) *and* calendar-impossible dates (`2026-02-30`, which SQLite rolls over instead of rejecting) — both previously counted as eligible forever while never being claimed. `sleep_all_sessions` additionally gained the pinned filter its sibling gates already had, so import-quarantine rows (epoch + pinned=1) can no longer make a session look eligible on every maintenance pass.
- **Import normalization** (both paths): accepted offset/`Z` timestamps are now converted to naive UTC before binding (with `.strip()` before parsing, matching the acceptance guard), so recall's lexicographic date filters and the UTC cutoffs stay aligned; the epoch-quarantine branch is unchanged.
- **Overflow guard**: the `get_context` bump now catches `AttributeError` alongside `ValueError`/`TypeError`/`OverflowError` — numeric-looking `last_recalled` values are converted by SQLite's NUMERIC affinity on storage and previously crashed the hot path.
- **Tests strengthened per the comments**: exact naive-UTC cutoff assertion with a frozen aware clock (UTC+2 host shape) in the root provider gate test, plus a twin test for the packaged `mnemosyne_hermes` gate; five weak assertions tightened (schema-default precision, production cutoff shape, `SLEEP_BATCH_SIZE`-derived junk volumes, reclaim clamp exact values + cutoff parse); the poison test now seeds a value that genuinely exercises the intended path.

Two adversarial rounds (12 + 13) verified the delta: all fixes held under mutation (14 mutation trees; eligibility gates now byte-identical across all four sites and agree with the claim on a 43-row mixed corpus; the overflow guard closed over int/float/bool/numeric-string/blob/NaN/empty). Residuals are LOW and reachable only through raw/legacy/foreign DB writes (year-0000, `T24:00`, lowercase-`z` suffixes — families that neither import nor any producer can create, since both quarantine-normalize at the boundary); three test-gap residuals (julian exclusion at the claim site with a 2026-dated real row, episodic padded-import strip, count-gate pinned filter) are noted for a follow-up round.

linouxis9:
Thanks — all five comments are addressed on `bce7343`:

- **One shared predicate definition**: the placeability predicate and the chronological CASE expression are now module constants (`_SQL_PLACEABLE_TS`, `_SQL_CHRONO_TS`) interpolated at every site — trim's DELETE, `_count_unconsolidated_before`, sleep's claim + ORDER BY, the blocked-counter inverse, and `sleep_all_sessions` eligibility + MIN ordering. The literal text no longer exists anywhere else in the module, so the drift class that produced three separate starvation/miscount bugs is structurally closed.
- **`update_working` invalidates the enhanced-recall cache** after its commit when a row actually changed — a remediated quarantine row (timestamp/pinned) can no longer leave a warmed entry with the pre-remediation result set.
- **`consolidate_to_episodic` defers invalidation under caller-owned transactions**: the `finally` now gates on the same `_owned_txn` flag as the commit, so a concurrent reader cannot repopulate the cache from the pre-consolidation state between our (suppressed) invalidation and the caller's later commit. The caller-owned contract is documented at the gate.
- **Pinned-quarantine visibility**: sleep's `no_op` now reports `pinned_exempt` (with an actionable message) when the only remaining unconsolidated rows are pinned imports, and `get_working_stats` exposes `pinned_unconsolidated` — a growing quarantine set from a corrupted backup is now observable instead of invisible.
- **One event-date validator**: the shape+calendar check is extracted to `_event_date_valid()` and used by the import sanitizer, the public `consolidate_to_episodic` validator (raise path) and sleep's aggregation degrade — three identical copies reduced to one definition.

The four-placeability-site SQL is now byte-identical by construction (constants); new pinning tests cover the count-gate pinned exclusion, the newer-real-row-behind-julian starvation shape, the episodic padded-import strip, and the deferred-invalidation contract; all are mutation-verified. Full suite 3581 passed / 0 failed; ruff clean.

dplush:
Thanks for the thorough follow-up. The prior fixes are moving in the right direction, but one P1 data-integrity path remains.

In `sleep()`, stored `event_date` and `event_date_precision` are `.strip()`ed before their types are validated. A legacy or externally written non-text value, such as a BLOB, raises after the source rows have already been claimed and committed. The result is stranded working-memory rows with `consolidated_at` / `consolidation_claimed_at` set but no episodic summary.

Please sanitize non-string stored values before aggregation and degrade them to an undated summary with `unknown` precision rather than raising. Please add a regression through the real `sleep()` path that verifies the summary is written and the claim markers are cleared.

In the same pass, please make the public override validation consistently raise `ValueError` for non-string `event_date` values, rather than leaking `TypeError`.

Once that is addressed, please rerun the focused matrix and request a fresh review on the new head.

linouxis9:
Thanks — the P1 path is closed:

**Non-string stored `event_date`/`event_date_precision` no longer raise post-claim.** sleep()'s aggregation now guards every strip site with an isinstance check: a BLOB/number stored value degrades to undated (`unknown` precision) instead of reaching `_event_date_valid(bytes)` after the claim commit. The same class was also fixed on the sibling column `valid_until` (its regex check sat outside the try — a BLOB there stranded the group identically; now mapped to no-expiry).

Regressions through the real `sleep()` path: same-BLOB event_date on all rows → summary written undated/unknown, every row claimed and summarized, zero stranding; BLOB precision on agreeing dates → date kept, precision `unknown`; BLOB `valid_until` → group consolidates. All mutation-verified (a revert of each guard kills its test).

**Public override validation now raises `ValueError` consistently**: any non-string `event_date` (including falsy non-strings `b""`, `0`, `[]`, `{}`) raises `ValueError` — never the `TypeError` from the shape regex — and `None` alone remains "not provided"; `event_timestamp` and `event_date_precision` got the same isinstance-first treatment. Witness tests for each corner.

An independent adversarial pass re-ran the whole matrix on real CPython 3.10 and 3.11; suites green; ruff clean.

@coderabbitai resume

dplush (approved):
Re-reviewed on `2bcda5a`. The UTC chronology, transaction boundary, timestamp quarantine, and Hermes auto-sleep concerns are addressed. The focused checks and full CI matrix are green. I have no further changes requested. Approving.

PR #913, opened 2026-09-04
fix(recall): configurable content cap with unchanged 500-character default

linouxis9:
## Summary

Make the per-result recall `content` limit configurable while keeping the shipped **500-character default**.

- `MNEMOSYNE_RECALL_CONTENT_CAP` accepts a positive integer at import time. Unset, empty, invalid and non-positive values fall back to 500.
- Apply the cap at every public recall boundary, including direct `fact_recall()`, linear, polyphonic, fact-merge and explain results.
- Cap enhanced associative additions before cache persistence/return and enforce the boundary on cache hits.
- Include the cap in enhanced-recall cache identity. This branch increments current main's cache version from **7 to 8**.
- Leave stored records unchanged; this controls returned `content`, not full-record inspection or total response bytes.

For example, `MNEMOSYNE_RECALL_CONTENT_CAP=4000` explicitly opts into a larger limit. It is **not** the default.

## Maintainer follow-up

- Direct `BeamMemory.fact_recall()` now applies the shared cap after score ordering and `top_k`, covering both raw and consolidated fact stores.
- Direct over-cap regressions cover both stores; SHMR's automatic fact retrieval is verified not to receive the uncapped tail.
- The shipped-default mutation test copies only `mnemosyne/`, excludes developer `.venv` data, and verifies the scratch import path.
- No vector-admission environment scrub or cross-PR version coordination. Zero is covered through direct parsing and a fresh subprocess import.

## Current revision

Based on main `8df6af2` after #917 merged. Main has cache version 7; this PR uses **8**. Persistent-cache regressions reject entries from versions 6 and 7 while retaining cold/warm cache-hit behavior and the opaque `v2:` key prefix.

The rebase reuses main's positive-integer parser unchanged. Conflict budgets, sleep atomicity and postcommit cache behavior remain unchanged. The public default is still 500.

## Verification at `d692abe`

One logical commit on main `8df6af2`, validated with unchanged source/tree/diff.

- Full core suite: **4060 passed, 6 skipped, 7 subtests passed in 495.98s (0:08:15)**.
- Full provider suite: **465 passed, 1 skipped in 85.64s (0:01:25)**.
- Real Python 3.10 focused checks: **175 passed in 47.20s**.
- Focused cap/cache/conflict/fact/SHMR battery: **177 passed**.
- Nine independent behavioral mutations caught, including direct fact output, both enhanced-cache guards, copy scope, zero/default parsing, cap identity, missing version bump and missing version hashing; controls passed.
- Ruff `F,RUF022` and `git diff --check`: passed. No test-file exclusions; environmental skips remain explicit in receipts.

GitHub CI, CLA and maintainer approval are separate exact-head gates. No deployment, historical rewrite or release was performed.

## Summary

Adds configurable recall-result content capping through `MNEMOSYNE_RECALL_CONTENT_CAP`.

- Retains the shipped 500-character default.
- Falls back to 500 for unset, empty, invalid, or non-positive values.
- Applies the cap across BEAM, polyphonic, fact-merge, explain, enhanced associative, direct `fact_recall()`, SHMR, mapper, fetch, and cache-hit paths.
- Includes the resolved cap in enhanced-recall cache identity.
- Increases the enhanced-recall cache version to 9.
- Leaves stored records unchanged.

## Architectural impact

This change strengthens the BEAM retrieval boundary across working-memory, episodic, and fact-backed results. It does not change tier selection, retrieval strategies, consolidation, veracity processing, or the sync layer.

The cap limits content exposed to callers and SHMR agent prompts without changing persisted memory. This improves privacy and preserves the local-first design. No remote service, telemetry, or data-sharing path is added.

Hermes, MCP, and CLI surfaces retain their existing contracts. They receive bounded content through existing core recall paths.

Cache isolation prevents results created under one cap from being served under another cap. The cache-version bump rejects older enhanced-recall generations. This is the right call for correctness, privacy, and maintainability.

Tests cover configuration validation, all recall boundaries, cache identity and persistence, direct fact recall, SHMR retrieval, cache invalidation, and protection against accidental changes to the shipped default. Benchmark methodology remains unchanged.

### DIFF
--- a/mnemosyne/core/beam.py
+++ b/mnemosyne/core/beam.py
@@ -380,6 +380,58 @@ WM_PINNED_IDS = set(
     if pid.strip()
 )
 EPISODIC_RECALL_LIMIT = int(os.environ.get("MNEMOSYNE_EP_LIMIT", "50000"))
+
+def _env_int(name: str, default: int) -> int:
+    """Parse a positive integer knob, falling back on empty/invalid values."""
+    raw = os.environ.get(name, "").strip()
+    if not raw:
+        return default
+    try:
+        value = int(raw)
+        if value > 0:
+            return value
+    except ValueError:
+        pass
+    logger.warning("%s is not a positive integer; using default %s", name, default)
+    return default
+
+
+# Recall content cap: bounded by default at 500 chars per result to keep
+# the public recall contract stable (see #685). MNEMOSYNE_RECALL_CONTENT_CAP
+# opts into a higher limit; resolved once at IMPORT time (restart to
+# change). Empty/unset falls back silently, invalid or non-positive
+# values warn and fall back to 500. The cap is enforced at the shared
+# public-result boundary of recall() (see _cap_recall_results), covering
+# the linear producers, the polyphonic engine path, the MEMORIA
+# supplements and the fact-voice merge alike.
+RECALL_CONTENT_CAP = _env_int("MNEMOSYNE_RECALL_CONTENT_CAP", 500)
+
+
+def _cap_recall_content(content: Any) -> Any:
+    """Apply the public recall content cap to one result's content field."""
+    if isinstance(content, str) and len(content) > RECALL_CONTENT_CAP:
+        return content[:RECALL_CONTENT_CAP]
+    return content
+
+
+def _cap_recall_results(results: Any) -> Any:
+    """Shared public-result boundary for the recall content cap.
+
+    Enforces RECALL_CONTENT_CAP on every result leaving recall() or
+    recall_enhanced(), including associative additions and cache hits, whatever
+    producer built it: the six linear slices, the polyphonic engine mapper,
+    the MEMORIA structured-fact supplements (linear + polyphonic) and the
+    fact-voice merge. Feature-gated paths are exactly the ones that tend to
+    re-grow raw-content leaks, so capping at the boundary (not per producer)
+    keeps the #685 bounded-payload contract true by construction.
+    """
+    if not isinstance(results, list):
+        return results
+    for r in results:
+        if isinstance(r, dict) and "content" in r:
+            r["content"] = _cap_recall_content(r["content"])
+    return results
+
 SLEEP_BATCH_SIZE = int(os.environ.get("MNEMOSYNE_SLEEP_BATCH", "5000"))
 SCRATCHPAD_MAX_ITEMS = int(os.environ.get("MNEMOSYNE_SP_MAX", "1000"))
 RECENCY_HALFLIFE_HOURS = float(os.environ.get("MNEMOSYNE_RECENCY_HALFLIFE", "168"))  # 1 week default
@@ -445,21 +497,6 @@ def _env_float(name: str, default: float) -> float:
             name, raw[:80], default,
         )
         return default
-
-
-def _env_int(name: str, default: int) -> int:
-    """Parse a positive integer knob, falling back on empty/invalid values."""
-    raw = os.environ.get(name, "").strip()
-    if not raw:
-        return default
-    try:
-        value = int(raw)
-        if value > 0:
-            return value
-    except ValueError:
-        pass
-    logger.warning("%s is not a positive integer; using default %s", name, default)
-    return default
 
 
 # Conflict validation limits apply to one complete sleep(), not each source.
@@ -7172,13 +7209,13 @@ class BeamMemory:
                     "query": query,
                     "top_k": top_k,
                     "engine": "polyphonic",
-                    "results": poly_results,
+                    "results": _cap_recall_results(poly_results),
                     "explain": {
                         "unsupported": True,
                         "reason": "polyphonic recall explain is not implemented in v1; inspect per-result voice_scores instead.",
                     },
                 }
-            return poly_results
+            return _cap_recall_results(poly_results)
 
         results = []
         query_lower = query.lower()
@@ -7509,7 +7546,7 @@ class BeamMemory:
                 _track_literal_content("working", row["id"], row["content"])
                 results.append({
                     "id": row["id"],
-                    "content": row["content"][:500],
+                    "content": row["content"][:RECALL_CONTENT_CAP],
                     "source": row["source"],
                     "timestamp": row["timestamp"],
                     "tier": "working",
@@ -7627,7 +7664,7 @@ class BeamMemory:
                     _track_literal_content("working", row["id"], row["content"])
                     results.append({
                         "id": row["id"],
-                        "content": row["content"][:500],
+                        "content": row["content"][:RECALL_CONTENT_CAP],
                         "source": row["source"],
                         "timestamp": row["timestamp"],
                         "tier": "working",
@@ -7689,7 +7726,7 @@ class BeamMemory:
                     _track_literal_content("episodic", row["id"], row["content"])
                     results.append({
                         "id": row["id"],
-                        "content": row["content"][:500],
+                        "content": row["content"][:RECALL_CONTENT_CAP],
                         "source": row["source"],
                         "timestamp": row["timestamp"],
                         "tier": "episodic",
@@ -7916,7 +7953,7 @@ class BeamMemory:
             _track_literal_content("episodic", row["id"], row["content"])
             results.append({
                 "id": row["id"],
-                "content": row["content"][:500],
+                "content": row["content"][:RECALL_CONTENT_CAP],
                 "source": row["source"],
                 "timestamp": row["timestamp"],
                 "tier": "episodic",
@@ -8024,7 +8061,7 @@ class BeamMemory:
                     _track_literal_content("episodic", row["id"], row["content"])
                     results.append({
                         "id": row["id"],
-                        "content": row["content"][:500],
+                        "content": row["content"][:RECALL_CONTENT_CAP],
                         "source": row["source"],
                         "timestamp": row["timestamp"],
                         "tier": "episodic",
@@ -8194,7 +8231,7 @@ class BeamMemory:
                             )
                             results.append({
                                 "id": memoria_source_id,
-                                "content": _row["content"][:500],
+                                "content": _row["content"][:RECALL_CONTENT_CAP],
                                 "source": _row["source"],
                                 "timestamp": _row["timestamp"],
                                 "tier": "memoria_source",
@@ -8356,17 +8393,17 @@ class BeamMemory:
                 "query": query,
                 "top_k": top_k,
                 "engine": "linear",
-                "results": final_results,
+                "results": _cap_recall_results(final_results),
                 "explain": _explain_trace.to_dict(),
             }
 
-        return final_results
+        return _cap_recall_results(final_results)
 
     # Bump whenever the enhanced-recall candidate or ranking algorithm changes
     # so entries cached under an older digest are not reused. Part of the
     # hashed payload; the opaque key keeps the "v2:" prefix because QueryCache's
     # opaque-path recognition (_OPAQUE_V2_KEY_RE) keys off that prefix.
-    _ENHANCED_RECALL_CACHE_VERSION = 7
+    _ENHANCED_RECALL_CACHE_VERSION = 8
 
     def _enhanced_recall_cache_key(
         self,
@@ -8475,6 +8512,7 @@ class BeamMemory:
                 "synonym_module": expand_query is not None,
                 "associative_graph": self.episodic_graph is not None,
                 "embeddings_available": _embeddings.available(),
+                "recall_content_cap": RECALL_CONTENT_CAP,
                 "embedding_model": getattr(_embeddings, "_DEFAULT_MODEL", None),
                 "embedding_dimension": getattr(_embeddings, "EMBEDDING_DIM", None),
                 "embedding_query_prefix": os.environ.get("MNEMOSYNE_EMBEDDING_QUERY_PREFIX", ""),
@@ -8619,7 +8657,7 @@ class BeamMemory:
             # ``top_k`` is part of the v2 digest, so truncating here would
             # make a hit differ from the cached pipeline result (notably when
             # associative retrieval appends related memories after top-k).
-            return cached
+            return _cap_recall_results(cached)
 
         # 4. Run base recall with expanded query
         results = self.recall(
@@ -8714,7 +8752,10 @@ class BeamMemory:
             except Exception:
                 logger.info("Regex extraction failed, skipping", exc_info=True)
 
-        # 9. Cache results
+        # 9. Cap after all augmentation, before persistence and public return.
+        results = _cap_recall_results(results)
+
+        # 10. Cache results
         if use_cache and not explain and hasattr(self, '_query_cache') and self._query_cache is not None:
             self._query_cache.put_opaque(cache_key, results)
 
@@ -9262,10 +9303,12 @@ class BeamMemory:
     def _polyphonic_row_to_dict(self, row, *, tier_label: str) -> Dict:
         """Shared row → recall-dict mapper. /review caught the
         near-duplicate column mapping across episodic/working
-        branches -- single helper now."""
+        branches -- single helper now. Content is capped through
+        _cap_recall_content so the polyphonic result shape obeys the
+        same public recall content bound as the linear producers."""
         d = {
             "id": row["id"],
-            "content": row["content"],
+            "content": _cap_recall_content(row["content"]),
             "source": row["source"],
             "timestamp": row["timestamp"],
             "session_id": row["session_id"] if "session_id" in row.keys() else None,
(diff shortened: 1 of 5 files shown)


mnemosyne/core/beam.py
```diff
@@ -300,6 +300,35 @@ def _default_db_path() -> Path:
...
+            "%s=%r must be positive; falling back to default %s",
+            name, raw[:80], default,
+        )
+        return default
+    return v
+
+
+# Recall content cap: 500 chars silently truncates most memories in recall
+# results (see #685). Env-gated, default 4000; validated at import.
+RECALL_CONTENT_CAP = _env_int("MNEMOSYNE_RECALL_CONTENT_CAP", 4000)
```
dplush: Blocking: this changes the result representation, but `RECALL_CONTENT_CAP` is neither in the enhanced-recall cache key nor accompanied by a cache-version bump. `recall_enhanced()` returns opaque persistent cached result payloads directly, so an entry created on the base still returns its 500-character content after upgrading to this default 4,000-character version, without running base recall. Please include the cap in the key (appropriate for an env-configurable output contract) or bump the version, and add a base/old-cap persistent-cache upgrade regression.

dplush (changes requested):

The linear recall paths consistently apply the configured cap, but this expands the default public recall payload from 500 to 4,000 characters per result across agent-facing surfaces. That conflicts with the explicit #685 direction to preserve the bounded search contract and offer full inspection through an opt-in CLI path. Please obtain or document that broader Core/MCP/provider contract decision, or narrow this to an explicit opt-in path.

Independently, the enhanced-recall cache must be invalidated for this changed result representation and covered with an upgrade regression.

dplush:
Hi @AxDSan, I am holding this PR for two reasons, not because the truncation report itself is invalid:

1. This changes the default Core recall payload from 500 to 4,000 characters per result across MCP, Hermes and other agent consumers. For #685, our proposed safe contract was to keep bounded search results by default and expose full stored records through an explicit CLI `--full` path. A broader default expansion needs an explicit Core/provider contract decision first. With top_k=40, the theoretical size of MCP, Hermes, and other agent payloads would increase from 20,000 to 160,000 characters.

2. The changed result representation is not reflected in the persistent enhanced-recall cache key or cache version. An entry written by the prior version can still be returned at 500 characters after upgrade, without rerunning recall.

If we agree to widen the default, please add the cache invalidation/key coverage and end-to-end tests for default, env override and an old-cache upgrade. Otherwise I would recommend narrowing this to the explicit inspection path discussed on #685.

linouxis9:
Narrowed to the #685 safe contract as recommended:

Default stays 500 — behavior is byte-identical to main unless MNEMOSYNE_RECALL_CONTENT_CAP is set. The env var is the opt-in (validated at import: empty falls back silently, invalid/non-positive warns and falls back); the CLI --full inspection path from #685 remains available as a follow-up on top of this knob.

Cache: RECALL_CONTENT_CAP is part of the enhanced-recall key payload (a result-representation switch), and the version is bumped — aligned at 8 with the sibling vec-admission PR (#911), so whoever merges second resolves the one-line conflict trivially. Cross-process upgrade verified: entries written under one cap are never served under another.

Tests: default-500 asserted against a fresh _env_int recompute (fails under any mutation of the shipped default), env opt-in truncating exactly at a >4000-char boundary across working and episodic sites, invalid/negative/empty env, and collection-order independence — tests/conftest.py scrubs import-resolved knobs before any test module imports beam, so an exported deployment value can't bake into module constants during collection (the ordering failure was reproduced first, then pinned dead in both orders).

Also documented in-code: the polyphonic path and fact-voice merge (both flag-gated) don't apply the cap — pre-existing on main, noted for a follow-up.

mnemosyne/core/beam.py
```diff
@@ -300,6 +300,40 @@ def _default_db_path() -> Path:
...
+
+
+# Recall content cap: bounded by default at 500 chars per result to keep
+# the public recall contract stable (see #685). MNEMOSYNE_RECALL_CONTENT_CAP
+# opts into a higher limit; resolved once at IMPORT time (restart to
+# change). Empty/unset falls back silently, invalid or non-positive
+# values warn and fall back to 500. NOTE: the polyphonic path and the
+# fact-voice merge (flag-gated) do not apply this cap — pre-existing,
+# tracked separately.
+RECALL_CONTENT_CAP = _env_int("MNEMOSYNE_RECALL_CONTENT_CAP", 500)
```
dplush, line 407: **[P1] Enforce this cap at the shared public-result boundary, not only the six linear producers.** With `MNEMOSYNE_POLYPHONIC_RECALL=1`, `recall()` returns `_recall_polyphonic()` before all six changed slices and its mapper returns raw stored content. On this head, a configured cap of 100 returned 918 characters; the shipped `quality` profile enables this path. The same boundary must cover the linear/polyphonic MEMORIA supplements and fact merge. Please add `recall()` and `recall_enhanced()` regressions with the relevant flags enabled.

tests/test_beam_content_cap.py
```diff
@@ -0,0 +1,114 @@
...
+    return tmp_path / "content_cap.db"
+
+
+class TestCapDefaults:
+    def test_default_is_500(self):
+        # NOT the fixture-reset attribute (that would be vacuous — the
+        # fixture writes 500 before this runs): recompute what a clean
+        # import WOULD resolve and require the shipped default to match.
+        # Fails under any mutation of the production default.
+        assert beam_module.RECALL_CONTENT_CAP == beam_module._env_int(
```
dplush: **[P2] This does not verify the production default.** The autouse fixture has already assigned `beam_module.RECALL_CONTENT_CAP = 500`, then this assertion compares that fixture value with `_env_int(..., 500)`. It will still pass if the initializer in `beam.py` changes its default. Test a fresh import/subprocess (or otherwise inspect the unmodified production initializer) so the 500-character compatibility contract is genuinely pinned.

dplush (changes requested):
Current-head review: the linear paths and cache identity are improved, but the configured public recall bound is still bypassed by supported feature-gated result paths. Please enforce the cap at a shared public-result boundary and cover every affected recall mode before re-review.

linouxis9:
Both findings addressed on the new head; both were reproduced locally first, then fixed, then re-pinned with regressions.

**[P1] Shared public-result boundary.** The cap is now enforced at every `recall()` exit rather than only at the six linear producers:

- New module-level `_cap_recall_results()` / `_cap_recall_content()` (beam.py, next to `RECALL_CONTENT_CAP`). `recall()` routes ALL four of its return paths through it: polyphonic plain return, polyphonic explain payload, linear explain payload, and the linear plain return. That covers the polyphonic engine mapper, both MEMORIA structured-fact supplements (linear + polyphonic) and the fact-voice merge (`MNEMOSYNE_FACT_RECALL_ENABLED=1`) by construction — any future feature-gated producer that flows through `recall()` inherits the bound automatically.
- Defense in depth: `_polyphonic_row_to_dict()` (the shared engine-row mapper behind `_fetch_polyphonic_row`) now caps content at the mapping site too, so the polyphonic result shape is bounded even if a future caller bypasses `recall()` and calls `_recall_polyphonic()` directly.
- The stale comment claiming the polyphonic/fact-voice paths "do not apply this cap — pre-existing, tracked separately" is gone (it described the bug); replaced by the boundary description.

Reproduction note: I could not reproduce dplush's 918-char number through the engine itself in a no-embeddings environment (the vector voice has no candidates there, so `_recall_polyphonic` returned 0 rows), so I verified the leak at the mapper/fetch layer instead — raw stored content flowed through `_polyphonic_row_to_dict`/`_fetch_polyphonic_row` untouched (678 chars with cap=100), which is the same defect surfacing at the layer below the engine. The new tests pin the mapper and the fetch directly, so the regression class is covered even where the engine can't run.

**[P2] Non-vacuous default pin.** `test_default_is_500` no longer touches in-process state at all (the old form was exactly the tautology described: the autouse fixture had already assigned 500, then compared it with `_env_int(..., 500)`). It now spawns a fresh interpreter that imports production `beam.py` with a scrubbed env and asserts the resolved constant is 500. A companion meta-test (`test_shipped_default_mutation_is_caught`) copies the repo to a scratch tree, mutates the shipped initializer to 999, and requires the same probe to observe 999 there — so the probe is proven non-vacuous in both directions. (I reproduced the original tautology first: with the shipped default mutated to 999, the old assertion still passed.)

**Tests** (`tests/test_beam_content_cap.py`, now 21): the four boundary regressions (mapper, fetch, fact-merge path, explain payload, all at cap=100 against >100-char content), the subprocess default pin + its mutation meta-test, the previous default/env-fallback/opt-in/cache-identity coverage unchanged.

Full suite: 3520 passed / 0 failed (`MNEMOSYNE_NO_EMBEDDINGS=1`, minus the five environment-dependent files that fail identically on `main`). `ruff check --select F,RUF022` clean on both changed files. Cache identity is unchanged from the previous head (`recall_content_cap` still in the key payload, version 8).


**Round-4 addendum (CodeRabbit, review of `0ca8306`).** Both comments fixed:

- The four env-value tests now assert through the production import: a `_fresh_import_cap()` subprocess helper imports `beam.py` with each env value (`4000` / `garbage` / empty / `-400`) and asserts the resolved constant — so hardcoding the default or renaming the key fails them. (They complement the existing subprocess default-500 probe + mutation meta-test.)
- The persisted-cache test is no longer vacuous: the seeded row lexically matches the query, the warm call asserts a nonempty result before the cache check, and the cap-B call asserts nonempty before the cap-length assertions.

21 tests in the file, all green; ruff clean on changed files.

linouxis9:
Hi @coderabbitai — both fixed at HEAD (`ca6d61a`).

**1. Env tests now go through the production import.** Added a
`_fresh_import_cap()` subprocess helper: each of `4000` / `garbage` / `""` /
`-400` is now asserted via `beam.RECALL_CONTENT_CAP` after a clean
interpreter import of production `beam.py` with the env var set — so a
hardcoded default or a renamed env key fails these tests (they complement
the existing fresh-import default-500 probe and its mutation meta-test).

**2. The persisted-cache test is no longer vacuous.** The seeded row now
lexically matches the query (`"cap identity probe notes: …"`), the warm
call asserts a nonempty result before the cache check, and the cap-B call
asserts a **nonempty** result (not just `is not None`) before the length
assertion — so an empty result set can no longer silently satisfy the test.

Both new tests green: 21/21 in `tests/test_beam_content_cap.py`, full suite
3520 passed / 0 failed (known-environmental files excluded), ruff clean.

AxDSan (changes requested):
The narrowed shape is right and dplush's two holds are addressed. Three small things before I merge, none of them about the design:

1. `test_shipped_default_mutation_is_caught` does `shutil.copytree(_REPO_ROOT, ...)` with only `.git` and caches ignored. On a developer checkout that copies `.venv` too, which is 360 MB here, on every test run. Copy only `mnemosyne/` into the scratch dir (that is all the probe imports), or drop the meta-test and keep `test_default_is_500`.

2. Two pieces of #911 leaked in: the `_ENHANCED_RECALL_CACHE_VERSION = 8  # agreed with the vec-admission PR; resolver keeps 8` comment, and `MNEMOSYNE_EM_VEC_ADMIT` in the conftest scrub list. Neither name exists on main. Bump to 7 with no comment and scrub only `MNEMOSYNE_RECALL_CONTENT_CAP`. Whichever PR lands second rebases and bumps again; that is fine.

3. No CHANGELOG entry. Add one under Unreleased, Added: the env var, that the default stays 500, and that the cache key now includes the cap. Please also retitle the PR so the squash title does not say `500 -> 4000`; the default did not change.

linouxis9:
Hi @AxDSan!

Thanks a lot for your review on all my PRs! I'll handle the requested changes on all of them, thanks a lot to the both of you for your work on mnemosyne!

Valentin

linouxis9:
Addressed the remaining maintainer checklist in `1fb9a1d`:

- The mutation test now copies only `mnemosyne/` and verifies its scratch import path. A simulated developer `.venv` is excluded; reverting to whole-checkout copying fails the regression.
- Cache version is **7** against base version 6. Removed the cross-PR coordination comment and unrelated `MNEMOSYNE_EM_VEC_ADMIT` scrub.
- Added the Unreleased / Added entry and corrected the PR title/body. **Default remains 500**; larger values are explicit opt-ins.

The additional review found one remaining cap bypass in enhanced associative results. That is now capped after augmentation, before persistence/return, and on cache-hit return. Real-graph tests cover cold and cached output at 1/500/4000 characters; removing either guard independently fails the three corresponding cases.

@coderabbitai resume

dplush (approved):
Re-reviewed on `8e5895d`. The remaining content-cap boundary and cache invalidation concerns are addressed, the focused test matrix passes, and CI is green. I have no further changes requested. Approving.

dplush:
Thanks, the narrowed contract and cache isolation look good. One merge blocker remains: `BeamMemory.fact_recall()` still returns uncapped `fact_text` directly, so its public output bypasses `MNEMOSYNE_RECALL_CONTENT_CAP` (including reachable SHMR callers). Please cap those results and add direct `fact_recall()` coverage with over-cap content.

The CodeRabbit note about inline `[:RECALL_CONTENT_CAP]` slices is not blocking: it is behaviorally equivalent for schema-backed text and the public boundary caps again.

linouxis9:
Addressed in `d692abe`. `BeamMemory.fact_recall()` now applies the shared content cap after sorting and `top_k`, covering both raw and consolidated fact stores. Direct over-cap regressions cover both sources, and an SHMR regression verifies automatic fact retrieval does not receive the uncapped tail. Reverting the boundary fails all three cases.

Local full suites and the cumulative cap/cache mutation battery are green; GitHub CI is running.

dplush (approved):
Final review on d692abe: Sol and Claude found no merge-blocking defect. CI and CLA are green; the direct fact_recall/SHMR cap regression is covered.

dplush:
Final review is complete on `d692abe`.

The direct `fact_recall()` and SHMR cap coverage resolves the remaining blocker. CI and CLA are green, and both independent review gates found no merge-blocking issue.

I approved the PR. The non-blocking hardening ideas are tracked separately in #922.

linouxis9:
Hey @dplush! What are the next steps to get this merged?

Thanks and cheers,
Valentin

dplush:
Thanks for checking in, Valentin.

There are no further technical changes requested from my side. The final review is clear, and the additional hardening ideas are tracked separately in #922.

We are waiting for @AxDSan maintainer merge decision, since he requested the final cleanup and indicated he would take the merge decision once that was addressed.

The only remaining mechanical point is that the branch is now behind current `main`, so that will need to be reconciled as part of the merge step. I’ll ask @AxDSan to confirm the next step.

linouxis9:
Rebased onto current `main` (`a4f334ff84db538536cdc3e20fc353ba33c61690`) and force-pushed as `b7fc6c393e3aa9ae6943c99ff30143b127cb72f3`.

Rebase review:
- retained current main's single `_env_int` and conflict-budget semantics;
- advanced the enhanced-recall cache generation from main's 8 to 9;
- kept current vector-admission cache material and extended the upgrade regression across generations 6/7/8;
- preserved the final ordinary/enhanced/polyphonic/fact/SHMR cap boundaries.

Hostile validation on a frozen export:
- full core: **4281 passed, 6 skipped, 7 subtests passed**;
- full Hermes integration: **473 passed, 1 skipped**;
- Python 3.10 focused matrix: **181 passed**;
- focused cap/cache/fact/SHMR/conflict matrix: **223 passed**;
- 8 boundary/cache mutants killed, including the final direct fact/SHMR boundary and independent cold/warm enhanced-cache caps;
- candidate wheels built and passed an installed-wheel runtime probe.

No unresolved MEDIUM+ defect found. The branch is one commit on current main and clean. CI is the remaining external gate.

**CI update:** all seven Actions jobs (`build`, `docs-check`, `lint`, Python 3.10–3.13), CLA, and CodeRabbit completed successfully on this exact head. GitHub reports the PR mergeable and clean.

dplush:
@linouxis9, thanks. I found no functional blocker in the current implementation, but GitHub now reports this PR as conflicting with current `main`.

Please rebase onto current `main`, preserving both the new unreadable-marker cache regression from `main` and this PR’s cache-version regression. Then run the focused cache/content-cap tests and wait for fresh CI on the rebased head.

Please also classify the remaining current review threads. The direct cap slices are a non-blocking maintainability point, and the cache test could be strengthened separately, but neither requires a production change in this PR.

@AxDSan, once the rebased head has fresh green CI and the threads are classified, could you confirm the final merge decision? This touches the recall pipeline, so the final gate remains with you.

dplush (approved):
Reviewed the rebased head `b8031ab`.

The shared public-result boundary finding is resolved. The remaining persisted-cache request is valid test hardening, but not a demonstrated production defect: `recall_content_cap` is part of cache identity and the focused regression/probe verifies that a changed cap recomputes rather than reuses the prior entry. Further content-cap hardening is tracked in #922.

Approving this head. The final merge decision remains with AJ because this touches the recall pipeline.

Issue #773, opened 2026-08-17
Non-English (Cyrillic) memories never injected into context - Latin-only tokenizer in prefetch


olegmaisak:
## Short version

Mnemosyne-oss does **not work out-of-the-box for non-English languages** in auto-context injection.
Two layers are responsible: (1) the default embedding model `bge-small-en-v1.5` is English-only;
(2) **critical bug** in the Hermes plugin: `_PREFETCH_TOKEN_RE` tokenizes Latin only (`[a-z0-9]...`),
so any Cyrillic / non-Latin memory has 0 tokens and is silently dropped in `_semantic_dedup_prefetch`
-> the `## Mnemosyne Context` block is always empty for non-English content. `recall()` finds the
rows, but `prefetch()` (what actually injects) never does.

---

### Environment
- `mnemosyne-memory` 3.15.1, `mnemosyne-hermes` 0.5.0 (wrapper mode, side-venv)
- Hermes memory provider: `memory.provider: mnemosyne`
- User language: Russian (interface and memory entirely in Russian)
- Default embedding: `BAAI/bge-small-en-v1.5` (384-dim)

### Symptom
A Russian-speaking user. Memories are written (`beam.remember(..., scope='global')` -> row in DB),
`recall('Директриса Яндекс Директ')` **finds** the row (score 0.57, importance 0.9, topic_signal 0.92).
But `MnemosyneMemoryProvider.prefetch(<any Russian query>)` **always returns empty string** -- the
`## Mnemosyne Context` block in the system prompt is empty, injection never happens. The agent
"doesn't remember" Russian memory even though it is in the DB.

### Root cause (empirical: code + experiment)

**Bug 1 - Latin-only tokenizer in prefetch (Hermes plugin, critical):**

File `mnemosyne_hermes/__init__.py`, ~line 226:

    _PREFETCH_TOKEN_RE = re.compile(r"[a-z0-9][a-z0-9_./:-]*", re.IGNORECASE)

This regex matches Latin only. Cyrillic words ("олег", "директриса") never match.

Consequence - `_prefetch_tokens(content)` returns an **empty set** for Russian text. Then in
`_semantic_dedup_prefetch` (~line 402):

    tokens = _prefetch_tokens(row.get("content", ""))
    if not tokens:
        continue   # <- Russian row silently dropped

And `_prefetch_topic_signal` returns 0 (no tokens), which additionally rejects the row in the
[...]

PR #915, opened 2026-09-05
fix: force UTF-8 stdout/stderr in run_cli to prevent UnicodeEncodeError on Windows pipes

Ralph-Abejuela:
## Description

When `mnemosyne` is spawned by agent tooling with piped stdout on Windows (the common case: the [pi extension](https://github.com/mnemosyne-oss/pi-mnemosyne) and similar integrations `spawn()` the CLI and capture output), Python defaults `sys.stdout` to the cp1252 codec. Printing memory content that contains any character outside cp1252 then kills the whole command:

```
File "mnemosyne/cli.py", line 203, in cmd_recall
    print(f"  Content: {content[:150]}{'...' if len(content) > 150 else ''}")
UnicodeEncodeError: 'charmap' codec can't encode character '\u20b1' in position 116: character maps to <undefined>
```

This made `mnemosyne recall` fail 100% of the time for me the moment a stored memory contained a peso sign (₱); emoji, arrows, and CJK content in any recalled memory would do the same. Running the same command in an interactive shell worked because the agent sets `PYTHONUTF8=1` in its own process environment but not in the spawned child's, so the bug only surfaced through tool integrations and was very confusing to diagnose.

`run_cli()` now reconfigures `sys.stdout` and `sys.stderr` to UTF-8 with `errors='replace'` at startup. Output is correct when the stream supports UTF-8, and degrades to replacement characters instead of a traceback when it genuinely cannot. No behavior change on POSIX, where UTF-8 is already the default.

## Related Issue

No existing issue covers this (searched open/closed issues and PRs for encoding/Unicode/charmap; #217 and #773 touch other layers of non-ASCII handling). Reported from a real deployment.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] New regression test: `tests/test_cli_utf8_stdout.py` replaces `sys.stdout` with a cp1252 `TextIOWrapper` over a `BytesIO` and drives `run_cli()` end-to-end with a faked memory containing ₱, —, and →. Fails on current `main` with the exact `UnicodeEncodeError` above; passes with this change. Verified both directions locally (checked out `HEAD~1` for the without-fix run).
- [x] Manual verification on Windows 11 / Python 3.13: reproduced the crash by spawning the CLI from Node with `PYTHONUTF8`/`PYTHONIOENCODING` unset (`exit 1`, cp1252 traceback), and confirmed `exit 0` with correct ₱ output after the fix.
- [x] `pytest tests/test_cli_utf8_stdout.py` passes (full suite not run locally; no existing tests touch this path).

## Checklist

- [x] No version bump: deferred, following the existing precedent in the changelog ("the version bump is deferred to that PR")
- [x] `CHANGELOG.md` updated with a brief entry under Unreleased → Fixed
- [x] No README change needed (no user-facing flag or behavior contract changes)
- [x] No new dependencies, no cloud requirement, SQLite untouched

## Summary

- `run_cli()` reconfigures stdout and stderr to UTF-8 with `errors="replace"`.
- Piped CLI output on Windows no longer fails on characters outside cp1252.
- Regression tests cover Unicode recall output, stderr encoding, and unknown-command errors.
- The change does not affect memory tiers, retrieval, consolidation, veracity, sync, privacy, local-first guarantees, Hermes, MCP, or benchmark methodology.
- The change improves CLI reliability and maintainability without eroding the local-first design.
- The changelog documents the fix.


mnemosyne/cli.py
```diff
@@ -1771,6 +1771,13 @@ def cmd_migrate(args):
 
 def run_cli():
     """Main CLI entry point."""
+    # Force UTF-8 output: on Windows, stdout/stderr default to cp1252 when piped
+    # (e.g. by agent tooling spawning this CLI), and printing memory content that
+    # contains non-cp1252 characters (e.g. '\u20b1') raises UnicodeEncodeError and
+    # kills the command. See docs/integrations/pi.md tooling which pipes output.
+    for stream in (sys.stdout, sys.stderr):
+        if stream is not None and hasattr(stream, "reconfigure"):
+            stream.reconfigure(encoding="utf-8", errors="replace")
```
dplush, line 1780: **[P1] Treat stream reconfiguration as best effort.** `hasattr(stream, "reconfigure")` only proves the attribute exists; it does not guarantee it succeeds. A minimal host/proxy stream whose `reconfigure()` raises `OSError` makes even `mnemosyne --help` raise before the CLI error boundary. This is a regression from the prior behavior and is particularly relevant to embedded/agent runners that replace stdio. Catch `OSError`/`ValueError`/`io.UnsupportedOperation` (and continue with the existing stream), then add a regression for a stream exposing but rejecting `reconfigure()`.

dplush (changes requested):
The cp1252-pipe regression is addressed, but the new global stream setup must not make every command fail before normal CLI handling when a host-provided stream cannot be reconfigured.

dplush (approved):
Follow-up correction: the previous stream-proxy concern was not corroborated by the independent current-head review. It relied on a synthetic proxy, while direct cp1252-backed stdout and stderr checks preserve Unicode and the PR-scoped tests pass. Please treat that inline comment as withdrawn. The remaining gap is non-blocking coverage: assert Unicode preservation and exercise the stderr path.

Ralph-Abejuela:
Thanks for the review @dplush, and understood on the withdrawn P1: the stream-proxy scenario needed a synthetic `reconfigure()`-rejecting object to construct, so no real-world crash path; noted for any future embedded-runner reports.

Pushed to close the remaining coverage gap:

- stdout test now asserts Unicode preservation, not just ASCII: `₱` and `→` must survive verbatim, so a mangle-to-`?` regression fails the test instead of passing.
- Two new stderr tests: (1) `run_cli()` reconfigures a cp1252-backed `sys.stderr` to UTF-8; (2) an unknown-command path with a non-ASCII command name prints through stderr with replacement characters and exits `2` cleanly instead of raising.
- Docstrings added to the test module and helpers (also clears the CodeRabbit docstring-coverage warning).

A/B re-verified locally: all three tests fail on `main` with the original `UnicodeEncodeError: 'charmap' codec can't encode character '\u20b1'` and pass with the fix.

dplush (approved):
Re-reviewed on the current head after the two test-only follow-ups. Independent Builder review found no current-head findings. The controlled cp1252-pipe reproduction now preserves non-cp1252 recall output as UTF-8, and the focused tests, complete CI matrix, and diff check pass. The current CodeRabbit note is a non-blocking test-strengthening suggestion, not a production defect.

PR #917, opened 2026-09-05
fix(consolidation): never auto-invalidate heuristic conflict pairs

linouxis9:
## Summary

Heuristic similarity must not expire memories. With conflict-LLM validation disabled, detected pairs remain untouched and are reported as `conflicts_detected_only`.

The motivating deployment audit observed 142 of 243 items invalidated across consolidation passes. That is a specific motivating observation, not a general failure rate. This PR does not rewrite or repair historical records.

## Safety contract

- Count every detected pair exactly once in `conflicts_resolved` or `conflicts_detected_only`. Only a committed invalidation with validation provenance resolves a pair. False/malformed verdicts, exceptions, missing/unsafe endpoints, refused invalidations and budget skips remain detected-only.
- Allow at most one successful supersession per older ID per `sleep()` invocation, including across source groups. Failed attempts do not consume that ID. Supersession and provenance commit atomically; insert/commit failures roll back both.
- Defer query-cache I/O until after the conflict transaction commits. Cache failure cannot roll back the durable resolution. `invalidate(..., defer_cache_invalidation=True)` explicitly delegates cache clearing to the caller; its default remains false, preserving other callers' commit ownership and cache-failure behavior.
- Require HTTPS whenever an API key is configured, before sleep dispatch and at the transport boundary. Reject embedded URL authority credentials. Neither urllib nor httpx follows redirects, including same-origin redirects. Conflict-validation failure logs do not include raw prompts or credential-bearing exception text.
- Require a JSON boolean decision, finite numeric confidence in [0,1], and a string-or-null correction. A string such as `"false"` must not become a positive verdict through Python truthiness.
- `MNEMOSYNE_CONFLICT_PAIR_BUDGET` defaults to 20 and `MNEMOSYNE_CONFLICT_TIME_BUDGET_S` to 300. Positive finite overrides apply across all source groups in each `sleep()` call; empty/invalid/nonpositive/nonfinite values fall back. The cooperative deadline clips transport timeouts and stops additional calls/retry backoff; it does not preempt every blocking network operation.
- Both sleep APIs report stable counters on no-op paths and aggregate results consistently. Dry runs make no backend-availability, conflict-validation, summarization or model-refresh inference calls and do not write cost/memory records. Fallback warnings cannot initialize/download a model during dry runs. Budget-skip diagnostics aggregate once per sleep pass; failure diagnostics expose the exception class without its message.

## Current revision

Based on main `d7e116f` after #920 merged and that exact main commit passed push CI. Main has not advanced since the preceding #917 refresh, so no redundant rebase was performed. This update changes only #917.

The dry-run warning no longer calls backend availability or reaches model initialization/download. Tests exercise real availability logic for local, remote and host configurations across both sleep APIs, with positive non-dry-run failure controls. Nine witnesses fail behaviorally before this correction.

The valid review cleanup is included: exception-class-only warnings, one aggregate budget warning, a JSON1-independent provenance failure trigger, explicit SQLite reader closure and removal of a redundant assertion. The `log_cost` patch target remains unchanged: it is correctly imported into the caller module.

New contracts exercise real SQLite consolidation through public MCP and both Hermes-provider dispatchers, preserving both counters for single/all-session and dry/real runs. Independent mutations drop the counter in each consumer. The earlier atomicity/cache-deferral fixes remain unchanged. This pass adds 30 test cases.

## Scope

BEAM consolidation, conflict transport/response validation, tests and Unreleased/Changed documentation. No provider production changes, vector-admission changes, new network dependency, schema migration, historical rewrite or production deployment. No enhanced-recall cache-version change. The inherited regex-test measurement correction retains its actual production pattern and both two-second deadlines.

## Verification at `fb07cb5`

One logical commit; exact source/tree/diff unchanged through validation.

- Full core suite: **4030 passed, 6 skipped, 7 subtests passed in 469.77s (0:07:49)**.
- Full Hermes-provider suite: **465 passed, 1 skipped in 82.12s (0:01:22)**.
- Real Python 3.10 focused checks: **306 passed in 47.45s**.
- Ruff `F,RUF022` and `git diff --check`: passed. No test-file exclusions.
- All 30 behavioral mutants caught: 13 safety, three provenance-atomicity, four cache-boundary, nine review-regression and one reader-closure mutation, with passing controls and no collection/setup errors.

GitHub CI, CLA and maintainer approval are separate from these local validation results. No merge, deployment or new signature was performed.

## Summary

- Prevents heuristic similarity detection from automatically invalidating memories.
- Counts every detected pair as either `conflicts_resolved` or `conflicts_detected_only`.
- Commits supersession and validation provenance atomically.
- Limits supersession to one successful replacement per older memory per `sleep()` call.
- Adds configurable pair and time budgets for conflict validation.
- Requires HTTPS for credentialed endpoints and blocks redirects.
- Validates LLM response content and schema.
- Makes dry runs side-effect-free, including no conflict validation, model work, or record writes.
- Preserves stable conflict counters for no-op and completed consolidation.
- Adds regression coverage for accounting, atomicity, transport safety, budgets, cache timing, and integration surfaces.
- Documents the behavior change and the 142/243 audit observation.

### Consolidation and veracity

The BEAM consolidation pipeline now treats heuristic similarity as candidate evidence only. LLM validation must succeed before the system invalidates a memory.

Failed, malformed, refused, unavailable, and budget-skipped validations remain detected-only. Weak evidence no longer changes durable memory state.

Supersession and `memory_validations` provenance share one transaction. Failed inserts and commit failures roll back supersession. Query-cache invalidation occurs only after commit.

This is the right call for memory veracity and auditability.

### Local-first and privacy posture

The LLM remains optional. When validation is disabled or unavailable, Mnemosyne keeps stored memories unchanged and reports detected candidates.

Credentialed endpoints require HTTPS. Both supported transports block redirects, which prevents credential forwarding. Response validation, bounded retries, and deadlines limit unsafe external calls.

The change does not alter providers, schemas, vector admission, historical records, synchronization, or deployment. No local-first design erosion is identified.

### Agent integration surfaces

The core `sleep()` and `sleep_all_sessions()` results now expose stable conflict counters. MCP and Hermes paths preserve these counters through dispatch and serialization.

The PR does not change Hermes providers, MCP commands, or CLI commands. It preserves session-scoped consolidation behavior.

`BeamMemory.invalidate()` adds an optional cache-invalidation deferral argument for atomic callers.

### Unchanged architecture

The PR does not alter working, episodic, or BEAM tier definitions. It does not change retrieval strategies, vector admission, sync behavior, provider behavior, or benchmark methodology.

Only the BEAM consolidation and veracity path changes. Other memory tiers and integration layers remain unchanged.

### Maintainability

Environment-overridable budgets make conflict processing predictable and bounded. Explicit counters make partial validation and no-op runs easier to audit.

Focused tests cover accounting, transaction atomicity, transport safety, budgets, dry-run immutability, cache timing, and Hermes/MCP surfaces. The provenance record provides a durable explanation for each successful invalidation.


dplush:
The core change is important and directionally correct: a heuristic similarity signal must not silently expire memories. I support keeping this as a focused data-integrity fix, but the LLM-resolved path needs a few safeguards before merge.

**P1 — allow only one successful supersession per older memory in a sleep pass.** An `older_id` can occur in several detected pairs. After the first successful `invalidate()`, later pairs can overwrite `superseded_by` and add conflicting validation provenance. Please track successfully superseded older IDs, skip subsequent pairs for them, and count those skipped pairs as detected-only. Add a three-row regression with multiple candidate replacements.

**P2 — make the configured LLM transport safe.** When an API key is configured, require an HTTPS endpoint. The urllib fallback must also prevent cross-origin redirects from forwarding the Authorization header. Please add fallback-path coverage.

**P2 — preserve conflict accounting.** A heuristic pair that receives `is_conflict=False`, a malformed response, or a failed LLM call currently contributes to neither counter. For an auditable result, every detected pair should be represented exactly once as either resolved or detected-only.

Please include the stable no-op `sleep_all_sessions()` counter shape in the same pass, rerun the focused conflict matrix, and request a fresh review.

AxDSan:
Agree with the direction and with dplush's three points, so this waits for that pass. Two more for the same pass:

1. This is a default behavior change: heuristic-only sleep no longer invalidates anything, and `conflicts_detected_only` is a new key on `sleep()` and `sleep_all_sessions()`. It needs a CHANGELOG entry under Unreleased, Changed, that says so in one paragraph, including the 142/243 production number since that is the justification.

2. Lift `_CONFLICT_PAIR_BUDGET` and `_CONFLICT_TIME_BUDGET_S` out of the loop body to module-level constants next to the other `MNEMOSYNE_*` knobs, env-overridable through the existing `_env_int` / `_env_float` helpers. Hardcoded 20 and 300 inside `sleep()` are the kind of thing the next operator report is about.

dplush (changes requested):
Final review on `cb409b5`: LLM conflict resolution is not atomic with its validation provenance. `sleep()` first calls `invalidate()`, which commits the irreversible supersession, and only then writes `memory_validations`. If that insert fails, the exception is swallowed and the result still reports a resolved conflict.

I reproduced this by rejecting only the provenance insert with a SQLite trigger: the old memory was invalidated and superseded, `conflicts_resolved` was `1`, but `memory_validations` remained empty.

Please make the invalidation and provenance write transactional, rolling both back on a provenance failure, or treat that case as detected-only. Please also add a regression test that forces the provenance insert failure and asserts both database state and counters.

linouxis9:
@dplush The provenance-atomicity correction is now on `c9168c4`, rebased onto merged #918. Supersession and `memory_validations` insertion share the transaction; rejected/ignored inserts and deferred commit failure roll back both. Success tracking and resolved counters advance only after successful persistence, so a failed pair can still be retried with a later candidate. Both sleep APIs, summary/claim cleanup and a second-connection visibility probe are covered. Full core/provider suites and the focused Python 3.10 checks pass; all thirteen prior safety mutants and three atomicity mutants are caught. This refresh preserves both sleep functions unchanged from the prior corrected head. Ready for re-review.

dplush:
@linouxis9 Hi Valentin, could you please rebase? #917 would be my next merge candidate. Thank you!

linouxis9:
Hi @dplush! Sure, feel free to tell me the order you want to merge them so I rebase them in the correct order.

Thank you so much!
Valentin

dplush:
@linouxis9

#917 first 
then wait for green main-CI
#920 afterwards

regards
Denis

AxDSan:
Order, so nobody rebases twice: #920 just merged. Then this one (#917), then #913, then #919, then #911 last since it carries the largest beam.py diff. Rebase each onto main only after the one before it lands, and for #913 and #911 take whatever `_ENHANCED_RECALL_CACHE_VERSION` main has at that moment and add one. dplush and I will merge each as soon as it is green on the fresh base.

linouxis9:
Hi @AxDSan! Thank you, I'll proceed with the rebases in this order, waiting for the main CI turning green each time.

Valentin

dplush:
Thanks, reviewed against `cf88ee2`.

We will keep two test-level items in this PR because they protect the new contract:
- make the provenance trigger independent of SQLite JSON1, and
- add explicit MCP/Hermes surface coverage for `conflicts_detected_only`.

The logging, budget-warning aggregation, reader-close, and tautological-assertion suggestions are valid but non-blocking cleanup; we will handle them separately to keep this atomicity fix focused.

The `log_cost` patch-target suggestion is not applicable: `mnemosyne.core.llm_conflict_detector` imports `log_cost` directly and the existing target is valid.

linouxis9:
Thanks! Your review references `cf88ee2`; we pushed `fb07cb5` shortly before your comment. It includes both requested test additions, plus the cleanup.

One logging-related change fixes a behavioral issue: during a dry run, the warning’s availability check can initialize the model and trigger a download. The new guard prevents that.

Would you prefer we remove the non-blocking cleanup from this PR while retaining that dry-run fix and the two requested test additions?

dplush (approved):
Approved after final review of `fb07cb5`: the atomic provenance rollback path, SQLite-portable failure tests, and MCP/Hermes response-surface coverage are verified.

dplush (approved):
Reviewed current head fb07cb5. The atomic provenance fix and public counter coverage look good; focused CI-parity matrix passes.

linouxis9:
Thanks a lot @dplush! Proceeding with the rebase of #913 once Main CI is green.

Cheers,
Valentin

Issue #872, opened 2026-08-26
Design Hermes v2 pre-compress checkpoint support for durable transcript evidence


dplush:
## Why

Hermes now offers an opt-in v2 pre-compress checkpoint contract. With `compression.checkpoint_required: true`, compaction fails closed unless an active provider advertising v2 completes a durable checkpoint. The current Mnemosyne provider does not advertise that API or implement `on_pre_compress()`.

This is a real provider capability gap, but not an incident: the Hermes gate is off by default and existing Mnemosyne deployments keep their current behavior.

## Goal

Decide whether and how Mnemosyne should support Hermes v2 pre-compress checkpoints for durable direct transcript evidence before a lossy Hermes compaction.

## Proposed contract to decide

- Opt in only with the v2 API marker and an explicit `on_pre_compress()` implementation.
- Treat the host-provided v2 rows as direct evidence only. Do not use tool results, system messages, tool-call payloads, or prior summaries as checkpoint input.
- Make writes durable and idempotent across retries and overlapping compaction attempts.
- On any checkpoint failure, raise rather than report partial success; the host then preserves the uncompressed transcript.
- Keep checkpoint evidence separate from ordinary semantic memory unless an explicit retention and retrieval contract says otherwise.

## Acceptance

- A focused Hermes integration test proves the provider advertises v2 and checkpoints before a checkpoint-required compaction may proceed.
- Retrying the same checkpoint produces no duplicate durable evidence.
- Failure leaves no partial-success claim and blocks the lossy compaction through the Hermes contract.
- Tests prove the stored checkpoint boundary excludes tool/system/summary-derived content.
- Existing deployments remain unchanged unless an operator explicitly enables the Hermes checkpoint gate.

## Non-goals

- Enabling `compression.checkpoint_required` by default.
- Treating checkpoints as automatic long-term semantic-memory extraction.
- Changing the HERMES_HOME/update-survival work in #859.
- Implementing before the retention, storage, and ownership decisions above are agreed.

## References

- https://hermes-agent.nousresearch.com/docs/developer-guide/memory-provider-plugin#pre-compress-checkpoints-fail-closed

PR #918, opened 2026-09-05
fix(recall): opt-in self-echo suppression with compression-boundary release

linouxis9:
## Summary

Replace the original session-wide time-window filter with **opt-in, provider-owned self-echo suppression during automatic prefetch**. The default is ordinary recall. There is no six-hour window, timestamp comparison, recent-turn ring, or claim of exact live-context membership.

Enable with `MNEMOSYNE_SELF_ECHO_ENABLED=1` before constructing the provider. Both Hermes provider packages implement the same shared core contract; Hermes itself is unchanged.

## Compression-boundary release

- Even with the option enabled, nothing is suppressed until the provider observes an `on_pre_compress` callback for the session. Hosts without that callback continue ordinary recall.
- Every observed callback releases the entire previous capture batch. Released rows become eligible for ordinary ranking, not forcibly injected. A no-op, failed compression or preserved tail can therefore allow some redundant recall intentionally.
- Subsequent captures qualify only with fresh owned-row provenance and post-boundary ordering evidence from the optional `sync_turn(messages=...)` transcript: a unique unchanged boundary-tail anchor must precede the source.
- Generation checks and released-source evidence prevent queued or in-flight pre-boundary writes from re-arming old exclusions. Missing/ambiguous evidence, unsupported projections or resource overflow favor ordinary recall.
- Explicit `mnemosyne_recall` never inherits the prefetch exclusions.

## Ownership and retrieval safety

A shared `exclude_captures` snapshot replaces the former session-key kwargs. Suppression is tied to exact newly captured working-memory rows, their actual stored session, capture metadata and unchanged stored content. Text equality or a deduplicated insert result is not proof of ownership. NULL-session, imported, legacy, edited and otherwise unproven rows remain eligible, including separate rows with identical text. No schema migration or historical rewrite is required.

Linear and polyphonic retrieval filter working-memory candidates before consuming the relevant candidate limits/fusing contributions. Eligible episodic representations retain their own content, tier and scores; a bare-ID match must not remove an episodic twin. Duplicate-heavy excluded pools are regression-tested with a surviving candidate below the original pool limit.

## Compatibility limits

This is best-effort suppression, not an eviction acknowledgement or durable checkpoint implementation:

- Callback observation does not prove that every future host context transition will invoke it. Suppression lifetime follows observed callbacks and provider state; unobservable host compactions are not covered.
- Missing optional sync transcripts or image-bearing host projections that cannot be matched safely cause abstention, not guessed ownership.
- Provider snapshots are revoked at release, and an in-flight prefetch retries ordinary recall if revoked. The plugin cannot rewrite a prefetch string already held by Hermes during a tool loop; the next user turn re-fetches released rows.
- #872 remains separate. This PR neither advertises v2 checkpoint capability nor bypasses a host requirement for a durable v2 checkpoint.

## Runtime and privacy boundaries

The provider-local fallback is bundled in both distributions so missing optional ledger support on an older core preserves ordinary memory writes. SQLite-limit handling supports Python 3.10, and standalone recall closes only the connection it owns. Missing-transcript tests assert the actual stored memory, not only empty exclusions.

The ledger itself uses local in-memory state and SQLite and makes no network calls. It does not disable separately configured remote embedding, canonical/shared-memory or sync paths; those retain their existing privacy implications. The enablement guide now states that boundary explicitly.

## Producing-tier boundary

The enablement guide explicitly states that matching working-memory candidates are filtered before ranking and fusion. Eligible consolidated episodic representations remain recallable when their working source is excluded from automatic prompt context; normal episodic eligibility rules still apply. This follow-up changes documentation only.

## Current revision

Rebased on current main `4f1e53a`, preserving the merged Korean-query changes.

MEMORIA source-row hydration now includes the complete `wm_where` predicate and binds `(*_source_ids, *wm_params)`. A proof-excluded raw capture cannot return as `memoria_source_*`. The same query also respects ordinary working-row session, validity, supersession and explicit recall filters. Independently eligible structured facts and unowned same-text rows remain available.

## Verification at `2c0293e`

Single logical commit on `4f1e53a`; source/tree/diff stayed unchanged during final validation.

- Full core: **3852 passed, 6 skipped, 7 subtests passed**; no test-file exclusions.
- Full provider integration: **465 passed, 1 skipped**; no test-file exclusions.
- Ruff `F,RUF022` and `git diff --check`: passed.
- Real Python 3.10 focused checks: **121 passed, 3 skipped**. Skips cover environment-only host dependencies and the runtime SQLite-limit API unavailable before Python 3.11.
- `tests/test_memoria_source_exclusions.py`: **16 passed** using real storage, fact extraction and public recall. Fourteen cases fail on the previous source, and three independent predicate/binding mutations are detected. Positive controls protect eligible hydration, proof release and structured-fact availability.

GitHub CI and maintainer approval remain separate current-head gates. No Hermes host change, deployment, historical rewrite or package release.

## Summary

- Replaces timestamp and session-key self-echo filtering with an opt-in, provider-owned `VerbatimLedger`.
- Enables suppression only when `MNEMOSYNE_SELF_ECHO_ENABLED=1` and an `on_pre_compress` callback has been observed.
- Uses stored row ownership, capture metadata, session identity, and transcript ordering as proof.
- Fails open when evidence is missing or ambiguous.
- Keeps explicit recall unrestricted.

## Architecture

- Filters working-memory candidates before ranking, vector retrieval, and result fusion.
- Preserves eligible episodic representations when a working-memory twin is excluded.
- Applies exclusions consistently to linear BEAM recall, polyphonic recall, and MEMORIA source hydration.
- Adds generation invalidation, callback-boundary release, bounded SQLite resolution, and post-compression re-arming.
- Leaves the veracity system and Hermes core unchanged.
- Does not alter the sync layer beyond optional `messages` capture metadata in `sync_turn`.

## Privacy and local-first guarantees

- Keeps suppression local to the provider and database.
- Makes no network calls and adds no host or database migrations.
- Avoids broad time-based hiding of conversation history.
- Does not retroactively suppress imported, legacy, edited, NULL-session, or otherwise unproven rows.
- Provides a conservative fail-open path when ownership, provenance, transcript evidence, or SQLite capacity is insufficient.
- No change is reported for MCP or CLI integration surfaces.

## Agent integration

- Implements the contract in both Hermes provider packages.
- Adds `on_pre_compress` and `on_session_switch` provider hooks.
- Keeps the feature disabled by default.
- Preserves ordinary prefetch behavior when the optional ledger is unavailable.

- Centralizes proof and exclusion behavior in `mnemosyne.core.verbatim_ledger`.
- Adds compatibility fallbacks for standalone provider packages.
- Documents configuration, limitations, restart behavior, and best-effort scope.
- Adds coverage for both providers, both recall engines, queued and in-flight writes, duplicate candidates, missing hooks, resets, source hydration, and compatibility boundaries.
- Reported validation passed: 3,852 core tests, 465 provider integration tests, focused Python 3.10 checks, Ruff, and diff checks.

This is the right architectural direction. It removes heuristic global filtering and uses explicit, bounded, provider-owned evidence while preserving local-first behavior.


dplush:
The self-echo goal and the recent-working-memory boundary are sound. I think this should remain one focused PR, but the linear and polyphonic filters need to agree on timestamp and session semantics before merge.

**P1 — separate naive and aware timestamp predicates.** In the linear working-memory clause, `timestamp <= naive_cutoff` is evaluated for every row, including `Z`- and offset-bearing timestamps. Those rows can therefore be kept or excluded by a lexicographic wall-string comparison before the UTC-normalized arm is relevant. Restrict the naive cutoff comparison to truly naive timestamps, and evaluate aware values only through the normalized aware-time path. Please add offset/Z regressions on both sides of the window under a non-UTC timezone.

**P2 — do not include NULL-session rows in the polyphonic exclusion fetch.** The `OR session_id IS NULL` condition can remove fresh import/legacy rows that do not belong to either excluded session key. The linear path keeps these rows, so please make the polyphonic path match that contract and add a NULL-session regression for both engines.

**P2 — pass the actual beam session key as the alternate provider variant.** The root Hermes provider currently constructs `hermes_<session_id>` instead of consistently passing `self._beam.session_id` when it differs from the raw caller ID. Deployments with another session-key derivation can otherwise still self-echo. Please align both provider copies and add a mismatched-key regression.

Please also restore `TZ` before `tzset()` in the test cleanup, rerun the focused linear/polyphonic/provider matrix, and request a fresh review.

dplush:
One additional blocker from a full current-head check: the polyphonic filter can also remove an episodic candidate despite the PR’s stated contract that episodic summaries remain eligible. Fusion and vector voice deduplication use the bare `memory_id`; filtering the fused result solely by that ID means a matching fresh working-memory row can remove the corresponding episodic representation as well. Preserve tier/provenance through fusion, or apply the exclusion before fusion only to working-memory candidates.

Given the provider-key mismatch, naive/aware timestamp split, NULL-session regression, and this tier-provenance issue, I recommend redesigning this as a narrow shared exclusion contract plus provider wiring rather than extending the current implementation. Please add parity coverage for both providers and both recall engines before resubmitting.

AxDSan:
Holding with dplush's review; I hit the same NULL-session over-exclusion reading the polyphonic path. One design point on top, because it decides the default:

A six-hour window excludes the session's own rows exactly when they stop being in the context window. Hermes compacts long sessions, and after compaction the memory layer is the only way back to what was said three hours ago in the same conversation. With this default, a user asking "what did I decide earlier about X" in a long session gets nothing for six hours, and that is worse than an echo. Please either ship it off by default (`MNEMOSYNE_SELF_ECHO_HOURS` unset means no exclusion) or scope the exclusion to what is actually still on screen, for example the last N turns the provider itself just synced rather than a wall-clock window. The provider knows what it wrote this session; the core does not.

Also needs a CHANGELOG entry once the shape settles.

linouxis9:
@dplush @AxDSan I've replaced the time-window/session-key design rather than extending its predicates. The title, body and CHANGELOG now describe the new contract.

- **Off by default:** `MNEMOSYNE_SELF_ECHO_ENABLED=1` opts in. There is no six-hour cutoff or recent-turn ring. Even enabled, a provider that has not observed `on_pre_compress` performs ordinary recall.
- **Compression releases the old batch:** every observed pre-compress callback makes all previously tracked captures eligible again, even if compression retains a tail or does nothing. This deliberately accepts an occasional echo rather than using a pre-hook as proof of eviction. Subsequent captures require transcript ordering evidence; generation and released-source checks also cover delayed queued/in-flight writes.
- **Shared ownership contract:** both providers use actual stored capture provenance, session and row identity, not guessed `hermes_` key variants or blanket content equality. NULL-session/import/legacy/edited rows and unowned same-text rows stay eligible. The old naive/aware timestamp predicates and their timezone-dependent tests are removed because row timestamps no longer determine exclusion.
- **Tier safety:** working contributions are filtered before candidate selection/fusion; eligible episodic twins keep their own content, tier and scores. Both engines and both provider call paths have parity coverage, including duplicate-heavy pools with an eligible survivor. Explicit recall remains unrestricted.

Hermes itself is unchanged. Missing callbacks or ambiguous transcript projections fail open. This does not implement #872 or advertise v2 checkpoint support; a required unsupported v2 checkpoint still fails. It also cannot rewrite a prefetch string already held by the host mid-tool-loop, and does not claim coverage for compaction paths that bypass the hook. Actual next-user-turn re-fetch was verified.

Validation on the published candidate's tree: **3,665 passed, 6 skipped, 7 subtests passed** across the full local core suite, no test-file exclusions; changed-file ruff and diff checks passed. Isolated tests against the unchanged host exercised no-op and durable compression, missing hooks, queued/in-flight sync and post-compression re-arming for both providers. An independent offline audit found no MEDIUM+ defect; its remaining notes are LOW coverage/compatibility limitations, not a claim of exhaustive proof. The PR body records those boundaries and mutation results explicitly.

Ready for a fresh review of this replacement design. GitHub CI on the new head is separate from the local results above.

dplush (approved):
Re-reviewed on `0afa45a`. The opt-in self-echo contract, ownership boundaries, compression release, and recall fallback behavior are covered. The focused matrix and full CI are green. The remaining session-key test point is already covered by `tests/test_verbatim_compat.py`. I have no further changes requested. Approving.

dplush (changes requested):
Re-review on `becc4f1`: the self-echo exclusion can be bypassed through the MEMORIA source-row path.

The normal working-memory queries apply `wm_where`, including the proof-based `exclude_captures` clause. The MEMORIA source query at `beam.py:7813-7817` only selects `WHERE id IN (...)`, so an excluded provider-owned capture can be reintroduced as `tier="memoria_source"` with its full content.

Please apply `wm_where` to this source-row query and pass the corresponding parameters. Please also add a regression test proving that a proof-excluded capture cannot return as `memoria_source_*`.

linouxis9:
@dplush Fixed in `2c0293e`. The MEMORIA source query now applies `wm_where` and binds the source IDs followed by `wm_params`, as requested.

The new `test_proof_excluded_capture_cannot_return_as_memoria_source` uses real `remember()` fact extraction and public recall, first proves the capture exclusion resolves, then checks that neither the working row nor its `memoria_source_*` alias returns. It also verifies that independently eligible structured facts remain available and revocation restores source hydration.

The surrounding regressions cover expiry, supersession, session and explicit recall filters, equal-text unowned sources, parameter ordering and the legacy SQLite variable budget. All 16 new cases pass; 14 fail against the previous source. Reverting the predicate and breaking the bindings independently are detected.

Final local validation: **3852 core tests passed, 6 skipped, 7 subtests passed; 465 provider tests passed, 1 skipped**. Real Python 3.10 focused checks: **121 passed, 3 environment/runtime skips**. Ruff is clean. The PR remains one logical commit; the body and changelog describe the corrected boundary. GitHub CI is a separate current-head check.

Rebased on `4f1e53a` after #896 merged; the new Korean-query tests pass too.

dplush (approved):
Re-reviewed on `2c0293e`. The MEMORIA source-row exclusion now applies the complete working-memory filter and parameter set. The targeted regression suite and adversarial SQLite-limit probe pass, and full CI is green. I have no further changes requested. Approving.

PR #920, opened 2026-09-07
fix(plugins): isolate discovery module keys and restore failed loads

linouxis9:
## Summary

Extract the plugin-loader changes previously mixed into #911 and #919 into a focused PR, as requested in their reviews.

- Derive each discovery module key from the resolved plugin path using lossless filesystem-byte hex encoding in a private namespace. `logging.py` cannot replace stdlib `logging`; `a-b.py` and `a_b.py`, and same-stem files in separate directories, cannot overwrite each other.
- Preserve successful modules in `sys.modules` so discovered classes remain importable under their defining module.
- On execution, class inspection or registration failure, restore the exact previous module state (including an existing `None` entry), but only while the failed load still owns the key. Do not clobber a replacement or resurrect a deliberately removed entry.
- Inspect classes before registering them, then roll back any owned partial registrations on failure. This also covers module `__dir__`/`__getattr__` failures after successful execution.
- Clean up before propagating `KeyboardInterrupt`/`SystemExit`; ordinary plugin exceptions retain the existing warn-and-continue behavior.

## Scope

Plugin discovery, regression tests, existing plugin-test cleanup and an Unreleased CHANGELOG entry, plus a shared regex-test measurement correction needed for reliable full-suite validation (described below). No vector scoring, query sanitization, provider changes, schema migration, new dependencies or Hermes changes.

This is module/registry bookkeeping, not a sandbox or rollback of arbitrary plugin side effects. Discovery is not being turned into a thread-safe concurrent loader or hot-reload API. Paths resolving to the same canonical file intentionally share module identity.

The duplicate loader delta is removed from #911 and #919. Both feature PRs link here and remain independently testable without this unmerged implementation. #919 addresses its sanitizer/provider review requirements in its own revision; that behavior is outside this PR.

## Successful rediscovery

Successful canonical-path discovery is reusable, not hot reload. Cache the inspected classes on their owning module and reuse those exact classes on later discovery, including discovery by another manager or a canonical alias. Do not execute an edited file again while retaining an old registry class. Module, registry, instance type and pickle reconstruction stay coherent. Previously failed loads remain retryable; pre-existing uncached modules retain the existing ownership-aware rollback rules.

Focused tests cover same/second managers, canonical aliases, editing after success, instance and pickle identity, all existing failure paths, and exact prior-state restoration. Reverting only successful-module reuse causes three behavioral failures.

## Current revision

Base-only refresh onto merged #918. The approved plugin loader and module-identity regression file are byte-identical to the prior head. The plugin fixture cleanup already merged into main correctly disappears from this PR delta. No hot-reload, loader or rollback redesign.

## Verification at `7467e6e`

One logical commit on main `b4b6539`, validated with unchanged source/tree/diff.

- Full core suite: **3882 passed, 6 skipped, 7 subtests passed in 466.24s (0:07:46)**.
- Full provider suite: **465 passed, 1 skipped in 85.69s (0:01:25)**.
- Real Python 3.10 focused checks: **104 passed in 3.46s**.
- Ruff `F,RUF022` and `git diff --check`: passed. No test-file exclusions. Environmental skips remain explicit in the receipts.

GitHub CI, CLA status and maintainer approval are separate exact-head gates. No deployment, historical rewrite or release was performed.

## External checks

CLA Assistant has posted a successful `license/cla` status on this exact head. No maintainer recheck request, new signature or substitute status was necessary. GitHub Actions and maintainer review remain separate gates.


dplush (approved):
Re-reviewed on `fd57ce1`. The path-based module identity, module-cache rollback, cross-plugin isolation, and manager reuse behavior are covered. The focused tests and full CI matrix are green. I have no further changes requested. Approving.

PR #923, opened 2026-09-08
fix(#914): per-write author identity for Hermes provider memory writes

### LINKED ISSUE #914: Hermes provider: writes never carry author_id (multi-agent identity layer incomplete)
**Environment:** mnemosyne-memory 3.15.1, Hermes plugin (`hermes_memory_provider`)

**Design intent (docs.mnemosyne.site/use-cases/multi-agent):** "Memories are auto-tagged with who stored them" — identity resolved from instance args or `MNEMOSYNE_AUTHOR_ID`/`_TYPE`/`CHANNEL_ID` env.

**Gap:** the SDK (`Mnemosyne(session_id, author_id=...)`) and MCP server (`mcp_tools._create_instance`, env-fallback) both set identity on the instance, so writes auto-tag. The Hermes plugin path constructs `BeamMemory(session_id=self._session_id)` with no identity at all, and neither `mnemosyne_remember` nor `mnemosyne_batch` forwards `author_id` (drop-through confirmed in handler). Result: every write lands with `author_id = NULL` (indices `idx_wm_author`/`idx_em_author` exist but stay unused).

**Note:** setting `beam.author_id` from env would ALSO scope the automatic prefetch (`_prefetch_bank` reads `self._beam.author_id or env`), which the provider comment warns bypasses thread/session scoping — so the ideal fix is a **write-side stamp** (author_id applied to the INSERT) that leaves `beam.author_id` unset for reads.

**Expectation from core comment** (`beam.py`, BeamMemory.__init__): "Integrations that know the active profile should set this attribute after construction." The Hermes integration is that missing leg.

Proposed: honor `MNEMOSYNE_AUTHOR_ID`/`_TYPE` in the plugin write path only (remember + batch), mirroring mcp_tools.py resolution.

roadtogideon222-byte:
## Summary

Fixes #914 (writes never carry `author_id` in the Hermes provider path).

Root cause (confirmed against main): both Hermes provider surfaces construct their `Mnemosyne`/`BeamMemory` instances without author identity, so every write stores NULL `author_id`. The documented `MNEMOSYNE_AUTHOR_ID` env var was only consulted in the recall path (`_prefetch_bank`), never for writes — unlike the MCP path, which reads all three identity env vars into the constructor correctly.

## Approach: narrow per-write identity (per the design note on #914)

Instead of constructor injection, identity is attached **at the write**:

- **Core:** `BeamMemory.remember()` accepts `author_id`/`author_type` params. `None` falls back to the instance identity (exact legacy behavior). Both the dedup-update and new-row INSERT paths use the per-write identity. Read identity (`self.author_id`, used by recall author-scoping) is never mutated.
- **Both providers, strict parity:** `hermes_memory_provider` and the standalone `integrations/hermes` surface resolve identity once (`agent_identity` kwarg > `MNEMOSYNE_AUTHOR_ID`/`_TYPE` env > none) and pass it at every `remember()` call site.
- **Constructors deliberately unchanged:** recall author-scoping keys off `Beam.author_id`; setting it would make prefetch pass a non-empty `author_id`, which triggers the `(1=1)` clause that bypasses session/channel scoping. This regression is caught by `test_prefetch_scopes_to_thread` and is why the constructor path was reverted in favor of per-write.

I evaluated constructor injection first and hit exactly that failure against the thread-isolation suite before reverting — so the per-write direction isn't just preferred, it's the one that survives the existing tests.

## Tests

New: `tests/test_remember_per_write_identity.py` (core per-write semantics: storage, instance fallback, override, dedup-update, batch per-item) and `tests/test_hermes_provider_author_identity.py` (resolution precedence, read identity untouched, end-to-end storage with a real `BeamMemory`, NULL when unresolvable, env-var fallback).

All 129 tests across the Hermes provider, shared-surface, thread-isolation, and multi-agent-identity suites pass. (Full-suite runs in my environment show ~130 preexisting failures from cross-test state pollution, identical on clean HEAD — the sync-correctness/security failures reproduce without this branch and pass in isolation.)

## Notes for review

The maintainer note on #914 also flagged provenance for pending-approval writes and sleep consolidation. Pending writes (`mnemosyne_apply_pending`) now carry identity via the same per-write path. Consolidation provenance (`self.author_id` in `consolidate_to_episodic`) is **not** touched here — it needs the row-provenance discussion from the issue thread and seemed right to keep in a focused follow-up rather than bundle into this fix.

Found while evaluating Mnemosyne as shared memory for a multi-agent Hermes deployment — the identity layer is exactly what that use case needs.

## Summary

- Adds per-write `author_id` and `author_type` support to `BeamMemory.remember()`.
- Applies identity to both insert and deduplication update paths without changing read identity.
- Resolves identity in Hermes from `agent_identity`, then environment variables, then no identity.
- Covers regular, batch, identity-signal, and pending-approval writes.
- Preserves thread and session scoping because `BeamMemory.author_id` remains unset for recall and prefetch.
- Keeps constructors unchanged and does not modify consolidation provenance.

- Working-memory attribution now supports multi-agent writes. Episodic memory, BEAM retrieval, tiering, consolidation, veracity handling, and sync behavior remain unchanged.
- Hermes and standalone `integrations/hermes` now provide write-side identity parity with the MCP integration.
- CLI behavior is unchanged.
- The implementation preserves local-first operation. Identity resolution uses local arguments and environment variables. No remote service or new sync dependency is introduced.
- Per-write identity is the right architectural choice. It separates write attribution from read scoping and avoids privacy and retrieval regressions caused by setting the Beam read identity.
- Tests cover precedence, batching, storage, deduplication, unresolved identity, and read-identity preservation.

### DIFF
--- a/hermes_memory_provider/__init__.py
+++ b/hermes_memory_provider/__init__.py
@@ -1898,6 +1898,46 @@ class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
 
         return "default"
 
+    def _resolve_author_identity(self) -> dict:
+        """Resolve author identity for memory writes (issue #914).
+
+        Precedence (matches the MCP path in mnemosyne.mcp_tools._create_instance):
+        1. agent_identity kwarg (explicit, from Hermes)
+        2. MNEMOSYNE_AUTHOR_ID / MNEMOSYNE_AUTHOR_TYPE env vars
+        3. No identity -> empty dict (constructors called without author
+           kwargs, preserving the legacy NULL-author behavior exactly).
+
+        'primary' is a generic identity, not a specific author, and is
+        excluded the same way _resolve_profile_bank treats it.
+        """
+        identity = getattr(self, "_agent_identity", None) or ""
+        if identity and identity.lower() not in ("primary", "default", "none", ""):
+            result = {"author_id": identity}
+            author_type = os.environ.get("MNEMOSYNE_AUTHOR_TYPE")
+            if author_type:
+                result["author_type"] = author_type
+            return result
+
+        env_author = os.environ.get("MNEMOSYNE_AUTHOR_ID")
+        if env_author:
+            result = {"author_id": env_author}
+            author_type = os.environ.get("MNEMOSYNE_AUTHOR_TYPE")
+            if author_type:
+                result["author_type"] = author_type
+            return result
+
+        return {}
+
+    def _write_identity_kwargs(self) -> dict:
+        """Per-write author identity kwargs for remember() calls (issue #914).
+
+        Narrow per-write identity path per the #914 design note: identity is
+        attached at the WRITE, never by mutating the Beam's read identity
+        (self.author_id), because recall author-scoping keys off that field
+        and setting it would bypass session/channel scoping in prefetch.
+        """
+        return self._resolve_author_identity()
+
     def initialize(self, session_id: str, **kwargs) -> None:
         """Initialize Mnemosyne beam for this session."""
         # C27: clear stale state from any prior init attempt so a re-init
@@ -2530,6 +2570,7 @@ class MnemosyneMemoryProvider(HermesPersonaPromptMixin, MemoryProvider):
                     importance=0.85,
                     scope="global",
                     veracity="stated",
+                    **self._write_identity_kwargs(),
                 )
                 break  # One identity memory per turn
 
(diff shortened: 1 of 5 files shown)


AxDSan:
This is the shape I want for #914: identity attached at the write, beam read identity untouched, Hermes `agent_identity` preferred over the env var. Two blockers, both mechanical:

1. Lint: two unused imports, `os` and `pytest`, in the new test files. `ruff check --select F` on the changed files shows them.
2. Sign the CLA; the bot is still pending.

Fix those and I merge. #926 covers the same stamp plus more; I have asked its author to rebase on top of this once it lands and keep only what this PR does not do.

PR #928, opened 2026-09-08
fix(mcp): no second http.response.start after SSE stream ends

### LINKED ISSUE #910: Bug Report: SSE transport — silent session death (no keepalive) + double `http.response.start` on stream end
## Summary

Two bugs in the SSE transport (`mnemosyne/mcp_server.py`, `serve_sse` path)
degrade long-lived MCP clients:

1. **No SSE keepalive** — idle SSE streams die silently (TCP/NAT/proxy
   timeouts), and MCP clients never notice the session is gone.
2. **Double `http.response.start` after the stream ends** — `handle_sse`
   returns a `JSONResponse` after `connect_sse()` has already started the
   response, which starlette/uvicorn reject with
   `RuntimeError: Expected ASGI message 'http.response.body', but got 'http.response.start'`
   whenever the server shuts down (or the stream ends) with open sessions.

## Issue 1: no keepalive → silent session death

`SseServerTransport` (sse-starlette 3.x) defaults to `ping=None`, so an idle
SSE stream emits **no frames at all**. Any stateful middlebox (NAT table,
ingress, HTTP proxy, systemd socket timeout) can then drop the connection
without either side noticing.

Observed behavior with long-lived agent clients (uptime days):

- The SSE stream is dropped silently.
- The client keeps POSTing to `/messages/?session_id=...`; the server answers
  `404` (session no longer exists), but the client never re-runs the SSE
  handshake, because nothing told it the stream died.
- Tool calls fail with `404` / `-32602 Invalid request parameters` and memory
  writes are silently lost until the client process is restarted.

### Reproduction

1. Start the SSE server and connect an MCP client.
2. Leave the stream idle through a NAT/proxy idle-timeout (or kill the
   connection at the network layer without closing sockets cleanly).
3. Issue any tool call → client posts to a dead `session_id` and never
   reconnects.

### Proposed fix

Send periodic SSE comment frames (`: ping\n\n`) on the stream, interval
configurable via env (e.g. `MNEMOSYNE_SSE_KEEPALIVE_S`, default ~20 s,
`0` = off). Comment frames are ignored by MCP clients per the SSE spec, but
keep middleboxes alive and let clients detect a dead stream quickly.

Important detail: keepalive frames and `EventSourceResponse` frames must be
serialized through a shared `asyncio.Lock` around the ASGI `send` callable —
they interleave on the same stream and must never be sent concurrently.

## Issue 2: double `http.response.start` on stream end

```python
async def handle_sse(request):
    async with transport.connect_sse(request.scope, request.receive, request._send) as streams:
        await server.run(...)
    return JSONResponse({})   # ← second response on an already-started response
```

`connect_sse()` sends `http.response.start` itself and streams until
disconnect. Returning a `JSONResponse` afterwards emits a second
`http.response.start`, which starlette/uvicorn reject:

```
RuntimeError: Expected ASGI message 'http.response.body', but got 'http.response.start'
```

This fires on every service restart that has open SSE sessions (observed
repeatedly in production logs during systemd restarts).

### Proposed fix

Return a no-op response after the stream ends, e.g.:

```python
class _SseStreamDoneResponse(Response):
    async def __call__(self, scope, receive, send):
        return None  # response was already started by connect_sse()
```

## Environment

- mnemosyne-memory: 3.15.1
- Python: 3.14
- mcp SDK: 2.0.0
- OS: Ubuntu (Linux), uvicorn + starlette, systemd-managed service
- Transport: SSE

## Backward compatibility

- Keepalive frames are SSE comments — ignored by all spec-compliant clients;
  can be disabled via env (`0`).
- The no-op response only removes an exception path; no behavioral change for
  clients.

We run both fixes in production as a local patch (happy to upstream it as a PR
if there is interest).

dwmb5055:
Closes #910 (the confirmed half).

## Summary

- `handle_sse` returned `JSONResponse({})` after `SseServerTransport.connect_sse()` had already started the ASGI response, emitting a second `http.response.start` on every disconnect / server shutdown with open sessions.
- The handler now returns `_SseStreamDoneResponse`, a `Response` subclass whose `__call__` sends nothing — the transport owns the whole response lifecycle.
- New regression test drives the real transport with a hand-written scope through a controlled `http.disconnect` and asserts exactly one `http.response.start`, status 200, `text/event-stream`.

## Why

With open SSE sessions, every service restart logged:

```
RuntimeError: Expected ASGI message 'http.response.body', but got 'http.response.start'
```

Observed in production on systemd-managed restarts. Related to #383 but not a duplicate: the #383 fix introduced the post-stream `JSONResponse`.

## Scope

Per maintainer guidance on #910: no keepalive addition (sse-starlette already pings by default when `ping=None`), no env knob, no send lock. The change is limited to the response-lifecycle defect plus its regression test.

## Tests

- `tests/test_mcp_sse_response_lifecycle.py`: exactly one `http.response.start` across a controlled disconnect (pre-fix: two).
- `tests/test_s1_mcp_sse_auth.py` and adjacent MCP suites pass unchanged.

- No changes to memory tiers, retrieval, consolidation, veracity, sync, or local-first guarantees.
- No changes to MCP tool contracts, auth (S1), or the streamable-HTTP transport.
- The only behavioral change is the removal of an exception path on SSE stream end.

## Summary

- Prevents duplicate `http.response.start` messages in the SSE lifecycle.
- Returns `_SseStreamDoneResponse` after `SseServerTransport.connect_sse()` owns the response.
- Adds a regression test for disconnect handling and `text/event-stream` responses.
- Does not add SSE keepalive frames.

## Architecture and scope

- No changes to working, episodic, or BEAM memory.
- No changes to retrieval, consolidation, veracity, synchronization, or benchmarks.
- No changes to privacy posture or local-first guarantees.
- Preserves the MCP integration surface and improves SSE shutdown reliability.
- Hermes and CLI behavior remain unchanged.

## Maintainability

The no-op response makes response ownership explicit and removes a known ASGI lifecycle error. The regression test protects this behavior with the real SSE transport. The focused change is the right call for the duplicate-response defect.

The separate SSE keepalive issue remains unresolved.


dplush (approved):
Reviewed current head `34be33e`.

- The fix correctly preserves response ownership in `SseServerTransport.connect_sse()` and prevents the route handler from emitting a second ASGI response.
- The regression test drives the real SSE transport through request, stream start, disconnect, and completion, and asserts exactly one `http.response.start`.
- Focused test: 1 passed.
- GitHub CI and CLA checks are green.
- No blocking correctness issue found.

Approving the current code. The branch is now behind `main` after #954; please rebase onto current `main` and let CI rerun before merging.

dplush (approved):
Independent re-review of current head `68a1824cbd001490d1c91a7e55a75ee310b38ff4`.

- Focused regression: `1 passed`.
- Related MCP SSE/Auth tests: `111 passed`.
- Ruff: `All checks passed!`.
- Negative control on the pre-fix base reproduced two `http.response.start` messages and failed as expected.

No blocking correctness, security, or scope issue found. Current-head CI and CLA are green.

dwmb5055:
Thanks for the triage — the PR is already scoped exactly this way: the only change is `_SseStreamDoneResponse`, a Response-compatible no-op returned after `connect_sse()`'s stream ends (no ping loop, no env knob, no lock — see the Scope section). The regression test drives the real `SseServerTransport` through a controlled disconnect and asserts exactly one `http.response.start`.

PR #957, opened 2026-09-14
feat(core): recall provenance logging

### LINKED ISSUE #956: Feature: recall provenance logging (persistent query→result audit trail)
## Problem

`recall_diagnostics.py` (C4) exposes per-window counters for recall-path signals, and polyphonic recall attaches `voice_scores` to results — but neither answers the question operators ask after the fact: **"which memory IDs did query X return on turn T?"**

Use case: an agent answers something wrong. The injected context came from recall. To debug *why*, you need the persistent mapping turn → query → returned memory ids (with scores). Today that mapping evaporates when the recall call returns.

## Proposal

Opt-in, persistent, stdlib-only provenance log:

- `MNEMOSYNE_RECALL_PROVENANCE=1` enables it (default OFF — zero-config, no surprise disk writes)
- When set, every linear-path `BeamMemory.recall()` appends one JSONL line to `<data dir>/recall_provenance.jsonl`: `{ts, query[:200], top_k, results: [{id, tier, score, importance, timestamp}]}`
- New `mnemosyne/core/recall_provenance.py` with `append_recall_provenance()` / `read_recall_provenance()` — never raises (logging must not break recall), thread-safe append, no content previews (ids suffice, records stay lean)
- Pairs with `recall_diagnostics.py`: counters tell you *how often* each signal fired; provenance tells you *exactly which rows* a query surfaced

Agent-agnostic: lives in core, works for any host. No schema changes, no API changes, no new dependencies.

## Implementation

Will open a PR referencing this issue (branch `feat/recall-provenance`): new module + one guarded hook at the single final return of linear `recall()` + tests (`tests/test_recall_provenance.py`) + version bump + changelog.

Questions for maintainers:
1. Is JSONL-next-to-DB the right shape, or would you prefer a SQLite table / ring buffer?
2. Should enhanced/polyphonic recall paths log too, or is linear-only acceptable for a first pass?

AxelFooley:
Closes #956 (filed first per CONTRIBUTING).

## What

When `MNEMOSYNE_RECALL_PROVENANCE=1` (default off, read per call), linear `recall()` appends one JSONL line to `recall_provenance.jsonl` next to the DB — query (200-char cap), `top_k`, and returned ids with tier/score/importance/timestamp. New `mnemosyne/core/recall_provenance.py` exposes `append_recall_provenance()` / `read_recall_provenance()` (newest-first, limit, malformed lines skipped).

## Principles check

- **Minimal dependencies:** stdlib only (`json`, `os`, `threading`, `datetime`, `pathlib`).
- **Zero-config:** opt-in flag, default OFF — no surprise disk writes.
- **Fast/never harmful:** appends are lock-serialized and never raise; a provenance failure is logged at debug level and cannot affect recall results.
- **Lean records:** ids and scores only, no content duplication.

## Scope

Linear path only — `recall_enhanced()` and polyphonic delegation return before the hook (documented in the module docstring). Happy to extend in a follow-up if maintainers want it.

## Tests

`tests/test_recall_provenance.py` — flag on writes a valid line containing both remembered ids; flag off (default) writes nothing; garbage value treated as off; reader ordering + limit; flag read per call. `MNEMOSYNE_NO_EMBEDDINGS=1`: 5 passed; `tests/test_beam.py`: 123 passed, 1 pre-existing failure (`TestSleepCycle::test_sleep_falls_back_to_aaak_for_token_truncated_reasoning` — fails identically on clean HEAD, environment/LLM-dependent, unrelated to this change).

## Housekeeping

`__version__` 4.0.0b2 → 4.0.0b3, CHANGELOG entry added.

By submitting this PR, I confirm I have read and agree to the CLA.

## Summary

Adds opt-in provenance logging for linear `BeamMemory.recall()` calls.

When `MNEMOSYNE_RECALL_PROVENANCE=1`, each call writes a bounded JSONL record with the query, `top_k`, and returned memory IDs with tier, score, importance, and timestamp.

## Architecture

- Adds `mnemosyne/core/recall_provenance.py`.
- Excludes enhanced, polyphonic, and `explain=True` recall paths.
- Does not change tiered memory, consolidation, veracity, sync, schemas, recall results, or benchmark methodology.
- Uses standard-library code only.
- Preserves local-first design because records stay next to each database.

## Privacy and integration

- Logging is disabled by default.
- Records contain up to 200 characters of query text and no content previews.
- Files use `0600` permissions.
- Rotation, bounded reads, and orphan cleanup limit retention.
- Hermes, MCP, and CLI APIs require no changes.
- Operators must treat the JSONL files as sensitive plaintext audit data.

## Reliability and maintainability

- Append failures are fail-open and debug-logged.
- Per-database locks prevent concurrent record interleaving.
- The linear-only scope is appropriate for the initial implementation because it avoids changing other retrieval semantics.
- Tests cover failures, concurrency, isolation, permissions, rotation, cleanup, bounded reads, empty results, and excluded paths.
- Reported verification includes 19 provenance tests, touched-file lint, documentation verification, and no new full-suite failures relative to `main`.


dplush (changes requested):
Thanks for the work here. I cannot approve the current head yet. The CI is green, but the following blockers need a design and scope correction before merge:

1. Release scope: please revert the `4.0.0b3` bump and all propagated version-only edits. Keep the feature entry under `CHANGELOG.md` Unreleased. Release/version surfaces are handled separately.

2. Legacy Hermes scope: please remove the `hermes_memory_provider/plugin.yaml` update. We are converging on the standalone `mnemosyne_hermes` path rather than maintaining the legacy provider as a parallel supported surface.

3. Store isolation: `_provenance_path()` maps every database in one directory to the same `recall_provenance.jsonl`. For example `a.db` and `b.db` share and read one undiscriminated audit trail. Use a database-specific store/path and add a same-directory two-database isolation regression.

4. Privacy and lifecycle: this persists up to 200 raw query characters in plaintext with normal `open(..., "a")`, no retention/deletion relationship to the database, and an unbounded full-file read. Please define and implement restrictive creation permissions, bounded retention or rotation, database-bound cleanup semantics, and a bounded tail reader. Add tests for those guarantees.

5. Recall failure and concurrency contract: the synchronous global lock plus file write can delay every enabled in-process recall, while the docs say it must never slow recall. It also provides no inter-process write contract. Either isolate this latency from recall or narrow the claim honestly, scope locking per store, define multi-process behavior, and add real-path write-failure and concurrency coverage.

6. Public contract/docs: decide whether `explain=True` linear recalls are intentionally excluded, then align implementation, tests, and prose. Register `MNEMOSYNE_RECALL_PROVENANCE` in the canonical env-only docs with the plaintext and lifecycle warning and regenerate the docs.

The earlier limit-boundary and enhanced-delegation findings are addressed on this head. Please bring the remaining storage/privacy contract back as a focused revision, then request a fresh review.

AxelFooley:
Rebased on current `main` (merge commit d7db08d) — only CHANGELOG.md conflicted, CI is green across all four Python versions. All six blockers from the Sep 14 review were addressed in 856e1f1 / 3652998, plus the orphaned-rotation cleanup in 5653dce. @dplush could you re-review when you get a chance? @coderabbitai review

dplush (approved):
Reviewed current head a0def51bd14ad0eab7ed359c0e72b302292defa7.

No blocking correctness, privacy, or lifecycle issues found in the current diff. The provenance path is per-database, opt-in, permission-restricted, bounded/rotated, fail-open, and the enhanced/polyphonic/explain behavior is covered and documented. Focused provenance tests pass (19), docs verification passes, and the repository CI checks are green.

PR #987, opened 2026-09-19
fix(beam): score working-memory int8 candidates from their stored bytes

### LINKED ISSUE #982: [BUG] Working-memory int8 vector arm guesses a similarity scale: dense_score saturates at 0.92-0.95
## Description

`_wm_vec_search_sqlite()` (working-memory vector arm) is the only vector arm
that still converts a sqlite-vec distance into a similarity with an ad-hoc
scale guess:

```python
sim = max(0.0, min(1.0, 1.0 - (max(distance, 0.0) / (2.0 * EMBEDDING_DIM))))
```

`_vec_distance_sim()` documents why that cannot work for int8 ("the
quantization scale is dimension- and distribution-dependent ... this arm
abstains rather than guessing a scale"), and the episodic paths already score
from the stored bytes with `_vec_int8_blob_cosine()` (#911). The
working-memory arm was not migrated.

Because the guess compresses every vector-matched candidate into a narrow high
band, the working-memory dense blend adds a near-constant term:

```python
if vec_sim > 0:
    base_score = base_score * (1.0 - _vec_blend) + vec_sim * _vec_blend
```

`MNEMOSYNE_WM_VEC_BLEND` (default 0.20) can therefore penalise weak candidates
(near-zero similarities) but cannot lift a strong one.

## Environment

- **OS:** Ubuntu-based Linux container
- **Python version:** 3.13
- **Mnemosyne version/commit:** core `3.15.1` (stable) in one install; the same
  code path is present on `main` (`mnemosyne/core/beam.py`, the
  `_wm_vec_search_sqlite()` distance mapping)
- **Installation method:** pip install into a virtualenv
- **Hermes version:** provider package in use, but the probe below is core-only

## Reproduction Steps

Read-only probe against a store copy (no writes), int8 vec table:

1. Sweep the blend on the same store copy:
   `MNEMOSYNE_WM_VEC_BLEND` in `0.0 / 0.2 / 0.5 / 0.8` and run the same query
   set with `recall()`.
2. Compare `dense_score` in the returned rows with the exact cosine of the same
   query/row pairs (computed from the stored vectors with numpy).

Minimal illustration of the mapping, no store needed:

```python
# Row-to-row cosine spread on an int8[768] store: 0.502 - 0.688 (95th pct)
# The legacy mapping turns that into: 1 - 127*sqrt(2-2*cos)/(2*768) -> 0.926 - 0.9459
import math
for cos in (0.502, 0.688):
    print(round(1 - 127 * math.sqrt(2 - 2 * cos) / (2 * 768), 4))
```

## Expected Behavior

`sim` for the working-memory arm should be comparable to the episodic arm's
absolute cosine, so `MNEMOSYNE_WM_VEC_BLEND` has amplitude to re-rank with.
Candidates that are far apart should receive visibly different `dense_score`
values.

## Actual Behavior

Every vector-matched candidate reports `dense_score` **0.92-0.95** regardless
of true similarity, while the exact cosines for the same pairs span
**0.60-0.79**:

```
# blend sweep on the store copy (linear path, polyphonic disabled)
blend 0.0  -> target in top-5: 7/40   dense_max 0.9452
blend 0.2  -> target in top-5: 5/40   dense_max 0.9452
blend 0.5  -> target in top-5: 5/40   dense_max 0.9452
blend 0.8  -> target in top-5: 5/40   dense_max 0.9452
```

Sweeping the weight moves orderings (30-38 of 40 queries) but never lifts a
strong candidate, because the added term is nearly constant.

## Mnemosyne Status

```
working memory:     600+ rows
episodic memory:    200+ rows
embedding rows:     1000+ (two models: a 768-dim multilingual embedder and a
                     legacy 384-dim English-centric store)
vector table:       int8[768]
```

(`hermes mnemosyne stats` equivalent; counts rounded.)

## Optional: Database Info

```
sqlite-vec version: 0.1.x
vec_working DDL:    CREATE VIRTUAL TABLE vec_working USING vec0(embedding int8[768])
```

## Relationship to existing issues

Not a duplicate, but it is adjacent to three open issues and it changes what
they should do first:

* **#937** (open, P2) — working-memory recall under-blends dense similarity and
  the linear blend is hardcoded at 0.20. The 0.95 dense similarity reported
  there sits exactly inside the saturated band this issue describes: the blend
  weight cannot matter while every admitted candidate carries ~0.93. Fixing the
  scale is a prerequisite for making `vec_weight` govern the tier.
* **#886** (open) — dense similarity should also *admit* working-memory
  candidates. Admission needs a calibrated score, which is what this issue is
  about; the saturated value cannot be compared against a floor.
* **#979** (open) — the default dense candidate pool excludes consolidated rows
  (a different code path: candidate generation, not scoring).

Measured on the same store: sweeping `MNEMOSYNE_WM_VEC_BLEND` from 0.0 to 0.8
with the legacy mapping moves orderings (30-38 of 40 queries) but never lifts
the correct row into the top 5; scoring from the blobs does.

## Proposed fix (happy to submit the PR)

* Score int8 candidates from their stored bytes with `_vec_int8_blob_cosine()`
  (the same treatment #911 gave the episodic arm); take the query blob from the
  same SQL quantizer the MATCH clause uses.
* Keep the legacy mapping when the store cannot return the vector column, and
  for every non-int8 arm, so older stores are unaffected.
* `MNEMOSYNE_WM_VEC_BLOB_SCORING` (default `1`) restores the legacy mapping
  when set to `0`.

Measured with a prototype of exactly this change on the same store copy:
`dense_max 0.9399 -> 0.7881`, target-in-top-5 `6/40 -> 8/40` (top-1 unchanged
at 2/40, n=40 — noisy, the structural claim is the restored scale).

## Checklist

      (#911 covers the episodic arm; #696/#789 are different concerns)
- [x] I have included the full traceback / error message (n/a: no exception,
      the arm silently returns a scale-guessed value)

manate-anagram:
## Description

Closes #982.

The working-memory vector arm (`_wm_vec_search_sqlite()`) was the only arm
converting a sqlite-vec distance into a similarity with an ad-hoc scale guess,
`1 - distance / (2 * EMBEDDING_DIM)`. `_vec_distance_sim()` documents that an
int8 distance cannot yield an absolute cosine (the quantization scale is
dimension- and distribution-dependent, so it abstains), and the episodic paths
already score from the stored bytes with `_vec_int8_blob_cosine()` (#911). This
PR gives the working-memory arm the same treatment.

The consumer is the working-memory dense blend:

```python
if vec_sim > 0:
    base_score = base_score * 0.80 + vec_sim * 0.20
```

The scale guess squeezed every vector-matched candidate into a narrow high band,
so the blend added a near-constant term: it could penalise weak candidates but
never lift a strong one. On a cosine ladder against the same rows the removed
mapping reports **0.974 → 0.884** for true cosines 0.95 → 0.00 — a ~10x
compression, and an *orthogonal* row kept a 0.88 similarity.

`_wm_vec_row_sim()` now scores int8 candidates from the stored blob
(`_vec_int8_blob_cosine()`) and **abstains** when a blob is unavailable or
length-mismatched: an int8 candidate is never scored from its distance. That
abstain makes `_wm_vec_search_sqlite()` return nothing, so `_wm_vec_search()`
serves the candidate set from the exact compatibility scan it already used for
older stores — candidate membership is unchanged, only the dense voice is.

**Scope, per your triage in #982:**
- blob-based int8 scoring in the Working Memory sqlite-vec path;
- a direct regression test for score separation;
- a `BeamMemory.recall()` gold-versus-distractor test;
- the old distance heuristic is **not retained as a fallback**: the opt-out env
  var from the first revision of this branch was removed, and no code path maps
  an int8 distance to a similarity any more;
- admission thresholds, blend weights, candidate-pool behaviour, migrations and
  reindexing are untouched. Non-int8 arms (float32/bit) keep their existing
  mapping.

## Related Issue

Closes #982

## Type of Change

- [x] Bug fix
- [x] Test improvement

## How Has This Been Tested?

New file `tests/test_wm_vec_blob_scoring.py` — **11 tests**:

- int8 scoring from blobs (identical / ~0.8 / orthogonal blobs);
- the abstain contract: missing, empty and length-mismatched blobs return
  "not scorable" instead of a distance-derived number;
- no legacy escape hatch remains (`MNEMOSYNE_WM_VEC_BLOB_SCORING=0` no longer
  restores the mapping);
- float32 / bit / unknown arms keep their mapping;
- **score separation** on a temporary sqlite-vec store: a cosine ladder
  (0.95 → 0.00) comes back as the true cosines while the removed mapping, run
  against the same rows through the same KNN query, stays inside a 0.09 band;
- abstaining candidates are served by the exact compatibility scan;
- `BeamMemory.recall()` ranks a gold row above a distractor: both rows share the
  query tokens (both reach the pool through FTS) and differ only in direction —
  gold `dense_score` ≥ 0.9, distractor ≤ 0.1.

**The tests bite:** with the removed mapping reinstated, **5 of 11 fail**,
including the two triage asked for (score separation and the recall
gold-versus-distractor case). With the fix, 11 pass.

Existing suites in this environment (missing optional deps here, so the
comparison is "no new failures"): `test_wm_vec_blob_scoring`,
`test_beam_vec_admission`, `test_beam_wm_vec_dialog_exclusion`,
`test_consolidate_vec_fallback` → 72 passed; `test_degrade_vector` reports the
**identical** 3 pre-existing failures before and after the change. CI will
confirm the real suite.

Manual verification: a read-only sweep on a store copy (same harness, same
queries) measured `dense_score` max **0.9399 → 0.7881** and
target-in-top-5 **6/40 → 8/40**. The scoring itself is unchanged since that
sweep; the switch and the fallback were removed afterwards, per your review
note.

## Checklist

- [x] Existing tests still pass (no new failures; see above)
- [x] New tests added for the fixed behaviour
- [x] Manual verification (numbers above)
- [x] `CHANGELOG.md` updated under `[Unreleased] > Fixed`
- [ ] Version bump: left to the release step, following the precedent of the
      recent fixes that did not bump `mnemosyne/__init__.py`. Happy to bump if
      you prefer.
- [x] No new dependencies, no cloud/API requirement, SQLite only

## Points worth a reviewer's eye

1. **Abstain instead of guess.** An int8 row whose blob cannot be read produces
   no `sim` value at all, and `_wm_vec_search()` then uses the exact
   compatibility scan for that candidate set. Alternative: score such rows 0.0
   and keep them in the vector arm — say which you prefer; abstaining keeps a
   fabricated number out of the blend, which is what #982 is about.
2. **Non-int8 arms are deliberately untouched.** float32/bit keep
   `1 - distance / (2 * dim)`. Treating them the same would go beyond the
   reported failure and the triage scope.
3. **The recall-level test pins the linear engine** (`MNEMOSYNE_POLYPHONIC_RECALL`
   unset): the polyphonic branch returns before the linear working-memory arm, so
   it exercises a different scoring path. I can add a polyphonic-specific
   assertion if that is wanted here.
4. **sqlite-vec only reports `distance` inside a MATCH query**, so the
   regression computes the removed mapping with the same KNN query the search
   path issues rather than reading the column directly.
5. **CHANGELOG wording** describes the removal explicitly, so anyone upgrading
   does not look for the old opt-out variable (it never shipped in a release).

## Summary

Working-memory sqlite-vec int8 candidates now use `_vec_int8_blob_cosine()` on stored vector blobs. This removes saturated distance estimation and improves dense-score separation.

If blob data is unavailable or invalid, the vector arm abstains. The existing exact compatibility scan handles the candidate set. Non-int8 scoring and retrieval controls remain unchanged.

## Architecture impact

- **Tiered memory:** Only working-memory retrieval changes. Episodic retrieval, consolidation, and BEAM tiering remain unchanged.
- **Retrieval:** Working-memory int8 scoring now matches the episodic stored-blob approach.
- **Veracity and sync:** No changes.
- **Hermes, MCP, and CLI:** No integration or public-surface changes.
- **Privacy and local-first design:** Preserved. Scoring uses local SQLite blobs and adds no network dependency or data exposure.
- **Maintainability:** This is the right call. It removes an invalid distance-scale assumption and aligns working-memory and episodic scoring.
- **Benchmark methodology:** Regression tests cover cosine scoring, abstention, fallback routing, non-int8 behavior, score separation, mixed candidates, and gold-versus-distractor ranking. The 12-test suite passed in three consecutive local runs.


tests/test_wm_vec_blob_scoring.py
```diff
@@ -0,0 +1,314 @@
...
+    With scoring unavailable the arm returns nothing, so `_wm_vec_search` serves
+    the candidate set from the compatibility scan, whose similarities are exact
+    cosines rather than a fabricated band.
+    """
+    beam = BeamMemory(session_id="wm-vec-fallback", db_path=temp_db)
+    if not beam_module._wm_vec_available(beam.conn):
+        pytest.skip("sqlite-vec vec_working table unavailable")
+    query = _blob_fixture(beam, "wm-vec-fallback")
+
+    monkeypatch.setattr(beam_module, "_wm_vec_row_sim", lambda *a, **kw: None)
```
dplush, line 263: **Blocking: missing mixed-candidate fallback coverage**

The implementation returns `[]` as soon as one `_wm_vec_row_sim()` call returns `None`, then `_wm_vec_search()` falls back. The current test monkeypatches the helper to return `None` for *every* candidate, so it does not prove that a partially scorable sqlite-vec result cannot lose candidates or scores. Please make exactly one of the three candidates unscorable, return the expected cosine for the other two, and assert that the fallback returns all three rows with their exact cosine scores.

dplush (changes requested):
## Review: changes requested

The production change is directionally correct and the current CI is green, but one required fallback invariant is not proved by the regression suite. `_wm_vec_search_sqlite()` abandons the whole vector arm when any candidate is unscorable, so the test must cover a *mixed* result set, not only the all-unscorable case. Please update the fallback test so exactly one of three candidates returns `None` from `_wm_vec_row_sim`, while the other two return their real cosine scores, and assert that the compatibility fallback returns all three rows with the expected exact scores. This is the remaining merge blocker; the existing all-unscorable test can remain as an additional case.

manate-anagram:
Added on `e917def`: `test_mixed_scorable_candidates_keep_every_row_and_score`.

The test reads the `wm-angled` candidate's stored blob from `vec_working` and patches `_wm_vec_row_sim` so that blob alone returns `None`, while the other candidates go through the real scorer. It then asserts two things: exactly one candidate abstained, and the compatibility fallback still returns **all three** rows with their exact cosines (`wm-same` 1.0, `wm-angled` 0.8, `wm-orthogonal` 0.0). The existing all-unscorable case stays as its own regression, so both halves of the invariant are pinned.

The arm's behaviour is unchanged by this: it still abstains on the first unscorable candidate and hands the complete candidate set to the exact scan. What the regression now proves is that a partially scorable sqlite-vec result cannot silently shrink the candidate set or downgrade its scores.

Local run: `tests/test_wm_vec_blob_scoring.py` 12 passed, stable across three consecutive runs.

dplush (approved):
Reviewed current head e917def. The int8 working-memory path now uses stored-byte cosine scoring and cleanly abstains to the compatibility scan; the mixed scorable/unscorable fallback invariant is covered. Local verification: new regression suite 12 passed; related vector/fallback suites 56 passed. CI is green across Python 3.10–3.13, lint, docs, build, and CLA.

Issue #1004, opened 2026-09-19
Design: atomic mutation and audit persistence in Hermes providers


dplush:
## Context
PR #999 adds audit parity for direct and approved/replayed updates, but mutation and audit persistence are still separate operations in the Hermes provider surfaces.

A failure between the durable memory mutation and its audit event can leave a committed mutation without the corresponding audit trail. Conversely, changing this casually could alter existing transaction and approval semantics.

## Decision needed
Define the transaction/event contract for Hermes provider mutations and approval replay:
- which mutation and audit rows must commit atomically;
- how audit failures affect the user-visible mutation;
- how direct writes and approved writes share the contract;
- how retries and already-committed mutations are identified;
- how streaming/listener events relate to the durable audit record.

Design alternatives may include a shared unit-of-work, transactional outbox, or an explicit documented non-atomic boundary. Do not add a local try/except patch that silently changes durability guarantees.

## Acceptance criteria
- documented transaction and failure semantics;
- coverage for direct and approval-gated remember/update/forget/invalidate paths;
- commit-failure and retry tests;
- provider parity across both Hermes implementations;
- compatibility and migration notes if audit behavior changes.

## Related
- PR #999: current audit parity work
- #766: transaction-aware streaming events for direct-wrapper writes

Issue #1003, opened 2026-09-19
Design: transactional atomicity for approved pending batches


dplush:
## Context
PR #999 (`https://github.com/mnemosyne-oss/mnemosyne/pull/999`) makes staged pending replay safe at the individual-record level: claims are atomic, failed records are retained, and committed claims are non-replayable.

CodeRabbit identified a separate contract question: one `mnemosyne_batch` approval can currently be represented by multiple staged replay records and replayed with partial success. A later failure can leave earlier operations committed.

## Decision needed
Define whether an approved batch must be all-or-nothing at replay time, or whether independently approved records intentionally permit partial success.

If all-or-nothing is required, design a transaction-aware batch replay contract that:
- groups records by their originating batch;
- replays all operations in one transaction;
- consumes claims only after the transaction commits;
- restores/retries the complete batch safely after failure;
- preserves provider/session/channel attribution.

Do not implement ad-hoc compensating deletes. The contract must cover mixed action shapes and audit events.

## Acceptance criteria
- documented batch atomicity semantics;
- regression tests for success, later mutation failure, retry, and claim state;
- tests for both Hermes provider surfaces;
- explicit compatibility decision for existing partial-success behavior.

## Related
- PR #999: individual pending-record replay safety
- #727: distinct JSON import atomicity problem

Issue #498, opened 2026-07-20
SEGV in sqlite3_clear_bindings (NULL stmt) — auto_sleep daemon-thread races with gateway pre_llm_call prefetch over shared check_same_thread=False connection


wardie:
## Summary

Running mnemosyne as the Hermes memory provider + MCP server, the mnemosyne process (and sometimes the host Hermes gateway) repeatedly segfaults mid-turn with a NULL-pointer deref inside `sqlite3_clear_bindings`. Disabling `auto_sleep` (`MNEMOSYNE_AUTO_SLEEP_ENABLED=false`) eliminates the crashes completely.

This looks closely related to #432 (write lock held across LLM calls), #472 (thread-local isolation), and the still-open #492 (working embedding parent race). My evidence points at a remaining race between mnemosyne's `auto_sleep` daemon thread and the gateway's `pre_llm_call` prefetch path, both using the same SQLite connection opened with `check_same_thread=False`.

## Environment

- mnemosyne-memory `3.14.0`
- mnemosyne-hermes `0.5.0` (Hermes Agent plugin)
- sqlite-vec `0.1.9`
- Python `3.11.15` (CPython, `uv`-managed standalone build)
- SQLite `3.50.4`
- Host: Hermes Agent gateway (`hermes-agent` 0.18.2) running as a systemd user service on Ubuntu (LXC)
- mnemosyne DB: `~/.hermes/mnemosyne/data/mnemosyne.db` (WAL mode, 9 working memories, 2 episodic)

## Symptom

The mnemosyne MCP server child process (comm = `mnemosyne`) and the Hermes gateway itself (comm = `hermes`) crash mid-turn with identical stack fingerprints:

```
[Mon Jul 20 18:26:45 2026] mnemosyne[1615549]: segfault at 0 ip 000063c273da504e sp 00007ffdb73f7570 error 4 in python3.11[9e004e,63c2735a1000+a2c000] likely on CPU 1 (core 1, socket 0)
[Mon Jul 20 18:26:47 2026] mnemosyne[1615560]: segfault at 0 ip 00005680d557104e sp 00007ffe73c9e5f0 error 4 in python3.11[9e004e,5680d4d6d000+a2c000]
[Mon Jul 20 18:43:58 2026] hermes[1622417]: segfault at 0 ip 00005e0616ec104e ... in python3.11[9e004e,...]
[...]

PR #1006, opened 2026-09-19
fix(hermes): serialize scoped replay beam access

### LINKED ISSUE #1005: Fix: serialize root-provider Beam access during scoped replay
## Context
PR #999 temporarily rebinds a provider Beam to the recorded session/channel while replaying a staged pending record. The replay scope is protected by a provider access lock, but root-provider operations such as `handle_tool_call()` and `on_memory_write()` do not necessarily hold the same lock across their Beam operation.

A concurrent callback could therefore write while the provider is temporarily rebound and land in the staged session/channel instead of the active scope.

## Proposed direction
Define and implement one root-provider Beam access protocol:
- use a reentrant lock where nested tool dispatch can re-enter the provider;
- hold it across root-provider Beam reads/writes and scoped replay;
- preserve correct session/channel restoration on success and failure;
- avoid deadlocks for `handle_tool_call()` dispatching `mnemosyne_apply_pending`.

## Acceptance criteria
- concurrency tests with replay plus `handle_tool_call()`/`on_memory_write()`;
- no writes attributed to the temporary replay scope from concurrent root callbacks;
- correct restoration after success and failure;
- both Hermes provider surfaces remain behaviorally aligned;
- no regression in normal session switching or background review.

## Related
- PR #999: current scoped replay implementation
- #327: broader Hermes gateway identity/provider scoping design

dplush:
## Summary

- serialize root-provider tool calls, memory-write callbacks, and background worker snapshots with the same Beam lock used by scoped pending replay
- make lazy Beam locks reentrant on both Hermes provider surfaces
- cover concurrent replay, success/failure restoration, background snapshot isolation, prefetch isolation, and nested apply-pending dispatch in the provider parity suite

Closes #1005.

## Why this boundary

`_replay_scope_locked()` temporarily rebinds the live Beam, so provider entry points and worker snapshots that use that Beam must share its lock for their complete operation. Reusing the live Beam preserves its author identity and connection. A second Beam would duplicate the connection and lose live provider state.

## Tests

- `tests/test_hermes_provider_parity.py`: 89 passed
- `tests/test_hermes_memory_provider.py`: 65 passed
- pending replay suites: 96 passed
- changed-file Ruff and `py_compile`: passed
- regression sabotage against the old root-provider behavior: both concurrent callback cases failed as expected
- adjacent provider suites: 151 passed; 3 unrelated failures reproduced unchanged on the exact base (`test_direct_initialize_blocks_sync_turn_until_beam_is_published` and both `test_validate_delete_removes_vec_working_row` variants)

## Scope

This does not change pending batch atomicity (#1003), audit transaction design (#1004), storage schema, or public APIs. Rollback is the provider lock and its focused parity tests.

## Summary

This PR serializes BEAM access across Hermes provider operations and scoped pending replay.

- Both Hermes provider surfaces use reentrant BEAM access locks.
- `handle_tool_call()`, `on_memory_write()`, prefetch, auto-sleep, and session-end consolidation use the shared lock.
- Nested `mnemosyne_apply_pending` replay can re-enter the lock without deadlock.
- Replay restores BEAM and session/channel state after success and failure.
- Tests cover concurrent callbacks, worker isolation, restoration, prefetch locking, and nested dispatch.

## Architectural impact

- **Core memory:** Protects BEAM writes, replay boundaries, prefetch, and consolidation snapshots. It does not change working memory, episodic memory, retrieval, veracity, storage schema, or pending-batch atomicity.
- **Privacy:** Prevents concurrent callbacks from writing to a temporary replay session or channel.
- **Local-first:** Preserves the live BEAM connection and author identity. No remote or sync changes are reported.
- **Agent integrations:** Keeps both Hermes provider surfaces aligned. No MCP or CLI changes are reported.
- **Maintainability:** Establishes one reentrant BEAM access protocol and adds focused concurrency regression tests.
- **Benchmarks:** No benchmark methodology changes are reported.

The shared reentrant access protocol is the right architectural choice. No local-first design erosion is reported.


dplush:
The new current-head availability finding about releasing `_beam_access_lock` after a background sleep timeout is not safe to apply in this #1005 fix. That lock was introduced by #498 specifically to serialize the auto-sleep worker’s SQLite access with live-Beam operations after a reproduced `sqlite3_clear_bindings` SEGV. Simply releasing it while `BeamMemory.sleep()` continues would reopen that crash race. Splitting `BeamMemory.sleep()` into separately locked database phases and unlocked LLM/model phases would be a broader core consolidation/transaction refactor, not a narrow replay-scope fix, and risks overlapping #1003/#1004. This PR therefore keeps the #498 serialization invariant while fixing the scoped-replay snapshot and prefetch leaks.

